PR #6, opened 2023-01-19
add health status reporter for storageclient

Madhu-1:
Adding the status health report of the storage client to the provider cluster.
    
Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

PR #7, opened 2023-01-24
ocs-client-operator implementation 

Madhu-1:
This replaces #3,  #4, #5, and #6, as this includes all the planned changes.


controllers/storageclient_controller.go
```diff
@@ -0,0 +1,307 @@
...
+		if st, ok := status.FromError(err); ok {
+			s.logGrpcErrorAndReportEvent(instance, OffboardConsumer, err, st.Code())
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to offboard consumer: %v", err)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (s *StorageClientReconciler) logGrpcErrorAndReportEvent(instance *v1alpha1.StorageClient, grpcCallName string, err error, errCode codes.Code) {
```
jarrpa, line 310: The fact there's an `And` in the function name immediately flags that this should be split. Looking at it deeper, the first part of the function is just a selection of strings, and the actual work of the function is two lines. Similarly, only one of these groups of lines is ever used more than once.

I can kind of understand want to have a single place to abstract error reporting, but I don't think this is really necessary.  Given that this is only even used in three places, that also indicates this is probably needless abstraction. I could still see an argument for making it easy to write and reference our error reporting strings, so it might make sense to just define some struct like this, and define a few of these at the top of some file:

```go
type struct StorageClientGrpcEvent {
    msg string
    eventReason string
    eventType string
    errorCode codes.Code
}
```

Then you could go to the three places this function is currently used and replace it with the following, which avoids us falling into the trap of constraining ourselves to an abstraction format we might not consistently need:
```go
		if st, ok := status.FromError(err); ok {
			var errStruct = SomeVarName
			s.Log.Error(err, "StorageProvider:"+grpcCallName+":"+errStruct.msg)
			s.recorder.ReportIfNotPresent(instance, errStruct.eventType, errStruct.eventReason, errStruct.msg)
		}
		return reconcile.Result{}, fmt.Errorf(errStruct.msg, err)
```

If you agree with this assessment, this can also come in as a future PR from you or someone else. If you don't, that's valid, let's hear why.

controllers/suite_test.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
```
jarrpa, line 27: Isn't this a duplicate of the one that was already here?

main.go
```diff
@@ -32,6 +32,7 @@ import (
 	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	configv1 "github.com/openshift/api/config/v1"
 	secv1 "github.com/openshift/api/security/v1"
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
```
jarrpa, line 35: Isn't this a duplicate of the one that was already here?

pkg/csi/monconfigmap.go
```diff
@@ -0,0 +1,224 @@
...
+
+	return nil
+}
+
+// UpdateMonConfigMap updates the config map used to provide ceph-csi with
+// basic cluster configuration. The clusterNamespace and clusterInfo are
+// used to determine what "cluster" in the config map will be updated and
+// the clusterNamespace value is expected to match the clusterID
+// value that is provided to ceph-csi uses in the storage class.
+// The locker l is typically a mutex and is used to prevent the config
```
jarrpa, line 165: I don't see an `l` here, does this refer to the global `configMutex`?

jarrpa (changes requested):
Overall this is a great PR, thank you for accommodating my commit requests. :pray: Apart from the more specific review comments, I have a couple of broad questions:

Regarding the StorageClient CRD:
 - I remember at some point having a conversation (months ago) about naming this StorageClient vs OcsClient. The reasoning for the latter probably make more sense when it was `ocsclient.odf.openshift.io`, but I can see `storageclient.ocs.openshift.io` also making sense these days. Is this related to the referenced (though unexplained) conversation about the API group name?
- Why are we limiting to one per namespace? I can imagine the argument being something about Secret/data isolation, but could I get a concrete example of where this would be an *actual* problem with the current design?

Regarding the StorageClassClaim implementation, this seems largely similar to the work I did in the previous PRs. If this is the case, could you add a `Co-authored-by: jarrpa@redhat.com` line to the relevant commit messages? If you feel any of the other commits are significantly based on that same prior work, please do add the line there as well.

Beyond that:
- Your commits are signed using your personal gmail account. I would think you should be using your @redhat.com account, was this intentional?
- I like the implementation of the status-report job, but boy does it look weird to have a `main.go` under `pkg`. Would it be possible to do as was done for the Provider and create a `service/status-reporter` directory for that job?

controllers/storageclient_controller.go
Madhu-1: Yes, let's take it as a follow-up, will open a GitHub issue to track it.

main.go
```diff
@@ -88,6 +108,48 @@ func main() {
...
+			setupLog.Error(err, "unable to initialize sidecars")
+			os.Exit(1)
+		}
+
+		err = csi.EnsureCSI(ctx, client, setupLog)
+		if err != nil {
+			setupLog.Error(err, "unable to create/update CSI")
+			os.Exit(1)
+		}
+	}
```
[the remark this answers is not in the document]
Madhu-1: In the standard kubernetes version, there is no clusterversion CR. AFAIK the server version will get updated first and during the kubelet upgrades on the nodes, the container will get restarted. the operator restart in this case, will read the server version and takes care of picking the right driver version. I was not sure about the upgrade in the OCP version and if the logic remains the same we can remove the controller for handling the csi resources.

pkg/csi/csi.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		log.Error(err, "unable to create mon configmap")
+		return err
+	}
+
+	err = createEncryptionConfigMap(ctx, client, operatorDeployment, log)
+	if err != nil {
+		log.Error(err, "unable to create encryption configmap")
+		return err
+	}
```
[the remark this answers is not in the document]
Madhu-1: sure added an explanation.

pkg/csi/csidriver.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if err != nil {
+		log.Error(err, "error creating/updating CSIDriver", "driverName", csiDriver.Name)
+	}
+
+	return err
+}
+
+// TODO need to check how to delete the csidriver object
```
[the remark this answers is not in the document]
Madhu-1: There is no caller for this function it's a dead code/place holder. it's a TODO to handle the resource deletion/cleanup for the cluster scoped resources.

.github/workflows/conf/commitlintrc.json
```diff
@@ -71,6 +71,8 @@
                 "api",
                 "bundle",
                 "ci",
+		"csi",
+		"cleanup",
```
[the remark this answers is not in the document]
Madhu-1: I dont see any indentation problem. I also expect commitlint to throw an error message saying not able to parse the config file if there is any indentation problem. Please let me know how to find the indentation problem.

images.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+- version: v1.24
+  containerImages:
+    provisionerImageURL: "registry.k8s.io/sig-storage/csi-provisioner:v3.3.0"
+    attacherImageURL: "registry.k8s.io/sig-storage/csi-attacher:v4.0.0"
+    resizerImageURL: "registry.k8s.io/sig-storage/csi-resizer:v1.6.0"
+    snapshotterImageURL: "registry.k8s.io/sig-storage/csi-snapshotter:v6.1.0"
+    driverRegistrarImageURL: "registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.5.1"
+    cephCSIImageURL: "quay.io/cephcsi/cephcsi:v3.7.2"
+    csiaddonsImageURL: "quay.io/csiaddons/k8s-sidecar:v0.5.0"
```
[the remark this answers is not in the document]
Madhu-1: Sure, let's have a discussion on this one, I will open an issue for further discussion and tracking.

pkg/csi/csidriver.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+	}
+	err := client.Get(ctx, types.NamespacedName{Name: csiDriver.Name}, actualDriver)
+
+	if errors.IsNotFound(err) {
+		err = client.Create(ctx, csiDriver)
+	} else if err == nil {
+		// As FSGroupPolicy field is immutable, if it requires an
+		// update we need to delete the CSIDriver object first and
+		// requeue the reconcile request.
```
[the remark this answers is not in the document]
Madhu-1: Only a few fields in the csidriver object are immutable and cannot be changed.

Madhu-1:
> Overall this is a great PR, thank you for accommodating my commit requests. pray Apart from the more specific review comments, I have a couple of broad questions:
> 
> Regarding the StorageClient CRD:
> 
> * I remember at some point having a conversation (months ago) about naming this StorageClient vs OcsClient. The reasoning for the latter probably make more sense when it was `ocsclient.odf.openshift.io`, but I can see `storageclient.ocs.openshift.io` also making sense these days. Is this related to the referenced (though unexplained) conversation about the API group name?

It was discussed in one of the meetings with @nb-ohad and @oritwas as it's more opt and as the name explains it's just a storageclient we decided on storage client name. it's not documented anywhere.

> * Why are we limiting to one per namespace? I can imagine the argument being something about Secret/data isolation, but could I get a concrete example of where this would be an _actual_ problem with the current design?
> 

To keep it simple for now, we will go with one storageClient per namespace but in future/follow-up PR we will block creating multiple storageclient per provider and allow multiple storageclient connecting to different providers in the same namespace.

> Regarding the StorageClassClaim implementation, this seems largely similar to the work I did in the previous PRs. If this is the case, could you add a `Co-authored-by: jarrpa@redhat.com` line to the relevant commit messages? If you feel any of the other commits are significantly based on that same prior work, please do add the line there as well.
> 

Sorry missed adding it. Done :+1:

> Beyond that:
> 
> * Your commits are signed using your personal gmail account. I would think you should be using your @redhat.com account, was this intentional?

yes, it's intentional I use my Gmail id for all the projects I work with. Never used redhat.com till now.

> * I like the implementation of the status-report job, but boy does it look weird to have a `main.go` under `pkg`. Would it be possible to do as was done for the Provider and create a `service/status-reporter` directory for that job?

Done :+1:

jarrpa (changes requested):

jarrpa:
> > * Your commits are signed using your personal gmail account. I would think you should be using your @redhat.com account, was this intentional?
> 
> yes, it's intentional I use my Gmail id for all the projects I work with. Never used redhat.com till now.

I haven't seen this before. Best I knew if you're doing work for redhat you should be signing it as such. But if that's what you've been doing and no one else has said anything about, I'm not gonna care. :laughing:

main.go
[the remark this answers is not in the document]
Madhu-1: >since we have a periodic job running for the heartbeat, that should be taking care of the majority of things that we have to try and reconcile between upgrades, right

The heartbeat makes 2 RPC calls , the heartbeat job on the client side The heartbeat job is only to refresh the monitors IP and it will not,

between upgrades we only need to update the csi objects. let's keep it for now if theory works we can take the ontroller out :D

pkg/csi/csidriver.go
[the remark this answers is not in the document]
Madhu-1: Won't break, but we need to make 2 API calls instead of one.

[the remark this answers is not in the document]
jarrpa: Perfect. :+1:

pkg/csi/csi.go
```diff
@@ -0,0 +1,153 @@
...
+func EnsureCSI(ctx context.Context, client client.Client, log logr.Logger) error {
+	err := ensureRBAC(ctx, client, log)
+	if err != nil {
+		log.Error(err, "unable to create/update RBAC")
+		return err
+	}
+
+	// createEncryptionConfigMap creates the monitor configmap for the csi
+	// driver but never updates it. This is because the monitor configirations
+	// are added to the configmap when user creates storageclassclaims.
```
jarrpa, line 66:
```suggestion
	// createMonConfigMap creates the monitor configmap for the csi
	// driver but never updates it. This is because the monitor configurations
	// are added to the configmap when user creates storageclassclaims.
```

main.go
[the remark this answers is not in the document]
jarrpa: Sounds good to me!

controllers/storageclient_controller.go
jarrpa: Issue #8 has been opened.

jarrpa (changes requested):
Excellent work! Just one last typo to fix, I don't entirely want to hold up the entire PR for this, but if not now then who knows when. :laughing: If you or someone else has permissions to hit the big green merge button once you update the PR I hereby consent to you doing so ASAP.

Madhu-1:
@jarrpa Please take a look at it; below are the major changes after discussing with @nb-ohad 

* Remove the OCP check and consider the cluster is always OCP for now. I will open an issue to track the gap to run with standalone kubernetes
* Move complete CSI logic to ClusterVersion CR.
* Move CSI RBAC from code to yaml files which help in security audit

Madhu-1:
Test result in OCP

```bash
[🎩︎]mrajanna@fedora ocs-client-operator $]oc get po -nocs-client-ns
NAME                                                     READY   STATUS    RESTARTS   AGE
csi-addons-controller-manager-7fc65dff49-7mpxs           2/2     Running   0          49m
csi-cephfsplugin-6wtgv                                   2/2     Running   0          2m39s
csi-cephfsplugin-bwn6g                                   2/2     Running   0          2m52s
csi-cephfsplugin-p9f97                                   2/2     Running   0          2m49s
csi-cephfsplugin-provisioner-7bbcd8d9bf-gqgj9            5/5     Running   0          23m
csi-cephfsplugin-provisioner-7bbcd8d9bf-gqrxx            5/5     Running   0          23m
csi-rbdplugin-5k5cp                                      3/3     Running   0          25m
csi-rbdplugin-6mv8w                                      3/3     Running   0          25m
csi-rbdplugin-7bzzl                                      3/3     Running   0          25m
csi-rbdplugin-provisioner-5f8fd786d-65rx5                5/5     Running   0          3m12s
csi-rbdplugin-provisioner-5f8fd786d-hppg9                5/5     Running   0          3m18s
ocs-client-operator-controller-manager-d8df78b86-ls2gs   2/2     Running   0          3m43s
```

controllers/clusterversion_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	ctx                context.Context
+	Log                klog.Logger
+	Scheme             *runtime.Scheme
+
+	cephFSDeployment *appsv1.Deployment
+	cephFSDaemonSet  *appsv1.DaemonSet
+	rbdDeployment    *appsv1.Deployment
+	rbdDaemonSet     *appsv1.DaemonSet
+	scc              *secv1.SecurityContextConstraints
+}
```
jarrpa, line 50: Is there a general pattern for which ones are capitalized and which ones aren't? I imagine the reconciler struct in general isn't meant to be used by anything other than itself, so I'd think they'd be all lowercase. I'm not hardline one way or the other, but a consistent rationale would help things down the line.

pkg/utils/deployment.go
```diff
@@ -0,0 +1,36 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetOperatorDeployment returns the operator deployment object
+func GetOperatorDeployment(ctx context.Context, c client.Client, namespace string) (*appsv1.Deployment, error) {
+	deployment := &appsv1.Deployment{}
+	err := c.Get(ctx, types.NamespacedName{Name: "ocs-client-operator-controller-manager", Namespace: namespace}, deployment)
```
jarrpa, line 30: There's not a 100% guarantee that this will always be the name of the *Deployment* resource... we can probably leave this as-is for the sake of expediency, but I would like to see something that gets the Deployment based on the metadata of the Pod.

controllers/storageclient_controller.go
```diff
@@ -0,0 +1,307 @@
...
+		return nil, fmt.Errorf("failed to create a new provider client: %v", err)
+	}
+
+	return ocsProviderClient, nil
+}
+
+// onboardConsumer makes an API call to the external storage provider cluster for onboarding
+func (s *StorageClientReconciler) onboardConsumer(instance *v1alpha1.StorageClient, externalClusterClient *providerClient.OCSProviderClient) (reconcile.Result, error) {
+
+	// TODO Need to find a way to this one as its OCP dependent
```
jarrpa, line 226: I have no idea what this is saying, I think there's a verb missing?

jarrpa (changes requested):
Just a few small things left. Apart from the code questions, there's a commit description that should be amended:

```
 csi: install ceph-csi driver

Install ceph-CSI driver based on the
kubernetes version where the ocs-client
operator is running.

This also includes clusterversion controller
which can trigger an upgrade of csi driver
if required when a OCP cluster is upgraded.

There is a if/else check in main.go which helps
to install/test csidriver changes by checking
this is a ocp cluster or kubernetes cluster.
```

The if/else mentioned has been removed.

controllers/clusterversion_controller.go
Madhu-1: The values that need to be set in  the main.go are exported all others are unexported.

controllers/storageclient_controller.go
Madhu-1: Reworded it.

pkg/utils/deployment.go
jarrpa: PR looks good to merge, please create an issue for this comment when you're back online.

jarrpa (approved):
LGTM!

Issue #9, opened 2023-01-26
Allow multiple storageClient per namespace


Madhu-1:
Block creating multiple storageclient per provider as it doesn't make sense to connect to the same provider with different names and allow multiple storageclient connecting to different providers in the same namespace.

PR #12, opened 2023-02-02
Allow multiple storageclients per namespace

Madhu-1:
This commit blocks the creation of multiple storageclients for the same provider and allows the creation of multiple storageclients in the same namespace.

fixes: #9


jarrpa (approved):
I love it, thank you for finding `IndexField()` and implementing this.

PR #17, opened 2023-02-09
Fix bugs found in testing

Madhu-1:
Fixed the bugs found in the testing where migrated from ODF to the new operator.


controllers/storageclassclaim_controller.go
```diff
@@ -406,7 +406,7 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 		}
 
 		// Delete configmap entry for cephcsi
-		err = csi.UpdateMonConfigMap(r.ctx, r.Client, r.log, r.storageClassClaim.Name, r.storageClient.Status.ConsumerID, nil)
+		err = csi.UpdateMonConfigMap(r.ctx, r.Client, r.log, r.OperatorNamespace, r.storageClassClaim.Name, r.storageClient.Status.ConsumerID, nil)
```
jarrpa, line 416: When you're passing in this many variables that's usually a sign that the function needs to be refactored. I think you need to reconsider what goes in the CSI package at this point, you might want to have anything that requires a Client be moved to the reconciler.

Makefile
```diff
@@ -87,8 +87,11 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	sed -i "s|STATUS_REPORTER_IMAGE_VALUE|$(IMG)|g" config/manager/manager.yaml
 	cd config/default && $(KUSTOMIZE) edit set image rbac-proxy=$(RBAC_PROXY_IMG)
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	# Revert back the STATUS_REPORTER_IMAGE_VALUE var
+	sed -i "s|value: $(IMG)|value: STATUS_REPORTER_IMAGE_VALUE|g" config/manager/manager.yaml
```
jarrpa, line 94: Can't you just pipe the output of  `kustomize build` to `awk` and then to `kubectl`?

jarrpa (changes requested):

controllers/storageclassclaim_controller.go
Madhu-1: One reason for not moving this to reconcile is that this is a common function for both reconcile and status-reporter job. Moved common input to a new struct and made this as method to avoid more inputs

jarrpa (approved):
LGTM!

PR #19, opened 2023-03-13
csi: use openshift registry to pull ocp images

Madhu-1:
Use the openshift registry to pull the ocp sidecar images and odf reqistry to pull odf cephcsi and sidecar image.

cc @nb-ohad 
cc @jarrpa

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -1,20 +1,20 @@
 ---
 - version: v4.11
   containerImages:
-    provisionerImageURL: "registry.k8s.io/sig-storage/csi-provisioner:v3.3.0"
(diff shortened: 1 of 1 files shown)


deepshikhaaa:
Looks good to me!

Madhu-1:
If this needs to be downstream-only PR, we need to create the downstream branch (i dont have permission to do it) and change the target.

jarrpa (approved):
We shouldn't be branching for "downstream-only" reasons, a branch should only be needed per official release, of which there should (ideally) only be one version. As long as the `registry.redhat.io` images are accessible over the public internet without an account these changes are fine.

Main thing to consider is how easy it'll be for someone to write automation to specify different image URLs as needed., but if that requires any changes they can come in a later PR.

PR #21, opened 2023-03-23
Use configmap for the sidecar images mapping

Madhu-1:
* replace downstream tag with upstream tag
* Create a configmap from the yaml
* Mount the configmap as a file to the operator deployment
* During downstream build, we will replace the downstream floating tag with SHA
* The configmap will point to the downstream yaml


jarrpa (approved):
Excellent, LGTM!

PR #23, opened 2023-03-27
Fix bugs found in migration of ODF to ocs-client-operator

Madhu-1:
This PR contains two fixes
* Increase the resource limit to fix the OOMKilled problem
* remove the code which generates the subvolumegroup name if its present in the RPC call.

### DIFF
--- a/controllers/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim_controller.go
@@ -347,14 +347,7 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 
 				if resource.Name == "cephfs" {
 					csiClusterConfigEntry.CephFS = new(csi.CephFSSpec)
-					subVolumeGroupName := data["subvolumegroupname"]
-					// In ODF 4.12 the subvolume group name is not passed in
-					// the GetStorageConfig RPC call, so we need to generate it
-					// here as this is expected to work with ODF 4.12
-					if subVolumeGroupName == "" {
-						subVolumeGroupName = fmt.Sprintf("cephfilesystemsubvolumegroup-storageconsumer-%s", r.storageClient.Status.ConsumerID)
-					}
-					csiClusterConfigEntry.CephFS.SubvolumeGroup = subVolumeGroupName
+					csiClusterConfigEntry.CephFS.SubvolumeGroup = data["subvolumegroupname"]
 					// delete groupname from data as its not required in storageclass
 					delete(data, "subvolumegroupname")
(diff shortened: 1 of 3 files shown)


jarrpa (changes requested):
The idea is sound, but I think the ClusterVersion discovery can be immensely simplified.

controllers/storageclassclaim_controller.go
```diff
@@ -352,7 +353,7 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 					// the GetStorageConfig RPC call, so we need to generate it
 					// here as this is expected to work with ODF 4.12
 					if subVolumeGroupName == "" {
-						subVolumeGroupName = fmt.Sprintf("cephfilesystemsubvolumegroup-storageconsumer-%s", r.storageClient.Status.ConsumerID)
+						subVolumeGroupName = fmt.Sprintf("cephfilesystemsubvolumegroup-storageconsumer-%s", r.OpenshiftClusterID)
```
nb-ohad, line 356: @Madhu-1 Why are we tieing the CephFS to the openshift  cluster ID?
This can and will result in having this un-migratable to any other openshift cluster. Which is something we support as part of our backup and restore procedure. 

I think there is another aspect the should invalidate this approach. The fact that we are tieing our current selves and future selves into an openshift specific construct while the roadmap needs this operator to be able to run on other k8s distributions. A decision here will cause us a headache in the not so far future

I am aware that 
A decision

nb-ohad (changes requested):

controllers/storageclassclaim_controller.go
nb-ohad: Talking with @jarrpa about this fix we now understand that this fix raises a number of other potential issues. The client should not have assumptions about naming conventions that the provider side grants to the resources its provisions. especially where the provider itself does not use that conversion to locate the resource in question. 

The underlying reason that storage consumers are named `storageconsumer-<cluster-id>` in the first place was and is arbitrary and was done only to have something identifying the consumer cluster for a HUMAN reader.

The point is that there is a bug here, but it is not in this operator, it is in the OCS operator and as such should be fixed there and then backported to ocs-operator 4.12 (presumably as part of ocs-operator 4.12.3)

Madhu-1: @nb-ohad here is the PR https://github.com/red-hat-storage/ocs-operator/pull/1976 in ocs-opeartor which need to be merged and backported to 4.12

nb-ohad (approved):

jarrpa (approved):

PR #29, opened 2023-06-05
Increase total quota size allocated to a consumer

bindrad:
Currently, consumer clusters are only able to utilize 1T storage on consumers. So the quota limit is set on blockpool which does not allow to utilize storage above 1T.


Madhu-1 (approved):

rchikatw:
Closing this PR as the design and requirement is changed and Same implementation is been addressed in [PR](https://github.com/red-hat-storage/ocs-client-operator/pull/32)

PR #30, opened 2023-06-22
Removing dependency on capacity

rchikatw:
Changes in API parameters for Onboarding consumers. Now no need to pass hardcode value for the capacity


rchikatw:
Closing this PR as this is not valid PR no longer.

PR #34, opened 2023-10-04
controllers: send platform and operator version while reporting status

leelavg:
- send storage client operator version and version of platform to which
cluster last updated to successfully
- the status will be sent on top of existing heartbeat mechanism


leelavg:
depends on #33, however last commit contains the sole feature and can be reviewed.

service/status-report/main.go
```diff
@@ -77,6 +85,43 @@ func main() {
...
+	}
+
+	var opVersion string
+	for idx := range csvList.Items {
+		candidate := &csvList.Items[idx]
+		if strings.HasPrefix(candidate.Name, csvPrefix) {
+			opVersion = candidate.Spec.Version.String()
+			break
+		}
+	}
```
nb-ohad, line 100: I believe we implemented somewhere GetCSVByPrefix(prefix) which we can use in this case as well.

But preferably I would like to see if we can implement a 
```
func Find[T any](list []T, func (item *T) bool) *T 
```
util in the utils package.

This will allow us to replace the code above with:
```
csv := find(csvList.Items, func(csv *opv1a1.ClusterServiceVersion) {
    return strings.HasPrefix(csv.Name, csvPrefix)
})
```

service/status-report/main.go
```diff
@@ -77,6 +85,43 @@ func main() {
...
+
+	if opVersion == "" {
+		klog.Exitf("Unable to find csv with prefix %q", csvPrefix)
```
nb-ohad, line 103: I am not sure that existing is a good result of not finding a piece of data. The reason is that this will basically indicate to the server that there is no connection from the client (because of the lack of a heartbeat) which is not true. 
It is always preferable to send what you have and let the server decide how to react

service/status-report/main.go
```diff
@@ -77,6 +85,43 @@ func main() {
...
+	for idx := range clusterVersion.Status.History {
+		candidate := &clusterVersion.Status.History[idx]
+		if candidate.State == configv1.CompletedUpdate {
+			cVersion = candidate.Version
+			break
+		}
+	}
+
+	if cVersion == "" {
+		klog.Exitf("Unable to find ocp version with completed update")
```
nb-ohad, line 122: I am not sure that existing is a good result of not finding a piece of data. The reason is that this will basically indicate to the server that there is no connection from the client (because of the lack of a heartbeat) which is not true. 
It is always preferable to send what you have and let the server decide how to react

service/status-report/main.go
```diff
@@ -77,6 +85,43 @@ func main() {
...
+	}
+
+	if opVersion == "" {
+		klog.Exitf("Unable to find csv with prefix %q", csvPrefix)
+	}
+
+	clusterVersion := &configv1.ClusterVersion{}
+	clusterVersion.Name = "version"
+	if err = cl.Get(ctx, types.NamespacedName{Name: clusterVersion.Name}, clusterVersion); err != nil {
+		klog.Exitf("Failed to get clusterVersion: %v", err)
```
nb-ohad, line 109: I am not sure that existing is a good result of not finding a piece of data. The reason is that this will basically indicate to the server that there is no connection from the client (because of the lack of a heartbeat) which is not true. 
It is always preferable to send what you have and let the server decide how to react

service/status-report/main.go
```diff
@@ -77,6 +85,43 @@ func main() {
...
+	}
+
+	var cVersion string
+	for idx := range clusterVersion.Status.History {
+		candidate := &clusterVersion.Status.History[idx]
+		if candidate.State == configv1.CompletedUpdate {
+			cVersion = candidate.Version
+			break
+		}
+	}
```
nb-ohad, line 119: With the utils function in place, this can be simplified into
```suggestion
	record := utils.find(clusterVersion.Status.History, func (record *<type>) bool {
		return record.state == configv1.CompletedUpdate
	})
```

nb-ohad (changes requested):

service/status-report/main.go
leelavg: - ack, so should the server consider empty values as legit?
- in this specific case, empty string fails a semver check and a result heartbeat will not get updated in consumer cr
- so the idea is to decouple everything else from timestamp?

service/status-report/main.go
leelavg: done.

service/status-report/main.go
leelavg: done.

service/status-report/main.go
leelavg: done.

service/status-report/main.go
leelavg: done.

service/status-report/main.go
leelavg: done.

go.mod
```diff
@@ -86,3 +89,5 @@ require (
 )
 
 exclude github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
+
+replace github.com/red-hat-storage/ocs-operator/v4 => ../ocs-operator
```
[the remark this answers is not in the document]
leelavg: definitely, I'm awaiting that commit to be merged and so this PR is in draft state.

service/status-report/main.go
```diff
@@ -77,6 +84,39 @@ func main() {
...
+
+	var pltVersion string
+	clusterVersion := &configv1.ClusterVersion{}
+	clusterVersion.Name = "version"
+	if err = cl.Get(ctx, types.NamespacedName{Name: clusterVersion.Name}, clusterVersion); err != nil {
+		klog.Warningf("Failed to get clusterVersion: %v", err)
+	} else {
+		item := utils.Find(clusterVersion.Status.History, func(record *configv1.UpdateHistory) bool {
+			return record.State == configv1.CompletedUpdate
```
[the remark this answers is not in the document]
leelavg: realizing Desired version is multi step process and so we want to know the current cluster version, https://github.com/openshift/api/blob/7b36cb38c7d060be9d3a9e1b1b9ad0bf752d07be/config/v1/0000_00_cluster-version-operator_01_clusterversion.crd.yaml#L18

nb-ohad:
LGTM

Madhu-1:
LGTM

PR #35, opened 2023-10-09
BUG 2212333: api: removing dependency on hardcoded value when onboarding consumer

rchikatw:
when onboarding a new consumer using API parameters are changed the same change needs to be implement here. Now no longer need to pass the hard coded value while onboarding a consumer.


rchikatw:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2212333

rchikatw:
Closing this PR As we discontinue FaaS development and in the current situation, customers would not see provider/consumer support for 4.12/4.13 so this backport is not needed. Hence Closing the bug.

PR #36, opened 2023-10-09
BUG 2230928: api: removing dependency on hardcoded value when onboarding consumer

rchikatw:
when onboarding a new consumer using API parameters are changed the same change needs to be implement here. Now no longer need to pass the hard coded value while onboarding a consumer.


rchikatw:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2230928

Madhu-1:
@rchikatw This BZ is not having the right version set, This PR cannot be merged until the BZ is targeting for right version

rchikatw:
closing this PR  As we discontinue FaaS development and in the current situation, customers would not see provider/consumer support for 4.12/4.13 so this backport is not needed. Hence Closing the bug.

PR #39, opened 2023-11-09
Bug 2247748: A storage-client CronJob create too many jobs and pods causing maxPod limit to be reached

bernerhat:
Fixes the issue created where consecutive CronJobs iterations create pods unnecessarily and caused maxPod limit to be reached thus resulting in future pods scheduling to be stuck in a pending state.

* Added a ConcurrencyPolicy field to the CronJob for reconcileClientStatusReporterJob in StorageClientReconciler
* ConcurrencyPolicy chosen in this case to be `Forbid`

The `Forbid` option specifies waiting for the current job to finish before starting a new one and the `Replace` option starts a new job no matter what is the current one's status is (which can also create a new pod while the previous job’s pod is still terminating and increase pod count unnecessarily).
Since the created pod for this cronJob purpose is a heartbeat to the provider server i don't see the point in choosing to replace the current job with another, therefore forbid made more sense.

### DIFF
--- a/controllers/storageclient_controller.go
+++ b/controllers/storageclient_controller.go
@@ -433,14 +433,24 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
 	cronJob.Namespace = s.OperatorNamespace
 	addLabel(cronJob, storageClientNameLabel, instance.Name)
 	addLabel(cronJob, storageClientNamespaceLabel, instance.Namespace)
+	var podDeadLineSeconds int64 = 120
+	jobDeadLineSeconds := podDeadLineSeconds + 35
+	var keepJobResourceSeconds int32 = 600
+	var reducedKeptSuccecsful int32 = 1 
+
 
 	_, err := controllerutil.CreateOrUpdate(s.ctx, s.Client, cronJob, func() error {
 		cronJob.Spec = batchv1.CronJobSpec{
 			Schedule: "* * * * *",
+			ConcurrencyPolicy: batchv1.ForbidConcurrent,
+			SuccessfulJobsHistoryLimit: &reducedKeptSuccecsful,
 			JobTemplate: batchv1.JobTemplateSpec{
 				Spec: batchv1.JobSpec{
+					ActiveDeadlineSeconds: &jobDeadLineSeconds,
+					TTLSecondsAfterFinished: &keepJobResourceSeconds,
 					Template: corev1.PodTemplateSpec{
 						Spec: corev1.PodSpec{
+							ActiveDeadlineSeconds: &podDeadLineSeconds,
 							Containers: []corev1.Container{
 								{
 									Name:  "heartbeat",


leelavg:
@bernerhat could you pls add `controllers` as the prefix to commit msg to pass github actions?

controllers/storageclient_controller.go
```diff
@@ -470,6 +470,7 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
 					},
 				},
 			},
+			ConcurrencyPolicy: "Forbid",
```
leelavg, line 473: 1. Move the fields that directly belong to a struct to the top, ie, move this line to after 439, ie, after `Schedule` which generates diff for easy viewing
2. Even though the string `Forbid` is technically correct, since this is a enum-ish field, you should be searching for any predefined constants as a value for this field, ie, `batchv1.ForbidConcurrent` should be used
3. Please comment on how you simulated the scenario as well

controllers/storageclient_controller.go
```diff
@@ -433,14 +433,24 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
 	cronJob.Namespace = s.OperatorNamespace
 	addLabel(cronJob, storageClientNameLabel, instance.Name)
 	addLabel(cronJob, storageClientNamespaceLabel, instance.Namespace)
+	var jobDeadLineSeconds int64 = 155
+	var podDeadLineSeconds int64 = 120
+	var keepJobResourceSeconds int32 = 600
+	var reducedKeptSuccecsful int32 = 1
```
bernerhat, line 439: Had to create passable variables to the pointer variables required in the new fields. looking into other options as well.

bernerhat:
Firstly addressing the issue at hand, simulated and tested the options to either `Forbid` or `Replace` the current running Job locally using minikube and kubernetes version 1.27.4 

Secondly, Concerns brought up to me by Ohad were:
1. Setting a proper deadline for the pod to finish it's operation to prevent an infinite loop scenario.
2. Keeping last run objects to investigate such option

In order to address these concerns i've added a few new specs to the crafted CR:
- `TTLSecondsAfterFinished` - keep iteration objects (job and pod). set the ttl to 10 minutes, can be changed.
- `ActiveDeadlineSeconds` - deadline before killing the objects. Pod 2 minutes after consultation, Job 2 minutes 35 seconds.
- `SuccessfulJobsHistoryLimit` - set to 1. default value is 3, reduced to prevent unnecessary object clutter.

`ActiveDeadlineSeconds` on the job level was set to 2 minutes and 35 seconds in order to keep the pod. for some reason the pod was not kept after the deadline reached, after many testing i've managed to find that the pod was only kept after moving into error status 30 seconds after the deadline is reached and the kill order was issued (extra 5 seconds to be sure).

controllers/storageclient_controller.go
bernerhat: ack. all addressed.

leelavg:
> @bernerhat could you pls add `controllers` as the prefix to commit msg to pass github actions?
- missed, GH action is still failing.

controllers/storageclient_controller.go
```diff
@@ -433,14 +433,24 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
 	cronJob.Namespace = s.OperatorNamespace
 	addLabel(cronJob, storageClientNameLabel, instance.Name)
 	addLabel(cronJob, storageClientNamespaceLabel, instance.Namespace)
+	var jobDeadLineSeconds int64 = 155
+	var podDeadLineSeconds int64 = 120
```
leelavg, line 437: - as these two are related, pls show that in code as well like `jobSec = podSec + 35`

controllers/storageclient_controller.go
```diff
@@ -433,14 +433,24 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
...
+	var keepJobResourceSeconds int32 = 600
+	var reducedKeptSuccecsful int32 = 1 
+
 
 	_, err := controllerutil.CreateOrUpdate(s.ctx, s.Client, cronJob, func() error {
 		cronJob.Spec = batchv1.CronJobSpec{
```
leelavg, line 443: use `startindDeadlineSeconds` as well, just to safegaurd against 100 times job failure w/ Forbid. can be b/n `30 to 60s`

bernerhat:
> > @bernerhat could you pls add `controllers` as the prefix to commit msg to pass github actions?
> 
> * missed, GH action is still failing.

did not miss, its failing for the first commit still.. is there a way to fix or should i re-create the pr?

controllers/storageclient_controller.go
bernerhat: its only referring to job scheduling failure, with the deadline (to finish) in place there wont be a skipped scheduling more than twice per iteration (at most). once the 155 seconds deadline is reached a new iteration will spawn and the counter for how many missed jobs occurred will be reset. there is still the option that the CronJob controller happens to be down for a long time (more than 100 minutes) which will cause the issue you are referring to but wouldn't that be considered a cluster issue at this point? i can add the field but i'm unsure if it will cause issues with the other fields i've added and will need additional testing

leelavg: It's not mandatory, the reasoning is good enough.

leelavg (approved):

PR #40, opened 2023-11-15
Bug 2247735: Adding toleration to schedule CSI pods on master nodes

rchikatw:
Adding toleration to schedule CSI pods on master nodes

### DIFF
--- a/pkg/utils/k8sutils.go
+++ b/pkg/utils/k8sutils.go
@@ -36,6 +36,8 @@ const StorageClientNamespaceEnvVar = "STORAGE_CLIENT_NAMESPACE"
 
 const StatusReporterImageEnvVar = "STATUS_REPORTER_IMAGE"
 
+const runCSIDaemonsetOnMaster = "RUN_CSI_DAEMONSET_ON_MASTER"
+
 // GetOperatorNamespace returns the namespace where the operator is deployed.
 func GetOperatorNamespace() string {
 	return os.Getenv(OperatorNamespaceEnvVar)
--- a/pkg/utils/placements.go
+++ b/pkg/utils/placements.go
@@ -1,0 +1,32 @@
+package utils
+
+import (
+	"log"
+	"os"
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func GetTolerationForCSIPods() corev1.Toleration {
+
+	runOnMaster := true
+	var err error
+	rom := os.Getenv(runCSIDaemonsetOnMaster)
+	if rom != "" {
+		runOnMaster, err = strconv.ParseBool(rom)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	if runOnMaster {
+		toleration := corev1.Toleration{
+			Key:      "node-role.kubernetes.io/master",
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoSchedule,
+		}
+		return toleration
+	}
+	return corev1.Toleration{}
+}
--- a/pkg/csi/cephfsdaemonset.go
+++ b/pkg/csi/cephfsdaemonset.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/red-hat-storage/ocs-client-operator/pkg/templates"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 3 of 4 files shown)


pkg/utils/k8sutils.go
```diff
@@ -36,6 +36,8 @@ const StorageClientNamespaceEnvVar = "STORAGE_CLIENT_NAMESPACE"
 
 const StatusReporterImageEnvVar = "STATUS_REPORTER_IMAGE"
 
+const CSIDaemonSetsTolerateAllNodes = "CSI_DaemonSets_Tolerate_All_Nodes"
```
Madhu-1, line 39:
```suggestion
const runCSIDaemonsetOnMaster = "RUN_CSI_DAEMONSET_ON_MASTER"
```

pkg/utils/placements.go
```diff
@@ -0,0 +1,27 @@
...
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func GetTolerationForCSIPods() corev1.Toleration {
+	enableCSIPodSchedulingOnMaster, err := strconv.ParseBool(os.Getenv(CSIDaemonSetsTolerateAllNodes))
+	if err != nil {
+		log.Fatal(err)
+	}
```
Madhu-1, line 15:
```suggestion
// Run on master by default
runOnMaster :=true
var err error
rom:=os.Getenv(runCSIDaemonsetOnMaster))
if rom!=""{
runOnMaster, err = strconv.ParseBool(rom)
	if err != nil {
		log.Fatal(err)
	}
}

if runOnMaster{
	toleration := corev1.Toleration{
			Key:      "node-role.kubernetes.io/master",
			Operator: corev1.TolerationOpExists,
			Value:    "true",
			Effect:   corev1.TaintEffectNoSchedule,
		}
		return toleration
	]
}
```

pkg/utils/placements.go
```diff
@@ -0,0 +1,32 @@
...
+	rom := os.Getenv(runCSIDaemonsetOnMaster)
+	if rom != "" {
+		runOnMaster, err = strconv.ParseBool(rom)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	if runOnMaster {
+		toleration := corev1.Toleration{
+			Key:      "node-role.kubernetes.io/master",
```
Madhu-1, line 24: it looks like this is deprecated and replaced with control-plane https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint, please check what we have in supported OCP versions

pkg/utils/placements.go
```diff
@@ -0,0 +1,32 @@
...
+			Operator: corev1.TolerationOpEqual,
+			Value:    "true",
```
Madhu-1, line 26: value is not required as we will not have it set for the master node and also the Operator should be exists not equal. please check https://docs.openshift.com/container-platform/4.14/nodes/scheduling/nodes-scheduler-taints-tolerations.html

Madhu-1 (commented):
please test and change and see if its working as expected

leelavg:
@Madhu-1 and @rchikatw 
- I'm wondering why not `runCSIDaemonSetOnAllNodes` and use `Toleration{Operator: TolerationExists}`

Reasoning being,
- no need to find and set the `Toleration.Key` explicitly.
- Only if workloads are scheduled on tainted nodes then only CSI pods does actual work (this is a con as well, they are always there consuming resources)
- let the default be false (or not set) to conform to existing deployments and take the input from Subscription `.spec.config.env` (or what is the planned deployment model, iirc env will be reset if not set in CSV?)

Madhu-1:
> @Madhu-1 and @rchikatw
> 
> * I'm wondering why not `runCSIDaemonSetOnAllNodes` and use `Toleration{Operator: TolerationExists}`
> 

The requirement was to run daemonset pods also on master node but not the nodes which might have tainted for something else.

> Reasoning being,
> 
> * no need to find and set the `Toleration.Key` explicitly.
> * Only if workloads are scheduled on tainted nodes then only CSI pods does actual work (this is a con as well, they are always there consuming resources)
> * let the default be false (or not set) to conform to existing deployments and take the input from Subscription `.spec.config.env` (or what is the planned deployment model, iirc env will be reset if not set in CSV?)

👍🏻  default is `true` but it need to be added to env to avoid upgrade problem

pkg/utils/placements.go
rchikatw: checked one of the fusion hci clusters the key was "node-role.kubernetes.io/master". we can go with this key

leelavg:
@rchikatw could you pls fix commitlint https://github.com/red-hat-storage/ocs-client-operator/actions/runs/6930384492/job/18849951078#step:4:25?

leelavg (approved):

PR #46, opened 2023-12-11
supply client operator version while onboarding

leelavg:
provider server makes it mandatory to  supply client operator version for successful onboarding and so we'll send the client operator version as part of onboarding process

[RHSTOR-5072]
depends on: red-hat-storage/ocs-operator#2308

PR #47, opened 2023-12-18
[RHSTOR-5205] Lock client operator upgrade if it current matches provider operator version

leelavg:
- Provider operator sends it's version as a response to the heartbeat mechanism
- Client operator should take that into account and place a not upgradeable condition if it currently matches provider operator version
- Failing to do above may result in client getting ahead of provider and interfere with communications among others
- Conversely, unlock the upgrades if we are currently not matching provider operator version

rebased on #46 and functionally dependent on red-hat-storage/ocs-operator#2329


leelavg:
superseded by #53

PR #53, opened 2024-01-04
new controller for managing upgrade of operator

leelavg:
As part of heartbeat response provider sends desired version that it wants client operator to match and this PR introduces a new (upgrade) controller to take action based on desired version.


controllers/upgrade_controller.go
```diff
@@ -0,0 +1,270 @@
...
+	r.ctx = ctx
+	r.log = log.FromContext(ctx)
+	r.log.Info("starting reconcile")
+
+	conditionStatus := metav1.ConditionTrue
+	conditionMessage := "No component reported errors."
+	result, err := r.reconcilePhases()
+	if err != nil {
+		// TODO (lgangava): we are being very restrictive as of now, not upgrading if we hit any error
+		// differentiate between reconciler errors and not upgradeable errors
```
raaizik, line 125: Don't those coincide? Meaning surely there are some reconciling errors that imply non-upgradability

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,270 @@
...
+	if err := r.list(storageClientList); err != nil {
+		return fmt.Errorf("failed to list storageclients %v", err)
+	}
+
+	if len(storageClientList.Items) == 0 {
+		r.log.Info("no storageclients exist")
+		return nil
+	}
+
+	desiredVersions := make([]string, 0, len(storageClientList.Items))
```
raaizik, line 154: Better to declare as a nil slice with `var`

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,270 @@
...
+	for idx := range storageClientList.Items {
+		desiredVersion := storageClientList.Items[idx].Status.Operator.DesiredVersion
+		// usually these versions will not have patch version, so we are adding an arbitrary num for semver comparison
+		desiredVersion += dummyPatchVersion
```
raaizik, line 158: Doesn't not having a patch version translate to `_._.0` and not some arbitrary number?

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,270 @@
...
+	channelVersion, err := getVersionFromChannel(channel)
+	if err != nil {
+		return fmt.Errorf("failed to convert channel name %q to semver %v", channel, err)
+	}
+
+	// decide whether to patch the subscription or not
+	platformVersion, _ := semver.Make(r.PlatformVersion)
+	if desiredVersion.Major > platformVersion.Major || desiredVersion.Minor > platformVersion.Minor {
+		r.log.Info("backing off from updating subscription as operator version can become ahead of platform version")
+	} else if desiredVersion.Major == channelVersion.Major && desiredVersion.Minor == channelVersion.Minor+1 {
```
raaizik, line 190: Wondering is this the only possible scenario where the minor versions differ?

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,270 @@
...
+func (r *UpgradeReconciler) list(list client.ObjectList, opts ...client.ListOption) error {
+	return r.Client.List(r.ctx, list, opts...)
+}
+
+func (r *UpgradeReconciler) patch(obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+	return r.Client.Patch(r.ctx, obj, patch, opts...)
+}
+
+func getOldestVersion(versions []string) (semver.Version, error) {
+	if len(versions) < 1 {
```
raaizik, line 242: Missing a preceding nil check (which could have been more straightforward without using `make` upon init)

controllers/upgrade_controller.go
```diff
@@ -0,0 +1,270 @@
...
+// returns the name of the subscription channel corresponding to the supplied operator version
+func getChannelFromVersion(version semver.Version) string {
+	return fmt.Sprintf("%s%d.%d", subscriptionChannelNamePrefix, version.Major, version.Minor)
+}
+
+// returns the semver version of operator version corresponding to the subscription channel name
+func getVersionFromChannel(name string) (semver.Version, error) {
+	version, _ := strings.CutPrefix(name, subscriptionChannelNamePrefix)
+	// usually channel names will not have patch version and semver will fail w/o patch version
+	version += dummyPatchVersion
```
raaizik, line 268: See above (`_._.0`)

controllers/storageclient_controller.go
```diff
@@ -445,6 +440,26 @@ func (s *StorageClientReconciler) delete(obj client.Object) error {
 	return nil
 }
 
+func (s *StorageClientReconciler) reconcileClientStatus(instance *v1alpha1.StorageClient) (reconcile.Result, error) {
+	currentVersion := instance.Status.Operator.CurrentVersion
+	operatorVersion, _ := semver.Make(s.OperatorVersion)
+	if currentVersion == "" {
+		// TODO(lgangava): should we set Major.Minor.Patch or Major.Minor only
```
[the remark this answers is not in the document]
leelavg: this is visible to user, desiredVersion will only be M.m and I'm setting currentVersion also as M.m and this TODO is to think of any other reasons that it should be M.m.p which I couldn't find.

controllers/suite_test.go
```diff
@@ -35,41 +47,207 @@ import (
...
+	testSubscriptionName = "client-sub"
 
-	By("bootstrapping test environment")
-	testEnv = &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
-		ErrorIfCRDPathMissing: false,
-	}
+	testAheadVersion   = "4.15"
+	testCurrentVersion = "4.14"
+	testLaggingVersion = "4.13"
```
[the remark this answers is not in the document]
leelavg: although it isn't urgent, having major version checks wherever possible also makes sense, will try to push in next revision.

controllers/upgrade_controller.go
leelavg: > there are some reconciling errors that imply non-upgradability
- currently we are implying all reconcile errors as non-upgradability
- do you mean to say, not all reconcile errors shouldn't be viewed as not upgradable?

controllers/upgrade_controller.go
leelavg: nope, if we know the capacity why to declare without it?

controllers/upgrade_controller.go
leelavg: nope, `0` is as good as any other random number and we just need a random number.
the background is:
- provider sends desiredversion and it'll have only Major.minor
- client queries it's version and sets currentVersion, also Major.minor
- but semver is strict about version be Major.minor.patch and for comparison we just need a constant random number added to it.

why not `0` then?
- if this code changes and it's logged, M.m.0 will actually be a release version but the release to be M.m.99 is slim to none so I chose 99

controllers/upgrade_controller.go
leelavg: No, this is the only scenario we are interested and wanted to take an action on. There's a corner case which I don't want to include which will complicate it.

4.20 -> released
4.21 -> released but whole minor version is faulty and redacted
4.22 -> released, now if desired is 22 but channel is at 20 this branch will not be taken

however, with an entire minor version being faulty we might have more serious issues that this.

do you mean any other scenarios?

controllers/upgrade_controller.go
leelavg: answered https://github.com/red-hat-storage/ocs-client-operator/pull/53#discussion_r1442492757

controllers/upgrade_controller.go
leelavg: nope, `len` is an all rounder, it's always better to use that for checking emptiness on items
``` go
> var first []string
: 0xc00011c0e0
> var second []string = nil
: 0xc0004040f0
> third := []string{}
: []
> fourth := make([]string, 0, 5)
: []
> fifth := make([]string, 5)
: [    ]
> fmt.Println(len(first)==0)
true
: 5
> fmt.Println(len(second)==0)
true
: 5
> fmt.Println(len(third)==0)
true
: 5
> fmt.Println(len(fourth)==0)
true
: 5
> fmt.Println(len(fifth)==0)
false
: 6
```

leelavg:
depends on red-hat-storage/ocs-operator#2329

leelavg:
not be revived

PR #60, opened 2024-01-10
new controller for restricting upgrade of operator

leelavg:
As part of heartbeat response provider sends desired version that it wants client operator to match and this PR introduces a new (upgrade) controller to stop upgrades if the current version is ahead of desired version.


leelavg:
Depends on red-hat-storage/ocs-operator#2329

leelavg:
not be revived

PR #61, opened 2024-01-22
csi: use static name for cephfs/rbd

Madhu-1:
use the static provisioner name for both cephfs and rbd csi driver irrespective of the namespace where they are deployed.


Madhu-1:
@nb-ohad PTAL

Madhu-1:
@nb-ohad this also needs `/approve` label as well. can you please help with that?

PR #64, opened 2024-01-25
Redesign for ClusterVersion controller into OperatorConfigMap controller

bernerhat:
Fixes the issue created where resources are leftover while trying to uninstall the odf client operator.

While trying to uninstall the operator clusterVersion resource, which is managed by OCP, remains in the cluster hence is not being deleted and causing the controller owned resources to remain as well.
Therefore a need arose to change the 'watched' resource from the clusterVersion into another resource who is more appropriate and in this case the ConfigMap of the operator was chosen.


controllers/operatorconfigmap_controller.go
```diff
@@ -91,19 +93,19 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 				},
 			}}
 		},
 	)
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&configv1.ClusterVersion{}, clusterVersionPredicates).
-		Watches(&corev1.ConfigMap{}, enqueueClusterVersionRequest, configMapPredicates).
+		For(&corev1.ConfigMap{}, configMapPredicates).
+		Watches(&configv1.ClusterVersion{}, enqueueConfigMapRequest, clusterVersionPredicates).
```
bernerhat, line 108: not entirely sure if we still need to watch the ClusterVersion resource, was added to be on the safe side.

Madhu-1: JFYI: ClusterVersion watcher was added to ensure we update the csi sidecar images as per the ocp cluster version where its running.

nb-ohad: @bernerhat Yes we do, we need to peek changes to the cluster version as an additional input and to issue reconciles to make sure we are still running to correct version of the csi sidecar, as @Madhu-1 mentioned in his comment

controllers/operatorconfigmap_controller.go
```diff
@@ -317,10 +322,35 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return nil
+		}
+		return err
+	}
+	csiAddonsNode.Finalizers = nil
+	if err := c.Client.Update(c.ctx, csiAddonsNode); err != nil {
+		c.log.Error(err, "Failed to remove finalizer from csiAddonsNode")
+		return err
+	}
+	return nil
```
nb-ohad, line 351: This is not the established way to remove finalizers. 
Please take a look at this code snippet from another repo we are working on: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L508-L513

nb-ohad (changes requested):

controllers/operatorconfigmap_controller.go
```diff
@@ -317,10 +322,35 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	c.log.Info("prometheus rules deployed", "prometheusRule", klog.KRef(prometheusRule.Namespace, prometheusRule.Name))
 
+	// deletion phase
+	if !instance.GetDeletionTimestamp().IsZero() {
```
leelavg, line 326: I don't see a finalizer on configmap, so I think we might get an event even if this cm was just recreated 🤔

leelavg (commented):
- pls follow `if err := func(); err != nil {}` convention if returning directly based on error

> While trying to uninstall the operator clusterVersion resource, which is managed by OCP
- I don't think so, no resource is being owned by clusterversion resource
- Actually bug fix should just be moving the DeepCopy line before `own`ing it and removing finalizer on CSIAddonNode
----
- unless we really don't want to watch for a delete event on configmap I don't think we need a redesign

@Madhu-1 pls correct me if my understand of existing code and the fix is wrong.

controllers/operatorconfigmap_controller.go
```diff
@@ -210,10 +219,10 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		},
 	}
 	err = c.createOrUpdate(c.cephFSDeployment, func() error {
+		csi.GetCephFSDeployment(c.OperatorNamespace).DeepCopyInto(c.cephFSDeployment)
```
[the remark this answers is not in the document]
Madhu-1: @bernerhat any specific reason to move this here, please put this kind of changes to its own commit so that we can understand what is the bug

@nb-ohad i think it was done to ensure that we update the deployment Template (for new changes)/update

controllers/operatorconfigmap_controller.go
Madhu-1: This is the user created configmap are we creating it now?

controllers/operatorconfigmap_controller.go
Madhu-1: AFAIC this can lead to unexpected errors as we are not the ones creating the `CSIAddonsNode`, I think this is happening due to the way the uninstall is happening but this is not the right component to fix this.

Madhu-1:
> * pls follow `if err := func(); err != nil {}` convention if returning directly based on error
> 
> > While trying to uninstall the operator clusterVersion resource, which is managed by OCP
> 
> * I don't think so, no resource is being owned by clusterversion resource
> * Actually bug fix should just be moving the DeepCopy line before `own`ing it and removing finalizer on CSIAddonNode
> 
> * unless we really don't want to watch for a delete event on configmap I don't think we need a redesign
> 
> @Madhu-1 pls correct me if my understand of existing code and the fix is wrong.

@leelavg are we trying to cleanup this things as part of uninstall? i mentioned about `DeepCopy` and CSIADDONS in my earlier comments. watching configmap is correct (i think) but there are many other changes in this PR)

controllers/operatorconfigmap_controller.go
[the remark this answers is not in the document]
nb-ohad: @Madhu-1 That does not make a lot of sense. 
`GetCephFSDeployment ` already creates a copy of the top-level structure from the template. If the concern is that somewhere in the nested structure there is a pointer to a shared sub-struct then I would suggest that `GetCephFSDeployment` should always create a deep copy before applying any logic and before returning it. 

With the current code, it seems that we are calling `deepCopy` every time `GetCephFSDeployment` is invoked which, to me, indicates that it should be part of the `GetCephFSDeployment` operation itself.

controllers/operatorconfigmap_controller.go
nb-ohad: @Madhu-1 where should the fix go?

controllers/operatorconfigmap_controller.go
nb-ohad: Yes, we should be the ones to create an empty one. After this change, the config map becomes the root resource for this controller as it is tied to the namespace lifespan. 

But, even with that in mind, I don't see why so we need a finalizer. A finalizer is used if you must keep a resource alive to complete nay cleanup operation, and I don't believe this is the case here

controllers/operatorconfigmap_controller.go
bernerhat: ack

controllers/operatorconfigmap_controller.go
Madhu-1: it should go to the component `csi-addons` but that needs to be discussed with the team first. Do we have a BZ for it?

nb-ohad: This entire PR is a bug fix. See the bug number int the PR title

controllers/operatorconfigmap_controller.go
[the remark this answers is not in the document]
bernerhat: @Madhu-1 this code was moved as a suggestion from @leelavg . we suspect the `deepCopy` is messing the ownership assignment (right below it) and causing the leftover issue for these components

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,454 @@
...
+				return ((namespace == c.OperatorNamespace) && (name == operatorConfigMapName))
+			},
+		),
+	)
+	// Reconcile the ClusterVersion object when the operator config map is updated
+	enqueueConfigMapRequest := handler.EnqueueRequestsFromMapFunc(
+		func(_ context.Context, client client.Object) []reconcile.Request {
+			return []reconcile.Request{{
+				NamespacedName: types.NamespacedName{
+					Name: operatorConfigMapName,
```
nb-ohad, line 101: A config map is  a namespaced resource you need to provide a namespace as well

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,454 @@
...
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
+func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var err error
+	c.ctx = ctx
+	c.log = log.FromContext(ctx, "OperatorConfigMap", req)
+	c.log.Info("Reconciling OperatorConfigMap")
+
+	c.operatorConfigMap = &corev1.ConfigMap{}
+	if err = c.Client.Get(context.TODO(), req.NamespacedName, c.operatorConfigMap); err != nil {
```
nb-ohad, line 136: Why `context.TODO` and not `c.ctx`?

controllers/operatorconfigmap_controller.go
nb-ohad: @bernerhat I agree with @Madhu-1 that the code that adds the finalizer should remove the finalizer. 
Are we even sure that with our ownership changes, there is still an issue with the removal of these finalizers? Are we making sure to add resource ownership to all the resources this reconciler reconciles?

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,454 @@
...
+	result, err := controllerutil.CreateOrUpdate(c.ctx, c.Client, obj, f)
+	if err != nil {
+		return err
+	}
+	c.log.Info("successfully created or updated", "operation", result, "name", obj.GetName())
+	return nil
+}
+
+func (c *OperatorConfigMapReconciler) own(obj client.Object) error {
+	return controllerutil.SetControllerReference(c.OperatorDeployment, obj, c.Client.Scheme())
```
nb-ohad, line 358: I think this is wrong. It `r.Own` should set the config map as the owning resource. It might be the exact issue we see with some resources no being deleted when the config map is deleted.

nb-ohad (changes requested):

controllers/operatorconfigmap_controller.go
bernerhat: I agree as well, it makes a lot more sense, I had some issues trying out the code so i did not make sure yet if the ownership changes solves the finalizer issue as well.
I believe that we do add ownership to all other resources reconciled here.

controllers/operatorconfigmap_controller.go
bernerhat: was actually a leftover from the previous logic that was here, fixed.

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,454 @@
+/*
```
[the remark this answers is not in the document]
bernerhat: i had to rename the file since it is no longer a clusterVersion controller, was i suppose to do something differently?

controllers/operatorconfigmap_controller.go
nb-ohad: No, as `client.GetNamespace()` will get the client of the watched resource. What you want is the operator namespace. Please look elsewhere in the code to how we obtain it. If not here try OCS operator.

controllers/operatorconfigmap_controller.go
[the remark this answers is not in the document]
nb-ohad: This does not look like a rename from git perspective. It looks like a new file and it presents issues in reviewing as we cannot differentiate between changes to the code and old code. I am not sure what you did to cause that. A simple FS rename locally on your side should have been caught by git as a rename and not a deletion of the old file and the creation of a different file.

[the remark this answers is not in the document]
bernerhat: i think the diff was too large before and somehow made git think it was a different file, i hadn't done anything other than a rename.
Anyway now i think it does recognize the file. does it look ok now @leelavg ?

controllers/operatorconfigmap_controller.go
bernerhat: Nice catch! yes certainly needs a fix, done.

controllers/operatorconfigmap_controller.go
leelavg: > Yes, we should be the ones to create an empty one
- should this be part of bundle?

> A finalizer is used if you must keep a resource alive to complete nay cleanup operation
- yes and at the same time we are tying the deletion of config map as deletion of csi deployments & ds
- it might so happen someone mistook the config map and deleted it and going by current logic we'll cleanup csi as well
- I know finaliser on configmap doesn't solve deletion of csi and my point only is, deletion and creation of this configmap deletes and creates csi w/o finalizer, are we aiming for it?

---

regarding configmap name:
- I hope all reviewers is aware that this configmap doesn't exist as of now and what is the suggestion on how to create it, via bundle? another PR?
- what we deploy as part of bundle is `ocs-client-operator-manager-config` cm ([ref](https://github.com/red-hat-storage/ocs-client-operator/blob/main/config/manager/kustomization.yaml#L10)) and do we want to create a new `ocs-client-operator-config` cm ([ref](https://github.com/red-hat-storage/ocs-client-operator/blob/main/controllers/clusterversion_controller.go#L54))

controllers/operatorconfigmap_controller.go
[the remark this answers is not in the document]
leelavg: 1. I couldn't understand why the DeepCopyInto exists in the mutate func in the first place, we could just update the desired spec and call it a day unless I'm missing these deployments/ds need to ensure some more which I don't know.
2. Yes I suggested to move it here and verify the bug still occurs or not.

----

I'm rephrasing what I already mentioned in https://github.com/red-hat-storage/ocs-client-operator/pull/64#pullrequestreview-1848015033

- the actual bug is two fold, one, csi deployments & ds are not getting deleted, two, CSIAddonNode finalizer is not removed & garbage collected
- for first bug, `own`ing resource after deep copy should be the fix (I asked Amit to verify this some time back)
- for second bug, we aren't the owner of CSIAddonNode, this morning I saw Madhu updated the BZ to delete csi-addons after deleting ocs-client-operator and that should be tested along w/ above (moving line)

that's it, we have a bug and an unverified fix w/ a possible root cause.

---

on redesign:

- I rightfully agree that configmap should back this controller but a bit sceptical whether to club that in this PR or not unless my suggestion above is wrong which mandates a redesign
- commented on configmap [here](https://github.com/red-hat-storage/ocs-client-operator/pull/64#discussion_r1471075608)

controllers/operatorconfigmap_controller.go
```diff
@@ -91,19 +92,20 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	)
 	// Reconcile the ClusterVersion object when the operator config map is updated
-	enqueueClusterVersionRequest := handler.EnqueueRequestsFromMapFunc(
+	enqueueConfigMapRequest := handler.EnqueueRequestsFromMapFunc(
 		func(_ context.Context, client client.Object) []reconcile.Request {
 			return []reconcile.Request{{
 				NamespacedName: types.NamespacedName{
-					Name: clusterVersionName,
+					Name:      operatorConfigMapName,
+					Namespace: client.GetNamespace(),
```
nb-ohad, line 100: @bernerhat As I already explined in a prev comment here: https://github.com/red-hat-storage/ocs-client-operator/pull/64#discussion_r1470035357

using `client.GetNamespace` is wrong! client in this case is the object that produces the event. In our use case, this will be the ClusterVersion CR which is a cluster-wide CR and has no namespace associated with it. But even if it was a namespaced CR, the namespace most likely would be different.  

You need to use the operator's namespace not the object's namespace

controllers/operatorconfigmap_controller.go
```diff
@@ -197,8 +204,7 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err := c.own(monConfigMap); err != nil {
```
nb-ohad, line 204: I beleive this is a copy-paste mistake.  
```suggestion
	if err := c.own(encConfigMap); err != nil {
```

controllers/operatorconfigmap_controller.go
```diff
@@ -279,35 +285,28 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// ownerReference on it as its cluster scoped resource
 	cephfsCSIDriver := templates.CephFSCSIDriver.DeepCopy()
 	cephfsCSIDriver.ObjectMeta.Name = csi.GetCephFSDriverName()
-	err = csi.CreateCSIDriver(c.ctx, c.Client, cephfsCSIDriver)
-	if err != nil {
+	if err := csi.CreateCSIDriver(c.ctx, c.Client, cephfsCSIDriver); err != nil {
```
nb-ohad, line 288: @bernerhat @Madhu-1 I think something is missing here. 
I don't see a parameter that will allow us to add an owner ref on the CSI driver deployment! 
This means that the CSI driver deployment will be detached from the ownership chain, that in turn will create issues for k8s to follow a proper deletion process

controllers/operatorconfigmap_controller.go
nb-ohad: @bernerhat Please don't resolve any conversation until you get an ack that the reviewer who issued the comment is ok with the fix. You have 2 options:
1. if the reviewer is a maintainer and can resolve the comments themselves then let them do it.
2. If they are not, wait for an ack on the comment thread and only then resolve.

controllers/operatorconfigmap_controller.go
Madhu-1: https://github.com/rook/rook/blob/3b1a4286885a78737e597945cc20826a6c3137cb/pkg/operator/ceph/csi/csidriver.go#L108-L116 talks about csidriver delete logic , which can be reused here

leelavg:
all my comments can be resolved and pls remove bug id from the title and commit, bugs are mentioned only during backports. you can mention the id in description or as a comment though.

controllers/operatorconfigmap_controller.go
nb-ohad: Discussed offline. Agreed that we need to have code in the operator to Delete/Remove the CSI driver as we cannot own cluster scoped resources and the removal is not just deletion of a resource

controllers/operatorconfigmap_controller.go
bernerhat: changed into `c.OperatorNamespace` is that acceptable?

controllers/operatorconfigmap_controller.go
bernerhat: yes, changed. can resolve?

controllers/operatorconfigmap_controller.go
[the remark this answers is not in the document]
bernerhat: @nb-ohad discussed offline with @Madhu-1 and decided to move the `deepCopy` into each ceph get function

controllers/operatorconfigmap_controller.go
```diff
@@ -123,204 +124,196 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-
-	instance := configv1.ClusterVersion{}
-	if err = c.Client.Get(context.TODO(), req.NamespacedName, &instance); err != nil {
-		return ctrl.Result{}, err
-	}
+	// deletion phase
+	if !c.OperatorConfigMap.GetDeletionTimestamp().IsZero() {
+		if err := c.deletionPhase(); err != nil {
+			return ctrl.Result{}, err
+		}
```
leelavg, line 137: I don't think this block will ever trigger, did you observe this in your testing? The reconciler is storing the configmap once at the program start and not refreshing after that so on the resource that's stored in the reconciler the deletion time stamp will never be set.

controllers/operatorconfigmap_controller.go
```diff
@@ -51,17 +51,17 @@ import (
...
 )
 
-// ClusterVersionReconciler reconciles a ClusterVersion object
-type ClusterVersionReconciler struct {
+// OperatorConfigMapReconciler reconciles a ClusterVersion object
+type OperatorConfigMapReconciler struct {
 	client.Client
 	OperatorDeployment *appsv1.Deployment
 	OperatorNamespace  string
+	OperatorConfigMap  *corev1.ConfigMap
```
leelavg, line 64: do not store the configmap here, let's deploy configmap as part of the bundle and fetch it in every reconcile.

controllers/operatorconfigmap_controller.go
```diff
@@ -123,204 +124,196 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	instance := configv1.ClusterVersion{}
-	if err = c.Client.Get(context.TODO(), req.NamespacedName, &instance); err != nil {
-		return ctrl.Result{}, err
-	}
+	// deletion phase
+	if !c.OperatorConfigMap.GetDeletionTimestamp().IsZero() {
+		if err := c.deletionPhase(); err != nil {
+			return ctrl.Result{}, err
+		}
+	} else {
```
leelavg, line 138: we need to add a finalizer first and then proceed with rest of the operations.

controllers/operatorconfigmap_controller.go
leelavg, line 137: we need to remove the finalizer if the deletionphase succeeds

controllers/operatorconfigmap_controller.go
```diff
@@ -123,204 +124,196 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return ctrl.Result{}, err
+		}
 
+		c.log.Info("prometheus rules deployed", "prometheusRule", klog.KRef(prometheusRule.Namespace, prometheusRule.Name))
+	}
 	return ctrl.Result{}, nil
 }
 
-func (c *ClusterVersionReconciler) createOrUpdate(obj client.Object, f controllerutil.MutateFn) error {
+func (c *OperatorConfigMapReconciler) deletionPhase() error {
```
leelavg, line 298: we need to proceed only if no PVs exist in the cluster using our provisioner name

controllers/operatorconfigmap_controller.go
leelavg, line 1: in general I still see a lot of diff which shouldn't be the case as we are mostly moving the lines.

leelavg (changes requested):

pkg/csi/cephfsdaemonset.go
```diff
@@ -270,4 +270,7 @@ func GetCephFSDaemonSet(namespace string) *appsv1.DaemonSet {
 			},
 		},
 	}
+	cephfsDaemonset := &appsv1.DaemonSet{}
+	daemonset.DeepCopyInto(cephfsDaemonset)
+	return cephfsDaemonset
```
leelavg, line 275: why do we need deepcopy methods here?

leelavg:
/retitle Redesign for ClusterVersion controller into OperatorConfigMap controller

controllers/operatorconfigmap_controller.go
leelavg: answering my own question. We can't make git track this file as a rename as the default similarity should be 50% but as of now the file is around ~42% similar. for viewing only edited changes, pull this pr and view `git status -M40%`

comment can be resolved.

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+//go:embed pvc-rules.yaml
+var pvcPrometheusRules string
+
+const (
+	operatorConfigMapName = "ocs-client-operator-manager-config"
+	// ClusterVersionName is the name of the ClusterVersion object in the
+	// openshift cluster.
+	clusterVersionName = "version"
+
+	OperatorConfigMapFinalizer = "operatorcm.ocs.openshift.io"
```
leelavg, line 61: `operatorcm.ocs.openshift.io` doesn't sound better as `operatorcm` seems generic, maybe exact one `ocs-client-operator.ocs.openshift.io` might be better, usually these names take after their CRs (but i believe storageclient & storageclassclaim is already taken by their corresponding controllers)

I don't think we need to export the constant, don't capitalize the var

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+)
+
+// OperatorConfigMapReconciler reconciles a ClusterVersion object
+type OperatorConfigMapReconciler struct {
+	client.Client
+	OperatorDeployment *appsv1.Deployment
```
leelavg, line 67: do we still need this?

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	configMapPredicates := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(client client.Object) bool {
+				namespace := client.GetNamespace()
+				name := client.GetName()
+				return ((namespace == c.OperatorNamespace) && (name == operatorConfigMapName))
+			},
+		),
+	)
+	// Reconcile the ClusterVersion object when the operator config map is updated
```
leelavg, line 97: change the comment to reflect the op

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+//+kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="apps",resources=deployments/finalizers,verbs=update
+//+kubebuilder:rbac:groups="apps",resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="apps",resources=daemonsets/finalizers,verbs=update
+//+kubebuilder:rbac:groups="storage.k8s.io",resources=csidrivers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;list;watch;create;patch;update
+//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheusrules,verbs=get;list;watch;create;update
+//+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=*
```
leelavg, line 127: just realized, many are high privileged rbac, pls raise an issue to revisit this at a later point

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
+func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var err error
+	c.ctx = ctx
+	c.log = log.FromContext(ctx, "OperatorConfigMap", req)
+	c.log.Info("Reconciling OperatorConfigMap")
+
+	OperatorConfigMap := &corev1.ConfigMap{}
```
leelavg, line 137: why is this capitalized?

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// deletion phase
+	if !OperatorConfigMap.GetDeletionTimestamp().IsZero() {
+		if err := c.deletionPhase(); err != nil {
+			return ctrl.Result{}, err
+		}
+		//remove finalizer
+		OperatorConfigMap.SetFinalizers(remove(OperatorConfigMap.GetFinalizers(), OperatorConfigMapFinalizer))
```
leelavg, line 149: pls use controllerutil for setting and removing finalizers.

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+		return err
+	}
+	if err := c.Client.Delete(c.ctx, c.scc); err != nil && !k8serrors.IsNotFound(err) {
+		c.log.Error(err, "unable to delete SCC")
+		return err
+	}
+	return nil
+}
+
+func (c *OperatorConfigMapReconciler) pvsPresent() error {
```
leelavg, line 344: the signature is wrong, pls don't club error and negative outcome.

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+			return ctrl.Result{}, err
+		}
+
+		c.log.Info("prometheus rules deployed", "prometheusRule", klog.KRef(prometheusRule.Namespace, prometheusRule.Name))
+	}
+	return ctrl.Result{}, nil
+}
+
+func (c *OperatorConfigMapReconciler) deletionPhase() error {
+	if err := c.pvsPresent(); err != nil {
```
leelavg, line 323: we aren't interested on all pvs as the func name suggests, `hasOCSVolumes` isn't a bad name.

controllers/operatorconfigmap_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	if err := c.Client.List(c.ctx, &storageClassList); err != nil {
+		return err
+	}
+
+	// create a set of storage class names who are using the ocs-client provisioner
+	// provisioner name prefixed with the namespace name
+	ocsClientStorageClass := make(map[string]bool)
+	for i := range storageClassList.Items {
+		storageClass := &storageClassList.Items[i]
+		if strings.HasPrefix(storageClassList.Items[i].Provisioner, c.OperatorNamespace) {
```
leelavg, line 356: the sample code provided in deployer mandates the namespace to be `openshift-storage` and it was correct there but not in here.

remove the prefix func and directly check for these matches
https://github.com/red-hat-storage/ocs-client-operator/blob/18a6a733dff7015d0415dba12061bd467efdff03/pkg/csi/csi.go#L93-L101

these are static funcs, compiler would mostly inline them and no need to worry about storing them again before the loop.

leelavg (changes requested):

controllers/operatorconfigmap_controller.go
```diff
@@ -123,23 +128,54 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
-func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var err error
 	c.ctx = ctx
-	c.log = log.FromContext(ctx, "ClusterVersion", req)
-	c.log.Info("Reconciling ClusterVersion")
+	c.log = log.FromContext(ctx, "OperatorConfigMap", req)
```
leelavg, line 134: recently I observed that name of reconciler and req is already part of `ctx` and this line is adding them to the log again, could you pls check and change this to only `c.log = log.FromContext(ctx)`

controllers/operatorconfigmap_controller.go
```diff
@@ -194,80 +229,55 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		},
-	}
 	err = c.createOrUpdate(c.cephFSDeployment, func() error {
-		if err := c.own(c.cephFSDeployment); err != nil {
+		c.cephFSDeployment = csi.GetCephFSDeployment(c.OperatorNamespace)
+		if err := c.own(c.cephFSDeployment, operatorConfigMap); err != nil {
 			return err
 		}
-		csi.GetCephFSDeployment(c.OperatorNamespace).DeepCopyInto(c.cephFSDeployment)
 		return nil
```
leelavg, line 245: not sure why golint didn't suggest this, just `return c.own(c.cephFSDeployment, operatorConfigMap)` same for remaining.

controllers/operatorconfigmap_controller.go
```diff
@@ -329,11 +389,11 @@ func (c *ClusterVersionReconciler) createOrUpdate(obj client.Object, f controlle
 	return nil
 }
 
-func (c *ClusterVersionReconciler) own(obj client.Object) error {
-	return controllerutil.SetControllerReference(c.OperatorDeployment, obj, c.Client.Scheme())
+func (c *OperatorConfigMapReconciler) own(obj client.Object, operatorConfigMap *corev1.ConfigMap) error {
```
leelavg, line 392:
```suggestion
func (c *OperatorConfigMapReconciler) own(obj client.Object, owner *corev1.ConfigMap) error {
```

controllers/operatorconfigmap_controller.go
```diff
@@ -320,7 +323,64 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	storageClassList := storagev1.StorageClassList{}
+	if err := c.Client.List(c.ctx, &storageClassList); err != nil {
+		return false, fmt.Errorf("unable to list storage classes: %v", err)
+	}
+
+	// create a set of storage class names who are using the client's provisioners
+	ocsStorageClass := make(map[string]bool)
+	for i := range storageClassList.Items {
+		storageClass := &storageClassList.Items[i]
+		if (storageClassList.Items[i].Provisioner == csi.GetCephFSDriverName()) || (storageClassList.Items[i].Provisioner == csi.GetRBDDriverName()) {
```
leelavg, line 362: I don't think in this statement you require extra pair of parens.
```suggestion
		if storageClassList.Items[i].Provisioner == csi.GetCephFSDriverName() || storageClassList.Items[i].Provisioner == csi.GetRBDDriverName() {
```

controllers/operatorconfigmap_controller.go
```diff
@@ -279,36 +289,29 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// ownerReference on it as its cluster scoped resource
 	cephfsCSIDriver := templates.CephFSCSIDriver.DeepCopy()
 	cephfsCSIDriver.ObjectMeta.Name = csi.GetCephFSDriverName()
-	err = csi.CreateCSIDriver(c.ctx, c.Client, cephfsCSIDriver)
-	if err != nil {
+	if err := csi.CreateCSIDriver(c.ctx, c.Client, cephfsCSIDriver); err != nil {
```
leelavg, line 292: I don't think there's need for cosmetic changes as we are returning on every non-nil value, changing from `err = func() bool; if err != nil {}` to `if err := func() bool; err !=nil {}` is kinda adding extra diff.

leelavg (changes requested):
last time I deployed this PR, I hit an RBAC issue around setting operator cm as owner for csi-mon cm, did you test the deployment?

controllers/operatorconfigmap_controller.go
```diff
@@ -194,80 +229,55 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		c.log.Error(err, "failed to create monitor configmap", "name", encConfigMap.Name)
 		return ctrl.Result{}, err
 	}
 
-	c.cephFSDeployment = &appsv1.Deployment{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      csi.CephFSDeploymentName,
-			Namespace: c.OperatorNamespace,
-		},
-	}
```
leelavg, line 211: as discussed, removals of this initialization is the reason for panic that I mentioned earlier, pls reinstate and test.

leelavg:
does other PRs #79, #80, #81 deprecates this PR?

bernerhat:
> does other PRs #79, #80, #81 deprecates this PR?

No, this code still needs to be tested and merged.

controllers/operatorconfigmap_controller.go
```diff
@@ -116,30 +116,38 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		return ctrl.Result{}, err
 	}
 
-	instance := configv1.ClusterVersion{}
-	if err = c.Client.Get(context.TODO(), req.NamespacedName, &instance); err != nil {
-		return ctrl.Result{}, err
+	clusterVersion := &configv1.ClusterVersion{}
+	if err := c.Client.Get(c.ctx, types.NamespacedName{Name: clusterVersionName}, clusterVersion); err != nil {
+		c.log.Error(err, "failed to get the clusterVersion version of the OCP cluster")
+		return reconcile.Result{}, err
```
nb-ohad, line 147: Please move all `get` operations to the top before running nay reconciliation code

controllers/operatorconfigmap_controller.go
nb-ohad: @leelavg moving any set of variables into a deeper scope (if it is only needed there) is a welcomed change.

controllers/operatorconfigmap_controller.go
```diff
@@ -329,11 +328,11 @@ func (c *ClusterVersionReconciler) createOrUpdate(obj client.Object, f controlle
 	return nil
 }
 
-func (c *ClusterVersionReconciler) own(obj client.Object) error {
-	return controllerutil.SetControllerReference(c.OperatorDeployment, obj, c.Client.Scheme())
+func (c *OperatorConfigMapReconciler) own(obj client.Object, operatorConfigMap *corev1.ConfigMap) error {
```
nb-ohad, line 331: The entire point of having an own util on the reconciler is to get an ownership ref by the input cr and not to take an owner as an argument. If you need to add a specialized owner you can use `SetControllerReference` instead. Please revert the change

nb-ohad (changes requested):

controllers/operatorconfigmap_controller.go
nb-ohad: Yes @leelavg,  you are right.
The context passed to the reconcile function (from the controller) already contain the request object

bernerhat:
code is now tested.

controllers/operatorconfigmap_controller.go
```diff
@@ -90,56 +90,66 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	c.log.Info("Reconciling ClusterVersion")
+	c.log = log.FromContext(ctx, "OperatorConfigMap", req)
+	c.log.Info("Reconciling OperatorConfigMap")
 
-	if err := c.ensureConsolePlugin(); err != nil {
-		c.log.Error(err, "unable to deploy client console")
-		return ctrl.Result{}, err
+	c.operatorConfigMap = &corev1.ConfigMap{}
+	c.operatorConfigMap.Name = operatorConfigMapName
+	c.operatorConfigMap.Namespace = c.OperatorNamespace
```
nb-ohad, line 134: You are getting a reconcile event on the config map. please take the information from there 
```suggestion
	c.operatorConfigMap.Name = req.Name
	c.operatorConfigMap.Namespace = req.Namespace
```

nb-ohad (changes requested):

PR #65, opened 2024-01-29
controllers: pick next best version of csi images

leelavg:
pick the csi sidecar images version that are best suitable for the platform

### DIFF
--- a/pkg/csi/csi.go
+++ b/pkg/csi/csi.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/util/version"
 )
@@ -43,9 +44,9 @@ type SidecarImages struct {
 	ContainerImages containerImages `yaml:"containerImages"`
 }
 
-var sidecarImages = new(SidecarImages)
+var sidecarImages *SidecarImages
 
-func InitializeSidecars(ver string) error {
+func InitializeSidecars(log logr.Logger, ver string) error {
 	// ready yaml files and yaml unmarshal to SidecarImages
 	// and set to csiSidecarImages
 	si := []SidecarImages{}
@@ -58,18 +59,33 @@ func InitializeSidecars(ver string) error {
 		return err
 	}
 
-	sv := version.MustParseGeneric(ver)
+	pltVersion := version.MustParseGeneric(ver)
 
-	for _, image := range si {
-		v := version.MustParseGeneric(image.Version)
-		if sv.Major() == v.Major() && sv.Minor() == v.Minor() {
-			sidecarImages = &image
-			break
+	closestMinor := int64(-1)
+	for idx := range si {
+		siVersion := version.MustParseGeneric(si[idx].Version)
+		log.Info("searching for the most compatible CSI image version", "CSI", siVersion, "Platform", pltVersion)
+
+		// only check sidecar image versions that are not higher than platform
+		if siVersion.Major() == pltVersion.Major() && siVersion.Minor() <= pltVersion.Minor() {
+			// filter sidecar closest to platform version
+			if int64(siVersion.Minor()) > closestMinor {
+				sidecarImages = &si[idx]
+				closestMinor = int64(siVersion.Minor())
+			}
+			if closestMinor == int64(pltVersion.Minor()) { // exact match and early exit
+				break
+			}
+		} else {
+			log.Info("skipping sidecar images: version greater than platform version")
 		}
 	}
-	if sidecarImages.Version == "" {
-		return fmt.Errorf("failed to find container details for %v version in %v", sv.String(), sidecarImages)
+	if sidecarImages == nil {
+		// happens only if all sidecars image versions are greater than platform
+		return fmt.Errorf("failed to find container details suitable for %v platform version", pltVersion)
 	}
+
+	log.Info("selected sidecar images", "version", sidecarImages.Version)
 
 	return nil
 }
--- a/controllers/clusterversion_controller.go
+++ b/controllers/clusterversion_controller.go
@@ -139,7 +139,7 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
-	if err := csi.InitializeSidecars(instance.Status.Desired.Version); err != nil {
+	if err := csi.InitializeSidecars(c.log, instance.Status.Desired.Version); err != nil {
 		c.log.Error(err, "unable to initialize sidecars")
 		return ctrl.Result{}, err
 	}


leelavg:
btw, below testing was done:
1. Create OCP 4.15 and deploy client-operator 4.14, picks 4.14 sidecar images
2. Update client-operator to 4.15, picks 4.15 sidecar images

pkg/csi/csi.go
```diff
@@ -60,13 +60,25 @@ func InitializeSidecars(ver string) error {
...
+			if sv.Minor() == v.Minor() {
+				sidecarImages = &image
+				break
+			}
+			// for a brief period of time platform gets ahead of operator and we
+			// need to use image versions that are lagging one minor version of platform
+			// as next best option
+			if sv.Minor()-1 == v.Minor() {
+				oneLessMinorImages = &image
+			}
```
nb-ohad, line 76: `else` is always preferable to `break` for code structure and there is no need to use a new variable and then just copy it back into the original one again later.

```suggestion
		if sv.Major() == v.Major() {
			if sv.Minor() == v.Minor() {
				sidecarImages = &image

			// for a brief period of time platform gets ahead of operator and we
			// need to use image versions that are lagging one minor version of platform
			// as next best option	
			} else if sv.Minor()-1 == v.Minor() {
				sidecarImages = new(SidecarImages)
			}
```

pkg/csi/csi.go
```diff
@@ -60,13 +60,25 @@ func InitializeSidecars(ver string) error {
...
 		}
 	}
+	if sidecarImages.Version == "" {
+		sidecarImages = oneLessMinorImages
+	}
```
nb-ohad, line 81:
```suggestion

```

pkg/csi/csi.go
```diff
@@ -60,13 +60,25 @@ func InitializeSidecars(ver string) error {
 
 	sv := version.MustParseGeneric(ver)
 
+	oneLessMinorImages := new(SidecarImages)
```
nb-ohad, line 63:
```suggestion
```

nb-ohad (changes requested):
@leelavg I have added some code structure comments but I think this PR has no merit as the behaviour should already be covered by the existing mechanized. The current mechanized is supposed to take the most matching CSI version based on the mapping and if an N-1 operator is installed on an N OCP then there will be no match for N and the N-1 version should be taken even without this code

leelavg:
@nb-ohad @Madhu-1 
> should already be covered by the existing mechanized.
- existing mechanism isn't covering it

As per offline discussion, the requirement seems to be selecting next best version.

However, shouldn't we have a ceiling on how farther we can go? Can that be normalized for all platform versions or any platform creates a jump point, IOW is there any criteria that OCP at X.Y can support sidecars that are lagging upto X.<?> ?

pkg/csi/csi.go
leelavg: removed

pkg/csi/csi.go
leelavg: removed

pkg/csi/csi.go
leelavg: removed

leelavg:
> The current mechanized is supposed to take the most matching CSI version based on the mapping 
done.

pkg/csi/csi.go
```diff
@@ -58,19 +59,28 @@ func InitializeSidecars(ver string) error {
...
 	}
 
-	sv := version.MustParseGeneric(ver)
-
-	for _, image := range si {
-		v := version.MustParseGeneric(image.Version)
-		if sv.Major() == v.Major() && sv.Minor() == v.Minor() {
-			sidecarImages = &image
-			break
+	pltVersion := version.MustParseGeneric(ver)
```
Madhu-1, line 62: @leelavg how about the below code, this is just a suggestion. 

```go
bestDifference := 0
	// Iterate over sidecar images and find the best match
	for _, sidecar := range si {
		sidecarVersion := version.MustParseGeneric(sidecar.Version)
		// if exact match is found break the loop
		if sidecarVersion.Major() == pltVersion.Major() && sidecarVersion.Minor() == pltVersion.Minor() {
			sidecarImages = &sidecar
			break
		} else if sidecarVersion.Major() == pltVersion.Major() && sidecarVersion.Minor() < pltVersion.Minor() {
			// Check if sidecar version is compatible with the platform version
			// Calculate the difference in minor versions
			difference := pltVersion.Minor() - sidecarVersion.Minor()

			// Update the best match if it's closer
			if difference <= bestDifference {
				sidecarImages = &sidecar
				bestDifference = difference
			}
		} else {
			log.Info("skipping sidecar images version greater than platform version", "sidecar", sidecarVersion, "platform", pltVersion)
		}
	}
	if sidecarImages.Version == "" {
		// happens only if all sidecars image versions are greater than platform
		return fmt.Errorf("failed to find container details for %v version in %v", pltVersion.String(), sidecarImages)
	}
```

leelavg:
retested

pkg/csi/csi.go
leelavg: may I also know what's missing from current code that requires change?

Madhu-1: As i mentioned this is just a suggestion, this breaks the loop if the match is found and add more comments

leelavg: I was just curious, this "breaks the loop if the match is found" address my question, I'll add more comments. thanks.

nb-ohad: @Madhu-1 @leelavg If we are already suggesting alternative implementations what about this?

```
	score := 999
	var images *SidecarImages = nil 
	
	for idx := range si {
		siVersion := version.MustParseGeneric(si[idx].Version)
		if siVersion.Major() != pltVersion.Major() {
			log.Info("skipping sidecar images version greater than platform version", "sidecar", siVersion, "platform", pltVersion)
			continue;
		}

		diff := pltVersion.Minor - siVersion.Mior;
		if diff < score {
			score = diff;
			images = &si[idx]
		}

		if score == 0 {
			break;
		}
	}
```

It has the benefit of skipping the additional array, it does stop if it finds an exact, removes the multiple comparisons on major versions, shorter, and easier to read

leelavg: > It has the benefit of skipping the additional array
- I'm only using additional array to only log in-case we didn't find any suitable version
- Existing code has a bug w/ this, if we didn't find a match it was logging empty var
----
> you can test if you found the images using a comparison to nil
- I coded this way not to use break/continue, intermediate vars and always select the best without preexisting knowledge of exact match being the best
----
the only issue I see w/ my implementation is logging of skipping sidecars (like we can have sidecars at 4.Y and platform at 3.Y which is close to impossible). I can skip major version check altogether, however configmap is user editable so didn't remove that.
----
I'm a bit confused what exactly needs to be changed, I can add comments and call it a day if that's enough. Or pls let me know, what changes are required (not in code but as a problem statement). thanks.

leelavg:
retested

pkg/csi/csi.go
leelavg: addressed below issues which I believe were the reasons for suggestions:
* [x] comments
* [x] extra array for logging
* [x] early exit on exact match
* [x] parsing version twice
I didn't implement difference matching as I need a conversion b/n uint & int, that same job is being done by `closestVersion` logically

leelavg:
retested
/unhold

pkg/csi/csi.go
```diff
@@ -58,19 +59,37 @@ func InitializeSidecars(ver string) error {
...
+
+	// this helps in not parsing the version twice in loop.
+	closestVersion := pltVersion.WithMinor(0)
+	for idx := range si {
+		siVersion := version.MustParseGeneric(si[idx].Version)
+		log.Info("verifying sidecar images version for best compatibility to platform",
+			"sidecar", siVersion, "platform", pltVersion)
```
nb-ohad, line 69: 1. This is not the way we do multiline statement in this codebase
2. You are not verifying here, you are checking
3. Cosntant parts should be part of the message itself

```suggestion
		log.Info(
			"Searching for the most compatible CSI image version: CSI: %s, Platform: %s",
			siVersion, 
			pltVersion,
		)
```

pkg/csi/csi.go
```diff
@@ -58,19 +59,37 @@ func InitializeSidecars(ver string) error {
...
+			if closestVersion.Minor() <= siVersion.Minor() {
+				sidecarImages = &si[idx]
+				closestVersion = siVersion
+			}
+			if closestVersion.Minor() == pltVersion.Minor() { // exact match and early exit
+				break
+			}
+		} else {
+			log.Info("skipping sidecar images version greater than platform version",
+				"sidecar", siVersion, "platform", pltVersion)
```
nb-ohad, line 83: You already logged the versions on the last message in the log. No need to log again.
```suggestion
			log.Info("skipping sidecar images: version greater than platform") 
```

pkg/csi/csi.go
```diff
@@ -58,19 +59,37 @@ func InitializeSidecars(ver string) error {
...
-
-	for _, image := range si {
-		v := version.MustParseGeneric(image.Version)
-		if sv.Major() == v.Major() && sv.Minor() == v.Minor() {
-			sidecarImages = &image
-			break
+	pltVersion := version.MustParseGeneric(ver)
+
+	// this helps in not parsing the version twice in loop.
+	closestVersion := pltVersion.WithMinor(0)
```
nb-ohad, line 65: Please initialize to `-1`. `0` creates an edge case where the z stream version is `X.0`

pkg/csi/csi.go
```diff
@@ -58,19 +59,37 @@ func InitializeSidecars(ver string) error {
...
+	for idx := range si {
+		siVersion := version.MustParseGeneric(si[idx].Version)
+		log.Info("verifying sidecar images version for best compatibility to platform",
+			"sidecar", siVersion, "platform", pltVersion)
+
+		// only check sidecar image versions that are not higher than platform
+		if siVersion.Major() == pltVersion.Major() && siVersion.Minor() <= pltVersion.Minor() {
+			// filter sidecar closest to platform version
+			if closestVersion.Minor() <= siVersion.Minor() {
```
nb-ohad, line 74: This condition is searching for the version that has the biggest minor version. From a readability perspective, it makes more sense to have written the other way around. Also, no need for the equal sign, we do not expect 2 entires to have the same version in the yaml. 

```suggestion
			if siVersion.Minor() > closestVersion.Minor  {
```

nb-ohad (changes requested):

pkg/csi/csi.go
leelavg: ack for the rest
> 3. Cosntant parts should be part of the message itself
- logr package support key & value pairs to be part of log, I'll need to use fmt.Sprintf to take the change

pkg/csi/csi.go
leelavg: 1. we can't create minor version w/ -1
2. I know of X.0 but didn't change that due to my incorrect assumption
how about
`if closestVersion.Minor() == pltVersion.Minor()` to `if closestVersion.Minor() == pltVersion.Minor() && sidecarImage != nil`

pkg/csi/csi.go
leelavg: > we do not expect 2 entires to have the same version in the yaml.
- this was for to match the initial closest version of `0`

pkg/csi/csi.go
leelavg: If we don't want to do an early exit this extra condition can go away

pkg/csi/csi.go
nb-ohad: @Madhu-1 @leelavg With the recent changes can we resolve this thread?

pkg/csi/csi.go
```diff
@@ -58,19 +59,35 @@ func InitializeSidecars(ver string) error {
...
+			// filter sidecar closest to platform version
+			if siVersion.Minor() >= closestVersion.Minor() {
+				sidecarImages = &si[idx]
+				closestVersion = siVersion
+			}
+			if closestVersion.Minor() == pltVersion.Minor() && sidecarImages != nil { // exact match and early exit
+				break
+			}
+		} else {
+			log.Info("skipping sidecar images: version greater than platform version")
```
nb-ohad, line 81: The code only uses the closest version for the minor part and never considers the rest of its parts. 
As such, it will be easier to treat it as an int and use `-1` to generalize the edge case.

```suggestion
	closestMinor := -1
	for idx := range si {
		siVersion := version.MustParseGeneric(si[idx].Version)
		log.Info("searching for the most compatible CSI image version", "CSI", siVersion, "Platform", pltVersion)

		// only check sidecar image versions that are not higher than platform
		if siVersion.Major() == pltVersion.Major() && siVersion.Minor() <= pltVersion.Minor() {
			// filter sidecar closest to platform version
			if siVersion.Minor() > closestMinor {
				sidecarImages = &si[idx]
				closestMinor = siVersion.Minor()
			}
			if closestMinor == pltVersion.Minor {
				break
			}
		} else {
			log.Info("skipping sidecar images: version greater than platform version")
```

nb-ohad (changes requested):

pkg/csi/csi.go
leelavg: done.

PR #72, opened 2024-02-07
makefile: update controller gen version

rewantsoni:
To add the XValidation we need to update the controller-gen version.
Hence updating it to the latest 0.9.x version

### DIFF
--- a/hack/make-bundle-vars.mk
+++ b/hack/make-bundle-vars.mk
@@ -70,7 +70,7 @@ CSI_ADDONS_BUNDLE_IMG ?= $(CSI_ADDONS_IMAGE_REGISTRY)/$(CSI_ADDONS_REGISTRY_NAME
 CATALOG_IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(CATALOG_IMAGE_NAME):$(IMAGE_TAG)
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+CRD_OPTIONS ?= "crd:generateEmbeddedObjectMeta=true"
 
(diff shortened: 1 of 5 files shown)


api/v1alpha1/storageclassclaim_types.go
```diff
@@ -73,6 +73,8 @@ type StorageClassClaim struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="oldSelf == self",message="spec is immutable"
```
Madhu-1, line 77: please push the generated CRD files as this validation need to be added to CRD yaml

Madhu-1 (approved):

hack/make-bundle-vars.mk
```diff
@@ -70,7 +70,7 @@ CSI_ADDONS_BUNDLE_IMG ?= $(CSI_ADDONS_IMAGE_REGISTRY)/$(CSI_ADDONS_REGISTRY_NAME
 CATALOG_IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(CATALOG_IMAGE_NAME):$(IMAGE_TAG)
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+CRD_OPTIONS ?= "crd:generateEmbeddedObjectMeta=true"
```
leelavg, line 73: could you pls mention what these change in options mean?

rewantsoni: the controllern-gen version 0.9.2 doesn't have the crd options: `crd:trivialVersions=true,preserveUnknownFields=false`.
From the [doc](https://book.kubebuilder.io/reference/controller-gen#generators):
```
schemapatch
generateEmbeddedObject - patches existing CRDs with new schemata.
It will generate output for each "CRD Version" (API version of the CRD type itself) , e.g. apiextensions/v1) available.
```

leelavg (approved):

PR #79, opened 2024-02-20
Change ownership logic to enforce after the DeepCopy

bernerhat:
Ownership enforcement is occurring after the DeepCopy method and causing the created ceph resources to be spawned without an owner.

### DIFF
--- a/controllers/clusterversion_controller.go
+++ b/controllers/clusterversion_controller.go
@@ -194,7 +194,7 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			"config.json": "[]",
 		},
 	}
-	if err := c.own(monConfigMap); err != nil {
+	if err := c.own(encConfigMap); err != nil {
 		return ctrl.Result{}, err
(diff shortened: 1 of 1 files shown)


controllers/clusterversion_controller.go
leelavg, line 1: we should not fix this bug if we are going to change the controller owner of these resources from deployment to configmap.

basically `SetControllerReference` sets `Controller` field to `true` and among many owners only one owner should set this field to `true`. So, if we are setting Deployment as controller for rest of the resources now, then the operation will fail when you want these rest of the resources to have configmap as controlling owner.

leelavg:
@bernerhat what's the update on this PR?

bernerhat:
This issue remains even after changing the controller into a configmap since it isn't related directly. the only change here is the order of which the ownership is set. since the get ceph resource functions return a new object in memory the ownership gets overwritten and that is the only thing i'm trying to fix here.

regarding the refactoring of the controller, the owner reference will change from the original deployment resource to the configmap in pr #64

PR #80, opened 2024-02-20
Handle csi drivers and scc deletion

bernerhat:
Create a deletion phase and delete the csi drivers and scc upon operator's removal as it is currently not being handled in the ClusterVersion controller.

### DIFF
--- a/controllers/operatorconfigmap_controller.go
+++ b/controllers/operatorconfigmap_controller.go
@@ -65,6 +65,8 @@ const (
 	deployCSIKey           = "DEPLOY_CSI"
 	subscriptionLabelKey   = "managed-by"
 	subscriptionLabelValue = "webhook.subscription.ocs.openshift.io"
+
+	operatorConfigMapFinalizer = "ocs-client-operator.ocs.openshift.io/storageused"
 )
 
 // OperatorConfigMapReconciler reconciles a ClusterVersion object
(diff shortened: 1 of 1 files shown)


controllers/clusterversion_controller.go
```diff
@@ -320,6 +348,22 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	return ctrl.Result{}, nil
 }
 
+func (c *ClusterVersionReconciler) deletionPhase() (reconcile.Result, error) {
+	if err := csi.DeleteCSIDriver(c.ctx, c.Client, csi.GetCephFSDriverName()); err != nil && !k8serrors.IsNotFound(err) {
+		c.log.Error(err, "unable to delete cephfs CSIDriver")
+		return ctrl.Result{}, err
+	}
+	if err := csi.DeleteCSIDriver(c.ctx, c.Client, csi.GetRBDDriverName()); err != nil && !k8serrors.IsNotFound(err) {
```
leelavg, line 356: where are these `DeleteCSIDriver` methods defined?

controllers/clusterversion_controller.go
```diff
@@ -139,6 +141,32 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	//ensure finalizer
+	if !controllerutil.ContainsFinalizer(&instance, ClusterVersionFinalizer) {
+		c.log.Info("finalizer missing on the ClusterVersion resource, adding...")
+		controllerutil.AddFinalizer(&instance, ClusterVersionFinalizer)
+		if err := c.Client.Update(c.ctx, &instance); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
```
leelavg, line 169: we'll mostly not be around when we a deletion event is triggered on `ClusterVersion` basically, it's a cluster teardown.

controllers/clusterversion_controller.go
bernerhat: The `DeleteCSIDriver` is a function that was already in the repo under `pkg/csi/`

controllers/clusterversion_controller.go
bernerhat: discussed offline to split the changes originally suggested under #64 to several PRs in order to better incapsulate a single change in each of them, that is correct that we will not be around when the cluster is tearing down but this change will also not break our controller (still needs to be tested) after PR #64 will take into affect this code should as well. if you think we should base it on one another that is an option as well but i thought this approach will allow us to get them both in since they are not directly related.

controllers/clusterversion_controller.go
```diff
@@ -139,6 +141,32 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		if !result.Requeue {
+			//remove finalizer
+			controllerutil.RemoveFinalizer(&instance, ClusterVersionFinalizer)
+			if err := c.Client.Update(c.ctx, &instance); err != nil {
+				return ctrl.Result{}, err
+			}
+			c.log.Info("finallizer removed successfully")
+		}
+		return result, nil
```
nb-ohad, line 158: There are 3 conditions where we should not remove the finalizer:
1. an error was returned
2. The result indicates the need for a requeue via the `Requeue` field
3. The result indicated the need for a requeue via the `RequeueAfter` field


```suggestion
		if err != nil || !result.Requeue || result.RequeueAfter > 0 {
			return result, err
		}
			
		//remove finalizer
		controllerutil.RemoveFinalizer(&instance, ClusterVersionFinalizer)
		if err := c.Client.Update(c.ctx, &instance); err != nil {
			return ctrl.Result{}, err
		}
		c.log.Info("finallizer removed successfully")
		
		return ctrl.Result{}, nil
		
```

controllers/clusterversion_controller.go
```diff
@@ -139,6 +141,32 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}
+
+	//ensure finalizer
+	if !controllerutil.ContainsFinalizer(&instance, ClusterVersionFinalizer) {
```
nb-ohad, line 162: the way the code is written right now there is no clear code separation between deletion and the other reconciliation phases. can you please move everything after deletion phase handling into an else section?

or even better, reorder to the deletion phase handling code will be on the else path

```
 if instance.GetDeletionTimestamp().IsZero() {
     (normal reconcile)
 } else {
     (deletion reconcile)
 }

File: controllers/clusterversion_controller.go [hunk 5]
```diff
@@ -320,6 +348,22 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		return ctrl.Result{}, err
+	}
+	if err := csi.DeleteCSIDriver(c.ctx, c.Client, csi.GetRBDDriverName()); err != nil && !k8serrors.IsNotFound(err) {
+		c.log.Error(err, "unable to delete rbd CSIDriver")
+		return ctrl.Result{}, err
+	}
+	if err := c.Client.Delete(c.ctx, c.scc); err != nil && !k8serrors.IsNotFound(err) {
+		c.log.Error(err, "unable to delete SCC")
+		return ctrl.Result{}, err
+	}
```
nb-ohad, line 363: I am not sure if Security Context Constraints resources are cluster-scopped resources or namespace-scoped resources. If they are namespace scope we can have an owner reference on them which will remove the need for this code

nb-ohad (changes requested):

controllers/clusterversion_controller.go
bernerhat: from what i've checked it is a cluster scoped resource so that is not a possibility unfortunately.

controllers/clusterversion_controller.go
bernerhat: i've tried to apply this approach in a previous iteration and got flagged by lint iirc, i'll try once more and update in case not possible.

controllers/clusterversion_controller.go
nb-ohad: Ok , resolving

controllers/clusterversion_controller.go
```diff
@@ -55,6 +55,8 @@ const (
 	// ClusterVersionName is the name of the ClusterVersion object in the
 	// openshift cluster.
 	clusterVersionName = "version"
+
+	ClusterVersionFinalizer = "ocs-client-operator.ocs.openshift.io/ocspvcused"
```
Madhu-1, line 59: i think `ocspvcused` is the right one to use here, we can have no PVC but we can have volumesnapshots and also we are not checking for any resources before deletion

controllers/clusterversion_controller.go
```diff
@@ -139,184 +141,225 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	})
-	if err != nil {
-		c.log.Error(err, "failed to create/update prometheus rules")
+	if err := c.Client.Delete(c.ctx, c.scc); err != nil && !k8serrors.IsNotFound(err) {
+		c.log.Error(err, "unable to delete SCC")
 		return ctrl.Result{}, err
 	}
-
-	c.log.Info("prometheus rules deployed", "prometheusRule", klog.KRef(prometheusRule.Namespace, prometheusRule.Name))
-
```
Madhu-1, line 319: https://github.com/red-hat-storage/ocs-client-operator/blob/a5c640e9dcf43bf9239bb75f436f553f316ae715/controllers/clusterversion_controller.go#L466-L472

The above doesnt have any ownerRef we might need to add to that as well?

nb-ohad: I think handling cleanup of console components should go into its own PR

controllers/clusterversion_controller.go
nb-ohad: @Madhu-1 I believe the resource checking is coming in a different PR (https://github.com/red-hat-storage/ocs-client-operator/pull/81)

But I agree about the naming for the finalizer:
@bernerhat can we change i to ocs-client-operator.ocs.openshift.io/storageinuse" (Storage in Use)
@Madhu-1 if you have a suggestion for a different name please reply here

Madhu-1: `ocs-client-operator.ocs.openshift.io/storageinuse` make sense 👍🏻

controllers/clusterversion_controller.go
Madhu-1: Yes we can , we just need to make sure we don't miss it out.

bernerhat: I dont have an issue with adding an ownerRef, but since this is not related to this PR i'll open another one addressing it.

controllers/operatorconfigmap_controller.go
```diff
@@ -170,209 +172,249 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-	} else if deployCSI {
-		clusterVersion := &configv1.ClusterVersion{}
-		clusterVersion.Name = clusterVersionName
-		if err := c.get(clusterVersion); err != nil {
-			c.log.Error(err, "failed to get the clusterVersion version of the OCP cluster")
-			return reconcile.Result{}, err
+		//ensure finalizer
+		if !controllerutil.ContainsFinalizer(c.operatorConfigMap, OperatorConfigMapFinalizer) {
+			c.log.Info("finalizer missing on the operatorConfigMap resource, adding...")
+			controllerutil.AddFinalizer(c.operatorConfigMap, OperatorConfigMapFinalizer)
```
leelavg, line 180: if your intention is to just add finalizer why can't the response (bool) of AddFinalizer be used directly?

controllers/operatorconfigmap_controller.go
```diff
@@ -170,209 +172,249 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			if err := c.Client.Update(c.ctx, c.operatorConfigMap); err != nil {
+				return ctrl.Result{}, err
+			}
+			c.log.Info("finallizer removed successfully")
+		}
 	}
+	return ctrl.Result{}, nil
+}
 
+func (c *OperatorConfigMapReconciler) deletionPhase() (reconcile.Result, error) {
```
leelavg, line 405: since the first return statement is always same can that be removed?

controllers/operatorconfigmap_controller.go
bernerhat: There is no real reason other than readability of the code. When i was writing the commit i was drawing inspiration from both the code base already existing here and the osd-deployer (there it's not exactly the same though).
This can be changed.

controllers/operatorconfigmap_controller.go
bernerhat: not sure i'm following, can you please elaborate?

controllers/operatorconfigmap_controller.go
```diff
@@ -65,6 +65,8 @@ const (
 	deployCSIKey           = "DEPLOY_CSI"
 	subscriptionLabelKey   = "managed-by"
 	subscriptionLabelValue = "webhook.subscription.ocs.openshift.io"
+
+	OperatorConfigMapFinalizer = "ocs-client-operator.ocs.openshift.io/storageused"
```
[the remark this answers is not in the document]
bernerhat: not, is a typo. thanks.

controllers/operatorconfigmap_controller.go
leelavg: You are always returning `ctrl.Result{}` in this func body and so that can be omitted right? Like,
```suggestion
func (c *OperatorConfigMapReconciler) deletionPhase() error {
```

Issue #85, opened 2024-02-22
storageclassclaim should identify a connected client by client name and namespace


leelavg:
affected places as of now

https://github.com/red-hat-storage/ocs-client-operator/blob/1b9b7fb23b8d9b6837296ddcf31d925e26787aeb/controllers/storageclassclaim_controller.go#L198
https://github.com/red-hat-storage/ocs-client-operator/blob/1b9b7fb23b8d9b6837296ddcf31d925e26787aeb/controllers/storageclient_controller.go#L351

PR #87, opened 2024-02-22
controllers: use annotations for claims and client mapping

leelavg:
a storageclassclaim can be attributed to a storageclient based on storageclient name and namespace, this commit makes sure to set annotations correctly.

fixes: #85

### DIFF
--- a/controllers/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim_controller.go
@@ -195,8 +195,6 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 	// Close client-side connections.
 	defer providerClient.Close()
 
-	addAnnotation(r.storageClassClaim, storageClientLabel, r.storageClient.Name)
-
 	cc := csi.ClusterConfig{
 		Client:    r.Client,
 		Namespace: r.OperatorNamespace,
@@ -241,13 +239,20 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 		// Configuration phase.
 		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimConfiguring
 
+		updateStorageClassClaim := false
 		// Check if finalizers are present, if not, add them.
 		if !contains(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer) {
-			storageClassClaimRef := klog.KRef(r.storageClassClaim.Name, r.storageClassClaim.Namespace)
-			r.log.Info("Finalizer not found for StorageClassClaim. Adding finalizer.", "StorageClassClaim", storageClassClaimRef)
+			r.log.Info("Finalizer not found for StorageClassClaim. Adding finalizer.", "StorageClassClaim", r.storageClassClaim.Name)
 			r.storageClassClaim.SetFinalizers(append(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer))
+			updateStorageClassClaim = true
+		}
+		if addAnnotation(r.storageClassClaim, storageClientAnnotationKey, client.ObjectKeyFromObject(r.storageClient).String()) {
+			updateStorageClassClaim = true
+		}
+
+		if updateStorageClassClaim {
 			if err := r.update(r.storageClassClaim); err != nil {
-				return reconcile.Result{}, fmt.Errorf("failed to update StorageClassClaim [%v] with finalizer: %s", storageClassClaimRef, err)
+				return reconcile.Result{}, fmt.Errorf("failed to update StorageClassClaim %q: %v", r.storageClassClaim.Name, err)
 			}
 		}
 
@@ -572,14 +577,18 @@ func (r *StorageClassClaimReconciler) own(resource metav1.Object) error {
 	return controllerutil.SetOwnerReference(r.storageClassClaim, resource, r.Scheme)
 }
 
-// addAnnotation add a annotation to a resource metadata
-func addAnnotation(obj metav1.Object, key string, value string) {
+// addAnnotation adds an annotation to a resource metadata, returns true if added else false
+func addAnnotation(obj metav1.Object, key string, value string) bool {
 	annotations := obj.GetAnnotations()
 	if annotations == nil {
 		annotations = map[string]string{}
 		obj.SetAnnotations(annotations)
 	}
-	annotations[key] = value
+	if oldValue, exist := annotations[key]; !exist || oldValue != value {
+		annotations[key] = value
+		return true
+	}
+	return false
 }
 
 func (r *StorageClassClaimReconciler) createOrReplaceVolumeSnapshotClass(volumeSnapshotClass *snapapi.VolumeSnapshotClass) error {
--- a/controllers/storageclient_controller.go
+++ b/controllers/storageclient_controller.go
@@ -57,10 +57,14 @@ const (
 	GetStorageConfig      = "GetStorageConfig"
 	AcknowledgeOnboarding = "AcknowledgeOnboarding"
 
-	storageClientLabel          = "ocs.openshift.io/storageclient"
+	storageClientAnnotationKey  = "ocs.openshift.io/storageclient"
 	storageClientNameLabel      = "ocs.openshift.io/storageclient.name"
 	storageClientNamespaceLabel = "ocs.openshift.io/storageclient.namespace"
 	storageClientFinalizer      = "storageclient.ocs.openshift.io"
+
+	// indexes for caching
+	storageProviderEndpointIndexName = "index:storageProviderEndpoint"
+	storageClientAnnotationIndexName = "index:storageClientAnnotation"
 
 	csvPrefix = "ocs-client-operator"
 )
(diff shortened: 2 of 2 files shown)


leelavg:
/hold testing in progress

@Madhu-1, @nb-ohad I hope I'm wrong, w/o this commit I believe we can remove storageclient even if corresponding storageclassclaim exists. Initially, I was try to fix something else and found this out. Happy to be corrected.

Madhu-1:
> /hold testing in progress
> 
> @Madhu-1, @nb-ohad I hope I'm wrong, w/o this commit I believe we can remove storageclient even if corresponding storageclassclaim exists. Initially, I was try to fix something else and found this out. Happy to be corrected.

yes it looks like we have bug where we are adding annotation in storageclassclaim but checking labels in storageclient 🤦🏻

leelavg:
ya, it's a bug, just deployed latest 4.15. Will update after verifying the fix.
```
# kc get storageclassclaim,storageclient
NAME                                                             STORAGETYPE        STORAGEPROFILE   STORAGECLIENTNAME   STORAGECLIENTNAMESPACE     PHASE
storageclassclaim.ocs.openshift.io/ocs-storagecluster-ceph-rbd   blockpool                           storageclient       openshift-storage-client   Ready
storageclassclaim.ocs.openshift.io/ocs-storagecluster-cephfs     sharedfilesystem                    storageclient       openshift-storage-client   Ready

NAME                                           PHASE       CONSUMER
storageclient.ocs.openshift.io/storageclient   Connected   a5d76cfb-19e6-484d-869a-7793ce47a5d9

# kc delete storageclient storageclient
storageclient.ocs.openshift.io "storageclient" deleted

# kc get storageclassclaim,storageclient
NAME                                                             STORAGETYPE        STORAGEPROFILE   STORAGECLIENTNAME   STORAGECLIENTNAMESPACE     PHASE
storageclassclaim.ocs.openshift.io/ocs-storagecluster-ceph-rbd   blockpool                           storageclient       openshift-storage-client   Ready
storageclassclaim.ocs.openshift.io/ocs-storagecluster-cephfs     sharedfilesystem                    storageclient       openshift-storage-client   Ready
```

leelavg:
/unhold

tested, working as expected.

controllers/storageclassclaim_controller.go
```diff
@@ -572,14 +578,18 @@ func (r *StorageClassClaimReconciler) own(resource metav1.Object) error {
 	return controllerutil.SetOwnerReference(r.storageClassClaim, resource, r.Scheme)
 }
 
-// addAnnotation add a annotation to a resource metadata
-func addAnnotation(obj metav1.Object, key string, value string) {
+// addAnnotation add an annotation to a resource metadata, returns true if added
```
nb-ohad, line 581:
```suggestion
// addAnnotation adds an annotation to a resource metadata, and returns true if added
```

controllers/storageclassclaim_controller.go
```diff
@@ -572,14 +578,18 @@ func (r *StorageClassClaimReconciler) own(resource metav1.Object) error {
...
 	if annotations == nil {
 		annotations = map[string]string{}
 		obj.SetAnnotations(annotations)
 	}
-	annotations[key] = value
+	if annotations[key] != value {
+		annotations[key] = value
+		return true
+	}
+	return false
```
nb-ohad, line 592: This will give the wrong answer if there was no annotation and you are trying to set an annotation with a value of an empty string

```suggestion
	oldValue, ok := annotations[key]
	if return !ok || oldValue != value {
		annotations[key] = value
		return true
	}
	return false
```

controllers/storageclient_controller.go
```diff
@@ -348,8 +348,8 @@ func (s *StorageClientReconciler) verifyNoStorageClassClaimsExist(instance *v1al
 
 	for i := range storageClassClaims.Items {
 		storageClassClaim := &storageClassClaims.Items[i]
-		sc := storageClassClaim.Labels[storageClientLabel]
-		if sc == instance.Name {
+		sc := storageClassClaim.GetAnnotations()
+		if sc[storageClientAnnotationKey] == client.ObjectKeyFromObject(instance).String() {
```
nb-ohad, line 352: Can't we instead add an indexer for the annotation and list by that indexer?

controllers/storageclassclaim_controller.go
```diff
@@ -241,13 +239,21 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 		// Configuration phase.
 		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimConfiguring
 
+		updateRequired := false
```
nb-ohad, line 242: `updateRequired` is a very generic name introduced at a high-level scope. Let's have it more specific, maybe: `storageClassClaimUpdateNeeded` or something similar

nb-ohad (changes requested):

controllers/storageclassclaim_controller.go
leelavg: done.

controllers/storageclassclaim_controller.go
leelavg: done.

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclassclaim_controller.go
leelavg: > at a high-level scope
- not so sure, the definition and usage is quite close and so picked a short var
- nevertheless, took in the suggestion as it doesn't hurt anything

controllers/storageclassclaim_controller.go
```diff
@@ -241,13 +239,21 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
...
 
+		updateStorageClassClaim := false
 		// Check if finalizers are present, if not, add them.
 		if !contains(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer) {
 			storageClassClaimRef := klog.KRef(r.storageClassClaim.Name, r.storageClassClaim.Namespace)
 			r.log.Info("Finalizer not found for StorageClassClaim. Adding finalizer.", "StorageClassClaim", storageClassClaimRef)
 			r.storageClassClaim.SetFinalizers(append(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer))
+			updateStorageClassClaim = true
+		}
+		if addAnnotation(r.storageClassClaim, storageClientAnnotationKey, client.ObjectKeyFromObject(r.storageClient).String()) {
```
rchikatw, line 250: Is `client.ObjectKeyFromObject(r.storageClient).String()` the same as `r.storageClient.Name` or are they different? If yes any specific reason you are using client.ObjectKeyFromObject

leelavg: `client.ObjectKeyFromObject` returns `types.NamespacedName` and a `String` method on that return `<namespace>/<name>` of the resulting resource by which every claim can be linked back to the corresponding client.

controllers/storageclassclaim_controller.go
```diff
@@ -572,14 +578,18 @@ func (r *StorageClassClaimReconciler) own(resource metav1.Object) error {
 	return controllerutil.SetOwnerReference(r.storageClassClaim, resource, r.Scheme)
 }
 
-// addAnnotation add a annotation to a resource metadata
-func addAnnotation(obj metav1.Object, key string, value string) {
+// addAnnotation adds an annotation to a resource metadata, returns true if added
```
Madhu-1, line 581:
```suggestion
// addAnnotation adds an annotation to a resource metadata, returns true if added else false
```

controllers/storageclassclaim_controller.go
```diff
@@ -241,13 +239,21 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
...
+			updateStorageClassClaim = true
+		}
+		if addAnnotation(r.storageClassClaim, storageClientAnnotationKey, client.ObjectKeyFromObject(r.storageClient).String()) {
+			updateStorageClassClaim = true
+		}
+
+		if updateStorageClassClaim {
 			if err := r.update(r.storageClassClaim); err != nil {
-				return reconcile.Result{}, fmt.Errorf("failed to update StorageClassClaim [%v] with finalizer: %s", storageClassClaimRef, err)
+				return reconcile.Result{}, fmt.Errorf("failed to update StorageClassClaim [%v]: %v", client.ObjectKeyFromObject(r.storageClassClaim), err)
```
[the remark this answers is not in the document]
leelavg: done.

controllers/storageclassclaim_controller.go
leelavg: well, return value can't be empty and bool has only two valid values, nevertheless took in the suggestion as being explicit doesn't hut.

Madhu-1 (commented):
LGTM.

controllers/storageclient_controller.go
```diff
@@ -90,6 +90,13 @@ func (s *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}
 		return res
 	})
+
+	if err := mgr.GetCache().IndexField(context.TODO(), &v1alpha1.StorageClassClaim{}, storageClientAnnotationKey, func(obj client.Object) []string {
```
nb-ohad, line 94: I don't think we want a new context for each operation within `SetupWithManager` as well as we don't want to use  `context.TODO()` where `context.Background()` makes more sense. Let's initialize one context to be used within `SetupWithManager` and reuse it

nb-ohad (changes requested):

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclient_controller.go
```diff
@@ -78,18 +78,26 @@ type StorageClientReconciler struct {
...
-	_ = mgr.GetCache().IndexField(context.TODO(), &v1alpha1.StorageClient{}, "spec.storageProviderEndpoint", func(o client.Object) []string {
+	_ = mgr.GetCache().IndexField(ctx, &v1alpha1.StorageClient{}, "spec.storageProviderEndpoint", func(o client.Object) []string {
 		res := []string{}
 		if o.(*v1alpha1.StorageClient).Status.Phase != v1alpha1.StorageClientFailed {
 			res = append(res, o.(*v1alpha1.StorageClient).Spec.StorageProviderEndpoint)
 		}
 		return res
 	})
+
+	if err := mgr.GetCache().IndexField(ctx, &v1alpha1.StorageClassClaim{}, storageClientAnnotationKey, func(obj client.Object) []string {
```
nb-ohad, line 95: @Madhu-1 @leelavg I think we need to make an effort to have a convention for indexers' names. 
Right now we have `spec.xxx` for one field and then `ocs.openshift.io/xxx` for another. 

I think the best way this whole and to give context on the list operation is to use an index prefix (as we are listing on an index not a filed) and then a meaningful index name.


So `spec.storageProviderEndpoint` becomes `index:storageProviderEndpoint` and for the annotation, we can have `index:storageClientAnnotation`

I also think we should use literals and not constants, or a constant like `storageClientAnnotationIndexName`  in the list operation because we want to have it very visible that we are listing on an indexer in the list call site

controllers/storageclient_controller.go
```diff
@@ -78,18 +78,26 @@ type StorageClientReconciler struct {
...
+		return []string{obj.GetAnnotations()[storageClientAnnotationKey]}
+	}); err != nil {
+		return fmt.Errorf("unable to setup FieldIndexer for storageclient annotation: %v", err)
```
nb-ohad, line 98:
```suggestion
		return fmt.Errorf("unable to set up FieldIndexer <index name>: %v", err)
```

nb-ohad (changes requested):

controllers/storageclient_controller.go
leelavg: > I also think we should use literals and not constants, or a constant like storageClientAnnotationIndexName in the list operation because we want to have it very visible that we are listing on an indexer in the list call site
- I suppose you mean, if we use constant we want to make it obvious that we are using the field for index or we don't want const at all?

controllers/storageclient_controller.go
leelavg: done.

PR #88, opened 2024-02-23
storageclassclaim: check for and configure RADOS namespace

jarrpa:
This PR should be "code complete", but without the provider side merged there's no way to fully test it yet.

Depends on red-hat-storage/ocs-operator#2478

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/controllers/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim_controller.go
@@ -340,14 +340,25 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 					return reconcile.Result{}, fmt.Errorf("failed to create or update secret %v: %s", secret, err)
 				}
 			case "StorageClass":
+				if rns, ok := data["radosnamespace"]; ok {
+					csiClusterConfigEntry.CephRBD = new(csi.CephRBDSpec)
+					csiClusterConfigEntry.CephRBD.RadosNamespace = rns
+					delete(data, "radosnamespace")
+				}
+
+				// The clusterID is an opaque value used by the CSI driver
+				// to identify the cluster config (e.g. mon IPs) to use
+				// for a volume from a given StorageClass. We set it to
+				// the claim name for ease of identification.
+				//
+				// NOTE: This is distinct from the notion of a "clusterID"
+				// used within Ceph and Rook-Ceph, despite sharing the
+				// same name.
 				csiClusterConfigEntry.ClusterID = r.storageClassClaim.Name
 				var storageClass *storagev1.StorageClass
 				data["csi.storage.k8s.io/provisioner-secret-namespace"] = r.storageClient.Namespace
 				data["csi.storage.k8s.io/node-stage-secret-namespace"] = r.storageClient.Namespace
 				data["csi.storage.k8s.io/controller-expand-secret-namespace"] = r.storageClient.Namespace
-				// generate a new clusterID for cephfs subvolumegroup, as
-				// storageclassclaim is clusterscoped resources using its
-				// name as the clusterID
 				data["clusterID"] = r.storageClassClaim.Name
 
 				if resource.Name == "cephfs" {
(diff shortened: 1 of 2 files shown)


controllers/storageclassclaim_controller.go
```diff
@@ -358,6 +358,9 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 				if err != nil {
 					return reconcile.Result{}, fmt.Errorf("failed to create or update StorageClass: %s", err)
 				}
+			case "CephBlockPoolRadosNamespace":
+				csiClusterConfigEntry.CephRBD = new(csi.CephRBDSpec)
+				csiClusterConfigEntry.CephRBD.RadosNamespace = resource.Name
```
nb-ohad, line 363: @jarrpa Is this old code? If I remember correctly, we agreed that the resource array indicates actual resources to create or edit. We are not creating or editing `CephBlockPoolRadosNamespace` on the client so we should do all to processing to aggregate into client side resources on the provider side

nb-ohad (changes requested):

nb-ohad:
LGTM

leelavg:
can this be merged as is?

controllers/storageclassclaim_controller.go
```diff
@@ -339,15 +339,19 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
...
+					data["clusterID"] = r.storageClassClaim.Name
+				}
 				var storageClass *storagev1.StorageClass
 				data["csi.storage.k8s.io/provisioner-secret-namespace"] = r.storageClient.Namespace
 				data["csi.storage.k8s.io/node-stage-secret-namespace"] = r.storageClient.Namespace
 				data["csi.storage.k8s.io/controller-expand-secret-namespace"] = r.storageClient.Namespace
-				// generate a new clusterID for cephfs subvolumegroup, as
-				// storageclassclaim is clusterscoped resources using its
-				// name as the clusterID
-				data["clusterID"] = r.storageClassClaim.Name
```
leelavg, line 350: why is this being removed, we actually depend on `clusterID` for cleanup?

leelavg:
/hold

for my comment.

controllers/storageclassclaim_controller.go
```diff
@@ -339,15 +339,19 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 					return reconcile.Result{}, fmt.Errorf("failed to create or update secret %v: %s", secret, err)
 				}
 			case "StorageClass":
-				csiClusterConfigEntry.ClusterID = r.storageClassClaim.Name
+				if rns, ok := data["radosnamespace"]; ok {
+					csiClusterConfigEntry.ClusterID = data["clusterID"]
```
Madhu-1, line 343: why clusterID is set from the provider to the consumer cluster? we use claimName as the clusterID for both cephfs and rbd ,something changed in any discussions

jarrpa: I was going off of your blog post for consuming RADOS namespaces in CSI: https://www.mrajanna.com/Consuming-storage-from-exteranal-ceph-in-rook/

Madhu-1: He above was for Rook and cephcsi , as we are not using Rook we can keep the claim name as clusterID and add radosnamespace name to the configmap , which seems to be the perfect fix here. ClusterIds are just identification here

controllers/storageclassclaim_controller.go
leelavg: i see the line is moved to above. closing.

controllers/storageclassclaim_controller.go
jarrpa: Any reason to not keep this change, however? If it's just an arbitrary identifier for the CSI driver, does it matter?

Madhu-1: IMO I don't think it should be sent/determined by provider what should be the local identifier on consumer cluster, I would still prefer to use claim names as identifier, it also helps in mapping clusterid belongs to which claim without looking at many resources

leelavg: I'm not entirely sure about technicalities, I'm indexing on clusterID to be same as storageclaim for validating the presence of persistentvolume and volumesnapshotcontent during cleanup, can that not be broken (unless we have a requirement to change it).

Deletion flow in [here](https://github.com/red-hat-storage/ocs-client-operator/pull/116#issuecomment-2007142821) all depends on clusterid from storageclass is transferred to pv & vsc

jarrpa: @leelavg Based on what I can tell, since the `clusterID` is an opaque value, it doesn't matter what it's actually set to since it'll still propagate  as you expect it to from SC to PV.

@Madhu-1 Good enough! It's nice to see someone else taking human troubleshooting into consideration. :joy:

jarrpa: And done.

leelavg:
LGTM.

Madhu-1 (approved):

PR #89, opened 2024-02-29
store desired subscription channel as annotation on storageclient CR

leelavg:
provider sends it's subscription channel name as response to client heartbeat, we will store that in the corresponding storageclient as an annotation value.

this will help in making upgrade decisions going forward.

part of [RHSTOR-5496]

### DIFF
--- a/service/status-report/main.go
+++ b/service/status-report/main.go
@@ -38,7 +38,8 @@ import (
 )
 
 const (
-	csvPrefix = "ocs-client-operator"
+	desiredSubscriptionChannelAnnotationKey = "ocs.openshift.io/subscription.channel"
+	csvPrefix                               = "ocs-client-operator"
 )
 
 func main() {
@@ -138,8 +139,18 @@ func main() {
 	status := providerclient.NewStorageClientStatus().
 		SetPlatformVersion(pltVersion).
 		SetOperatorVersion(oprVersion)
-	if _, err = providerClient.ReportStatus(ctx, storageClient.Status.ConsumerID, status); err != nil {
+	statusResponse, err := providerClient.ReportStatus(ctx, storageClient.Status.ConsumerID, status)
+	if err != nil {
 		klog.Exitf("Failed to report status of storageClient %v: %v", storageClient.Status.ConsumerID, err)
+	}
+
+	storageClientCopy := &v1alpha1.StorageClient{}
+	storageClient.DeepCopyInto(storageClientCopy)
+	if utils.AddAnnotation(storageClient, desiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
+		// patch is being used here as to not have any conflicts over storageclient cr changes as this annotation value doesn't depend on storageclient spec
+		if err := cl.Patch(ctx, storageClient, client.MergeFrom(storageClientCopy)); err != nil {
+			klog.Exitf("Failed to annotate storageclient %q: %v", storageClient.Name, err)
+		}
 	}
 
 	var csiClusterConfigEntry = new(csi.ClusterConfigEntry)
--- a/controllers/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim_controller.go
@@ -25,6 +25,7 @@ import (
 
 	v1alpha1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/csi"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 
 	"github.com/go-logr/logr"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
@@ -246,7 +247,7 @@ func (r *StorageClassClaimReconciler) reconcilePhases() (reconcile.Result, error
 			r.storageClassClaim.SetFinalizers(append(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer))
 			updateStorageClassClaim = true
 		}
-		if addAnnotation(r.storageClassClaim, storageClientAnnotationKey, client.ObjectKeyFromObject(r.storageClient).String()) {
+		if utils.AddAnnotation(r.storageClassClaim, storageClientAnnotationKey, client.ObjectKeyFromObject(r.storageClient).String()) {
 			updateStorageClassClaim = true
 		}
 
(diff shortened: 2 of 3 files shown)


api/v1alpha1/storageclient_types.go
```diff
@@ -52,6 +52,9 @@ type StorageClientStatus struct {
 
 	// ConsumerID will hold the identity of this cluster inside the attached provider cluster
 	ConsumerID string `json:"id,omitempty"`
+
+	// Name of the subscription channel that this client wants the operator to be at.
+	DesiredOperatorSubscriptionChannel string `json:"subscriptionChannel,omitempty"`
```
nb-ohad, line 57: We should stay away from field names that do not match the JSON/YAML prop name. 
Let's settle down on DesiredSubscriptionChannel. In addition, is this something that could be reconstructed if the status is wiped? if so, from where? if notI think we need a different approach.

leelavg: > if notI think we need a different approach.
- ok, may I know why do we need to solve that?

leelavg: moved to storing the channel in annotation.

service/status-report/main.go
```diff
@@ -138,10 +141,46 @@ func main() {
...
-	if _, err = providerClient.ReportStatus(ctx, storageClient.Status.ConsumerID, status); err != nil {
+	statusResponse, err := providerClient.ReportStatus(ctx, storageClient.Status.ConsumerID, status)
+	if err != nil {
 		klog.Exitf("Failed to report status of storageClient %v: %v", storageClient.Status.ConsumerID, err)
 	}
 
+	versionFromResponse, err := getVersionFromChannel(statusResponse.DesiredClientOperatorChannel)
+	if err != nil {
+		klog.Exitf("Failed to map channel name from status response to version: %v", err)
+	}
```
nb-ohad, line 152: Why are we doing this? why aren't we just saving the channel in the annotation?

service/status-report/main.go
```diff
@@ -138,10 +141,46 @@ func main() {
...
+		if err != nil {
+			klog.Exitf("Failed to map channel name from storageclient annotation to version: %v", err)
+		}
+		// bump the annotation if it is less than version value from response, having higher version in annotation implies user overrode the upgrade
+		// negation is used here for checking strictly greater version
+		if !versionFromResponse.LessThan(versionFromAnnotation) {
+			// since we are referring only version from channel, by checkig for strictly greater version we'll enforce the channel name is also same
+			updateAnnotation = true
+		}
+	}
```
nb-ohad, line 172: I am not sure why we are doing all of this. Channel names are assumed to have lexical ordering. You should be able to just compare the strings.

service/status-report/main.go
```diff
@@ -166,3 +205,17 @@ func main() {
...
+		return nil, fmt.Errorf("unable to refer version from channel name")
+	}
+
+	genericVersion, err := version.ParseGeneric(versionFromChannel)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse referred version from channel: %w", err)
+	}
+
+	return genericVersion, nil
+}
```
nb-ohad, line 221: I don't see a need to extract the version from the channel name

nb-ohad (changes requested):

service/status-report/main.go
leelavg: I don't think so
```
# rlwrap -A yaegi
> // Compare (a, b string) int ==> 0 for a==b, -1 for a<b, +1 for a>b
> strings.Compare("stable-4.10", "stable-4.9")
: -1
> 
```

leelavg:
part of #89

leelavg:
rebased.

service/status-report/main.go
```diff
@@ -138,10 +141,46 @@ func main() {
...
 		klog.Exitf("Failed to report status of storageClient %v: %v", storageClient.Status.ConsumerID, err)
 	}
 
+	versionFromResponse, err := getVersionFromChannel(statusResponse.DesiredClientOperatorChannel)
+	if err != nil {
+		klog.Exitf("Failed to map channel name from status response to version: %v", err)
+	}
+
+	var updateAnnotation bool
+	if storageClient.GetAnnotations()[desiredSubscriptionChannelAnnotationKey] == "" {
```
nb-ohad, line 155: storageClient metadata might not contain an annotation map

```suggestion
	anno := storageClient.GetAnnotations();
	if  anno == nil || anno[desiredSubscriptionChannelAnnotationKey] == "" {
```

service/status-report/main.go
```diff
@@ -138,10 +141,46 @@ func main() {
...
+		updateAnnotation = true
+	}
+
+	var versionFromAnnotation *version.Version
+	if !updateAnnotation {
```
nb-ohad, line 160:
```suggestion
	} else {
```

service/status-report/main.go
```diff
@@ -138,10 +141,46 @@ func main() {
...
+		var err error
+		versionFromAnnotation, err = getVersionFromChannel(storageClient.GetAnnotations()[desiredSubscriptionChannelAnnotationKey])
```
nb-ohad, line 162: You are already fetching the storage client's annotations above, use that map instead of fetching again.

service/status-report/main.go
```diff
@@ -138,10 +141,46 @@ func main() {
...
+		if err != nil {
+			klog.Exitf("Failed to map channel name from storageclient annotation to version: %v", err)
+		}
+		// bump the annotation if it is less than version value from response, having higher version in annotation implies user overrode the upgrade
+		// negation is used here for checking strictly greater version
+		if !versionFromResponse.LessThan(versionFromAnnotation) {
+			// since we are referring only version from channel, by checkig for strictly greater version we'll enforce the channel name is also same
+			updateAnnotation = true
+		}
```
nb-ohad, line 171: This entire check does not belong here. You should always store the desired channel. The code that uses the annotation should take into consideration the value itself.

nb-ohad (changes requested):

service/status-report/main.go
leelavg: The intention was for user to provide a way to bump the version by overriding the annotation, if we are updating the annotation w/ a lower version (as wanted by provider) then users intention will not stick, isn't it?

nb-ohad: Ok, but why is the version in the channel an indication that it was user-edited? Also, how are you expecting the user to understand the channel text format?

leelavg: > but why is the version in the channel an indication that it was user-edited
- we weren't going to update the annotation value if it is more than provider version and that signifies it's user edited

> how are you expecting the user to understand the channel text format
- the channel name is usually version prefixed w/ `stable-`, since there isn't any documentation that it'll stay the same dropping this as suggested offline.

service/status-report/main.go
leelavg: Removed code which extracts version from channel.

service/status-report/main.go
leelavg: Removed code which extracts version from channel.

service/status-report/main.go
leelavg: Removed code which extracts version from channel.

service/status-report/main.go
leelavg: Removed code which extracts version from channel.

PR #94, opened 2024-03-01
feat: register validating webhook and add labels to client operator subscription

leelavg:
after controller start up we register webhook and also label the subscription which ensures only updates on that specific subscription are routed to the webhook.

in conjunction with #89 we will ensure updates to the subscription channel is validated and rejected/allowed before persisting to etcd and corresponding controllers on the subscription.

part of [RHSTOR-5496]

Depends on #98


leelavg:
/hold

testing in progress and review only last two commits.

leelavg:
part of #115

PR #92, opened 2024-03-01
ci: update golang and golint versions

leelavg:
- Update to go 1.21 based on ocs-operator
- Update golint to latest available version 1.56.2

PR #97, opened 2024-03-05
ci: use action.yaml for setting up go during CI

leelavg:
create an action.yaml file for setting up golang and reuse in github workflows wherever golang installation is required.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,0 +1,14 @@
+---
+name: Setup Go Environment
+description: Fetch Go version from the go.mod file and use it to setup Go environment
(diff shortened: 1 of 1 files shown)


leelavg:
As per https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#testing-out-your-action-in-a-workflow, it seems `action.yaml` should be committed before using that in the workflows.

As part of #92 I tried using action.yaml with same commit which didn't work so the idea is to merge this PR and rebase #92 on top of this.

PR #102, opened 2024-03-06
rename storageclassclaim to storageclaim in storageclaim_controller

leelavg:
some renames are intentionally left out as they correspond to gRPC & Protobuf and the expectation is user should be able to create storageclaim and get all the functionality that storageclassclaim is used to provide.

part of [RHSTOR-5553]

Depends on #101

PR #103, opened 2024-03-06
migrate existing storageclassclaims to storageclaims

leelavg:
when storageclaims reconciler starts we migrate all the existing storageclassclaims to storageclaims and also delete storageclassclaims after the migration, the intention is to support existing user and later they should work w/ storageclaims directly.

part of [RHSTOR-5553]

Depends on #102


leelavg:
/hold testing in progress
only review last 2 commits.

leelavg:
rebased

leelavg:
now part of #116

PR #104, opened 2024-03-06
place a deprecation warning on StorageClassClaim API 

leelavg:
mark StorageClassClaim CRD as deprecated and add a suggestion to use StorageClaim CRD. This doesn't prohibit users creating the CR but it'll give a indication and encourages user to use new CRD.

part of [RHSTOR-5553]

Depends on #103


leelavg:
only review last comment, nominal testing done as
```
# k create -f config/crd/bases/ocs.openshift.io_storageclassclaims.yaml 
customresourcedefinition.apiextensions.k8s.io/storageclassclaims.ocs.openshift.io created

# k create -f config/samples/ocs_v1alpha1_storageclassclaim.yaml 
Warning: StorageClassClaim API is deprecated and will be removed in future version, please use StorageClaim API instead.
The StorageClassClaim "storageclassclaim-sample" is invalid: 
* spec: Required value
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
*
```

leelavg:
rebased

leelavg:
now part of #116

PR #105, opened 2024-03-06
csi: enable csi-omap based on key in operator config

rewantsoni:
The csi-omap generator needs to the enabled on the application cluster for RDR. 

Add a new key to operator configMap controller to enable omap. If the key is not set we will always enable this option on operator Config CR. If key is available we will set the value to what was provider in `ocs-client-operator-config`

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -67,6 +67,7 @@ const (
 	manageNoobaaSubKey     = "manageNoobaaSubscription"
 	subscriptionLabelKey   = "managed-by"
 	subscriptionLabelValue = "webhook.subscription.ocs.openshift.io"
+	generateRbdOMapInfoKey = "generateRbdOMapInfo"
 
 	operatorConfigMapFinalizer = "ocs-client-operator.ocs.openshift.io/storageused"
 	subPackageIndexName        = "index:subscriptionPackage"
@@ -341,6 +342,7 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		if err := c.own(rbdDriver); err != nil {
 			return fmt.Errorf("failed to own csi rbd driver: %v", err)
 		}
+		rbdDriver.Spec.GenerateOMapInfo = ptr.To(c.shouldGenerateRBDOmapInfo())
 		return nil
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile rbd driver: %v", err)
@@ -505,6 +507,11 @@ func (c *OperatorConfigMapReconciler) getNoobaaSubManagementConfig() bool {
 		return false
 	}
 	return val
+}
+
+func (c *OperatorConfigMapReconciler) shouldGenerateRBDOmapInfo() bool {
+	valAsString := strings.ToLower(c.operatorConfigMap.Data[generateRbdOMapInfoKey])
+	return valAsString == strconv.FormatBool(true)
 }
 
 func (c *OperatorConfigMapReconciler) get(obj client.Object, opts ...client.GetOption) error {
--- a/bundle/manifests/ocs-client-operator-config_v1_configmap.yaml
+++ b/bundle/manifests/ocs-client-operator-config_v1_configmap.yaml
@@ -1,4 +1,6 @@
 apiVersion: v1
+data:
+  generateRbdOMapInfo: "true"
 kind: ConfigMap
 metadata:
   name: ocs-client-operator-config
--- a/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
@@ -7,7 +7,7 @@ metadata:
     categories: Storage
     console.openshift.io/plugins: '["odf-client-console"]'
     containerImage: quay.io/ocs-dev/ocs-client-operator:latest
-    createdAt: "2024-11-13T08:34:27Z"
+    createdAt: "2024-11-14T15:13:37Z"
     description: OpenShift Data Foundation client operator enables consumption of
       storage services from a remote centralized OpenShift Data Foundation provider
       cluster.
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -9,7 +9,9 @@ configMapGenerator:
 - files:
   - controller_manager_config.yaml
   name: manager-config
-- name: config
+- literals:
+  - generateRbdOMapInfo=true
+  name: config
 
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization


controllers/clusterversion_controller.go
```diff
@@ -86,7 +89,7 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			func(client client.Object) bool {
 				namespace := client.GetNamespace()
 				name := client.GetName()
-				return ((namespace == c.OperatorNamespace) && (name == operatorConfigMapName))
+				return ((namespace == c.OperatorNamespace) && (name == operatorConfigMapName) || (name == csiSidecarConfigMapName))
```
leelavg, line 92: I think the associativity changes here due to precedence of `&&` being higher than `||` and existing grouping does the same w/ or w/o it, so better be
```suggestion
				return namespace == c.OperatorNamespace && (name == operatorConfigMapName || name == csiSidecarConfigMapName)
```
btw, I hope you are looking for cm of operator/csisidecar from operatorns?

controllers/clusterversion_controller.go
```diff
@@ -366,6 +383,15 @@ func (c *ClusterVersionReconciler) getOperatorConfig() (*corev1.ConfigMap, error
 	return cm, nil
 }
 
+func (c *ClusterVersionReconciler) getSideCarConfig() (*corev1.ConfigMap, error) {
+	cm := &corev1.ConfigMap{}
+	err := c.Client.Get(c.ctx, types.NamespacedName{Name: csiSidecarConfigMapName, Namespace: c.OperatorNamespace}, cm)
```
leelavg, line 388: pls refer the method `get` from other controllers which uses `ctx` and constructs `types.NamespacedName` by default.

controllers/clusterversion_controller.go
```diff
@@ -203,6 +206,20 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	sideCarConfig, err := c.getSideCarConfig()
```
leelavg, line 209: I don't think you need a func for this and also doing the error handling twice w/o any benefit.

controllers/clusterversion_controller.go
```diff
@@ -203,6 +206,20 @@ func (c *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if err != nil && !k8serrors.IsNotFound(err) {
+		c.log.Error(err, "failed to get csi side car configmap", "name", sideCarConfig)
+		return ctrl.Result{}, err
+	}
+	enableOMAPGenerator := false
+	if value, ok := sideCarConfig.Data[csiOMAPGeneratorKey]; ok {
```
leelavg, line 215: are you fine w/ not having this data in the cm?

leelavg (commented):
I think deployment generation will change after 107, will re-review after that.

leelavg:
pls rebase, it's a bit confusing to review in current state, thanks.

controllers/operatorconfigmap_controller.go
```diff
@@ -291,6 +294,25 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			sideCarConfig.Namespace = c.OperatorNamespace
+
+			err = c.get(sideCarConfig)
+			if err != nil && !kerrors.IsNotFound(err) {
+				c.log.Error(err, "failed to get csi side car configmap", "name", sideCarConfig)
+				return ctrl.Result{}, err
+			}
+
+			enableOMAPGenerator := false
+			if value, ok := sideCarConfig.Data[csiOMAPGeneratorKey]; ok {
```
leelavg, line 308: technically this'll work, however we'll run this code if the CM is not found and so why not check this only if CM is found?

controllers/storageclaim_controller.go
```diff
@@ -321,6 +321,24 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
+				csiSideCarConfig.Name = eResource.Name
+				csiSideCarConfig.Namespace = r.OperatorNamespace
+
+				data := map[string]string{}
+				err = json.Unmarshal(eResource.Data, &data)
+				if err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to unmarshal csi-sidecar-config configuration response: %v", err)
+				}
+
+				_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, csiSideCarConfig, func() error {
```
leelavg, line 335: how do you intend to clean this up? Is it missing owner ref? Should CM be the owner as CM (~CSI) lifetime is more than StorageClaim. What happens if one clients says `yes` and another `no`? Is that possible?

pkg/templates/csisidecars.go
```diff
@@ -162,6 +162,36 @@ var CSIAddonsContainer = corev1.Container{
...
+	VolumeMounts: []corev1.VolumeMount{
+		{
+			Name:      "ceph-csi-configs",
+			MountPath: DefaultCephCSIConfigPath,
+		},
+		{
+			Name:      "keys-tmp-dir",
+			MountPath: DefaultTmpDir,
+		},
+	},
```
leelavg, line 192: are these volumes already being set in deployment definition?

leelavg (commented):
after going through the code, just thinking, what's the down side of always deploying this sidecar?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -347,6 +351,26 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
...
+	sideCarConfig.Namespace = c.OperatorNamespace
+
+	err := c.get(sideCarConfig)
+	if err != nil && !kerrors.IsNotFound(err) {
+		c.log.Error(err, "failed to get csi side car configmap", "name", sideCarConfig)
+		return err
+	}
+
+	var deployOMAP bool
+	if kerrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
rewantsoni: right, missed that updating it

internal/controller/storageclient_controller.go
```diff
@@ -283,6 +283,28 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			configMap.Name = eResource.Name
+			configMap.Namespace = r.OperatorNamespace
+
+			data := map[string]string{}
+			err = json.Unmarshal(eResource.Data, &data)
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshal %s configuration response: %v", eResource.Name, err)
+			}
+
+			_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, configMap, func() error {
```
Madhu-1, line 297: in think we have internal `createOrUpdate` can we use it?

rewantsoni: done

Madhu-1:
LGTM

internal/controller/operatorconfigmap_controller.go
```diff
@@ -507,6 +509,22 @@ func (c *OperatorConfigMapReconciler) getNoobaaSubManagementConfig() bool {
...
+	}
+	val, err := strconv.ParseBool(valAsString)
+	if err != nil {
+		c.log.Error(
+			err,
+			"Unsupported value under enableOmapGenerator key",
+		)
+		return false
+	}
+	return val
```
nb-ohad, line 525:
```suggestion
	valAsString := strings.ToLower(c.operatorConfigMap.Data[enableOmapGeneratorKey])
	return valAsString != "" && valAsString != strconv.FormatBool(false) 

```

nb-ohad (changes requested):

PR #106, opened 2024-03-07
verify dependents before deletion of storageclaim

leelavg:
verify the existence of persistent volume or/and volume snapshot content before proceeding with removal of finalizer during storageclaim deletion.

part of [RHSTOR-5553]

Depends on #103


leelavg:
@Madhu-1 need your input on last comment. thanks.

controllers/storageclaim_controller.go
```diff
@@ -70,6 +72,19 @@ type StorageClaimReconciler struct {
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *StorageClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	ctx := context.Background()
+	if err := mgr.GetCache().IndexField(ctx, &corev1.PersistentVolume{}, pvClusterIDIndexName, func(o client.Object) []string {
+		pv := o.(*corev1.PersistentVolume)
+		if pv.Spec.CSI == nil {
+			return nil
+		}
```
Madhu-1, line 80: add a check for cephfs/rbd driver names as well

controllers/storageclaim_controller.go
```diff
@@ -690,3 +693,92 @@ func (r *StorageClaimReconciler) migrateStorageClassClaims() error {
...
+		driver := vsContent.Spec.Driver
+		if driver != csi.GetRBDDriverName() || driver != csi.GetCephFSDriverName() {
+			// skip the resources that doesn't belong to ceph csi
+			continue
+		}
+
+		vsClass := &snapapi.VolumeSnapshotClass{}
+		// shortest path for verifying snapshot dependents via volumesnapshotclass
+		if vsContent.Spec.VolumeSnapshotClassName != nil {
+			vsClass.Name = *vsContent.Spec.VolumeSnapshotClassName
```
Madhu-1, line 729: This may not be correct, extract the clusterID from the snapshotHandle that will cover all the cases

leelavg: `vsc.spec.source.volumeHandle` or `vsc.status.snapshotHandle`?

`volumeHandle: 0001-001b-ocs-storagecluster-ceph-rbd-0000000000000004-8cc378f9-7620-4eef-aa47-45372edecba9` vs `snapshotHandle: 0001-001b-ocs-storagecluster-ceph-rbd-0000000000000004-2360c9a6-46fa-4d31-b0e4-79104c1e2e65`

and what's the criteria to extract the clusterID?

btw, thanks a lot, good thing I didn't went down the rabbit hole and implement all those branches 😅

Madhu-1: `vsc.status.snapshotHandle` is the right one,

https://github.com/ceph/ceph-csi/blob/devel/docs/design/proposals/clusterid-mapping.md#design-to-handle-clusterid-and-poolid-for-dr will give you idea on how to extract clusterID

leelavg: ack, i checked above, is there any func that actually does the transformation rather than duplicating the code?

controllers/storageclaim_controller.go
leelavg: done.

controllers/storageclaim_controller.go

leelavg:
review only last 2 commits.

leelavg:
now part of #116

PR #107, opened 2024-03-07
Refactor CSI package code to reduce memory, GC overhead, and "order of operations" considerations

nb-ohad:
The existing code is very wasteful regarding memory management of large in-memory structures. On each reconcile cycle multiple big nested structures are either fully or partially copied multiple times and in multiple locations in the codebase. This creates unnecessary pressure on the garbage collector and wastes multiple CPU cycles moving around the same bits

A second side effect of the current code structure is the creation of a fresh nested set of objects on every reconcile cycle that overwrites any metadata on existing objects. This results in a contributor’s need to be very careful and consider the order of operations when writing reconciliation functions for CSI resources. As an example, I can give the fact that the current code contains a bug where all of the owner references on CSI reconciled objects are overwritten by a deep copy operation

This PR addresses the 2 points above and reduces the number of allocations and garbage collection of in-memory reconciled objects and the need to micro-manage the order of operation on CSI objects` reconciliation.

P.S
The bug mentioned above is now solved on the main branch by the contributor changing the order op operations, but that makes this point even more valid

### DIFF
--- a/pkg/csi/cephfsdaemonset.go
+++ b/pkg/csi/cephfsdaemonset.go
@@ -23,13 +23,14 @@ import (
 	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 )
 
 const (
-	CephFSDamonSetName = "csi-cephfsplugin"
+	CephFSDaemonSetName = "csi-cephfsplugin"
+
+	cephFSDaemonSetContainerName = "csi-cephfsplugin"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


jarrpa (changes requested):
This diff is maddening... can you split this into multiple commits? I see at least two possible commits out of this, if not more. At a bare minimum the `pkg/templates` changes paired with the change in functions from `Get` to `Set` would greatly improve the readability of this PR.

nb-ohad:
@jarrpa There was an issue as my base branch was way behind. Fixed now!

pkg/csi/cephfsdeployment.go
```diff
@@ -35,183 +34,172 @@ var cephfsDeploymentLabels = map[string]string{
...
+		Spec: corev1.PodSpec{
+			ServiceAccountName: cephFSProvisionerServiceAccountName,
+			Containers: []corev1.Container{
+				{Name: templates.ProvisionerContainer.Name},
+				{Name: templates.AttacherContainer.Name},
+				{Name: templates.ResizerContainer.Name},
+				{Name: templates.SnapshotterContainer.Name},
+				{
+					Name:            "csi-cephfsplugin",
+					Image:           sidecarImages.ContainerImages.CephCSIImageURL,
```
jarrpa, line 52: Looks like you need to initialize `sidecarImages` in `csi.go` global scope.

jarrpa (changes requested):

pkg/csi/csi.go
```diff
@@ -99,3 +101,9 @@ func GetCephFSDriverName() string {
 func GetRBDDriverName() string {
 	return "openshift-storage.rbd.csi.ceph.com"
 }
+
+func findContainer(spec *corev1.PodTemplateSpec, name string) *corev1.Container {
```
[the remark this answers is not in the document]
nb-ohad: @jarrpa Consider the following:
If I am taking your suggestion and accepting `PodSpec` then I could also just take a container slice instead.
The result will be that for all of the call sites (which we have many of) I will need to pass something like `&deploy.Spec.Template.Spec.Template`. At that point, there is no real value in having the util itself as I could replace the invocation with `utils.Find(spec.Spec.Containers, selectByName(name))`. 

The point of having a non-public util func is the consolidation of as much as possible of the common code shared between all call sites. As we are dealing with both daemon sets and deployments I technically couldn't pass the entire object and had to drill down to the common denominator which is the PodTemplateSpec. An alternative would be to have two different utils, one accepting a daemon set and one a deployment. But I deemed that an overkill

pkg/csi/cephfsdeployment.go
nb-ohad: I think you are correct in your observation, but my solution will be to move setting of the image into the function that process the template copy

nb-ohad: Done

pkg/csi/rbddeployment.go
```diff
@@ -32,223 +31,179 @@ var rbdDeploymentLabels = map[string]string{
 }
 
 const (
+	RBDDamonSetName   = "csi-rbdplugin"
```
jarrpa, line 34: Looks like a straggler? If so, please remove.

jarrpa (changes requested):

pkg/csi/rbddaemonset.go
```diff
@@ -52,349 +51,295 @@ var (
...
+				templates.DefaultKubeletDirPath,
+				GetRBDDriverName(),
+			))
+
+		case templates.CSIAddonsContainer.Name:
+			c.Image = sidecarImages.ContainerImages.CSIADDONSImageURL
+			vol := utils.Find(c.VolumeMounts, func(vol *corev1.VolumeMount) bool {
+				return vol.Name == "socket-dir"
+			})
+			vol.Name = "plugin-dir"
```
leelavg, line 339: I think this panics due to above mentioned reason, we are not adding volumemount to the container and so `vol` will be nil here.

pkg/csi/rbddaemonset.go
```diff
@@ -52,349 +51,295 @@ var (
...
-									Name:      "ceph-csi-kms-config",
-									MountPath: "/etc/ceph-csi-encryption-kms-config/",
-									ReadOnly:  true,
-								},
-								{
-									Name:      "oidc-token",
-									MountPath: "/run/secrets/tokens",
-									ReadOnly:  true,
-								},
+var rbdDeamonSetSpec = appsv1.DaemonSetSpec{
```
leelavg, line 58:
```suggestion
var rbdDaemonSetSpec = appsv1.DaemonSetSpec{
```

pkg/csi/cephfsdaemonset.go
```diff
@@ -23,13 +23,14 @@ import (
...
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 )
 
 const (
-	CephFSDamonSetName = "csi-cephfsplugin"
+	CephFSDaemonSetName = "csi-cephfsplugin"
+
+	cephFSDamonSetContainerName = "csi-cephfsplugin"
```
leelavg, line 33: I don't think we'll ever want the name of deployment and main pod be different, can use same var, same for remaining.

pkg/csi/cephfsdaemonset.go
```diff
@@ -42,231 +43,235 @@ var (
...
+	for i := range ds.Spec.Template.Spec.Containers {
+		c := &ds.Spec.Template.Spec.Containers[i]
+		switch c.Name {
+		case templates.DriverRegistrar.Name:
+			c.Image = sidecarImages.ContainerImages.DriverRegistrarImageURL
+			c.Args = append(c.Args, fmt.Sprintf(
+				"--kubelet-registration-path=%s/plugins/%s/csi.sock",
+				templates.DefaultKubeletDirPath,
+				GetCephFSDriverName(),
+			))
```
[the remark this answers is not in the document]
nb-ohad: Yes, you are correct, with the move to a switch statement, I lost the deep copy operations. 
I will add it in the next push

pkg/csi/cephfsdaemonset.go
nb-ohad: The reason I added the container names is because it was not the same on all of them (I believe the culprit is `RDBDeployment`). With that there is no established pattern, so when there is no pattern I fall back to explicit naming across the board.

pkg/csi/rbddeployment.go
nb-ohad: Yep

pkg/csi/cephfsdaemonset.go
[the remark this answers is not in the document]
nb-ohad: done

pkg/csi/rbddaemonset.go
nb-ohad: Fixed now

pkg/csi/rbddaemonset.go
nb-ohad: done

leelavg (commented):
LGTM, didn't approve if others wants to review.

jarrpa (approved):

Madhu-1 (commented):
LGTM

Madhu-1 (approved):

PR #108, opened 2024-03-08
create default storageclaim corresponding to storageclient

leelavg:
at all times we want to ensure storageclaims which creates storageclass that are same as internal mode.

however, if there are more that one storageclient connected we need to provide a way for user to choose which storageclient should host default storageclasses and as such we don't reconcile storageclaims we created once they are deleted by user.

Example:
scenario 1: only one storageclient is connected
1. controller creates a storageclaim automatically which eventually creates storageclasses with already established names
2. user can delete this storageclaim and controller will not reconcile providing a way for user to manually create storageclaim

scenario 2: two storageclients are connected
1. controller creates a storageclaim for one of the storageclients i.e, whichever reconciles first
2. user can delete that automatically created storageclaim and from then controller will behave exactly as scenario 1

part of [RHSTOR-5553]

Depends on #103


leelavg:
now part of #116

PR #117, opened 2024-03-18
controllers: deploy CSI when there is atleast one storageclient

leelavg:
when client operator is deployed alongside ocs-op we need to ensure that we deploy and manage CSI only when running under provider mode.

checking for storageclient is the simplest way to achieve the requirement without depending on context outside of client operator deployment.

part of [RHSTOR-5487]


leelavg:
change in requirements so the PR is in draft.

leelavg:
taking a different route.

PR #115, opened 2024-03-18
implement webhook for validating ocs-client-operator subscription channel

leelavg:
provider sends it's subscription channel name as response to client heartbeat, we will store that in the corresponding storageclient as an annotation value.

implement a webhook server that validates whether the channel name in the subscription routed to it conforms to all the storageclients desired operator subscription placed on the storageclient as an annotation value.

after controller start up we register webhook and also label the subscription which ensures only updates on that specific subscription are routed to the webhook.

part of [RHSTOR-5496]

Depends on #89

PR #120, opened 2024-03-24
implement ocs-client-operator auto upgrade 

leelavg:
update ocs-client-operator subscription when all the storageclients agree to a single subscription channel.

part of [RHSTOR-5496]

Depends on #115

### DIFF
--- a/controllers/operatorconfigmap_controller.go
+++ b/controllers/operatorconfigmap_controller.go
@@ -23,6 +23,7 @@ import (
 	// The embed package is required for the prometheus rule files
 	_ "embed"
 
+	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/console"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/csi"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/templates"
@@ -134,6 +135,7 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&extv1.CustomResourceDefinition{}, enqueueConfigMapRequest, builder.OnlyMetadata).
 		Watches(&opv1a1.Subscription{}, enqueueConfigMapRequest, subscriptionPredicates).
 		Watches(&admrv1.ValidatingWebhookConfiguration{}, enqueueConfigMapRequest, webhookPredicates).
+		Watches(&v1alpha1.StorageClient{}, enqueueConfigMapRequest, builder.WithPredicates(predicate.AnnotationChangedPredicate{})).
 		Complete(c)
 }
 
@@ -175,6 +177,11 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	if err := labelClientOperatorSubscription(c); err != nil {
 		c.log.Error(err, "unable to label ocs client operator subscription")
+		return ctrl.Result{}, err
+	}
+
+	if err := c.reconcileSubscription(); err != nil {
+		c.log.Error(err, "unable to reconcile subscription")
 		return ctrl.Result{}, err
 	}
 
@@ -590,4 +597,69 @@ func labelClientOperatorSubscription(c *OperatorConfigMapReconciler) error {
 
 	c.log.Info("successfully labelled ocs-client-operator subscription")
 	return nil
+}
+
+func (c *OperatorConfigMapReconciler) reconcileSubscription() error {
+
+	storageClients := &v1alpha1.StorageClientList{}
+	if err := c.list(storageClients); err != nil {
+		return fmt.Errorf("failed to list storageclients: %v", err)
+	}
+
+	var desiredChannel string
+	for idx := range storageClients.Items {
+		// empty if annotation doesn't exist or else gets desired channel
+		channel := storageClients.
+			Items[idx].
+			GetAnnotations()[utils.DesiredSubscriptionChannelAnnotationKey]
+		// skip clients with no/empty desired channel annotation
+		if channel != "" {
+			// check if we already established a desired channel
+			if desiredChannel == "" {
+				desiredChannel = channel
+			}
+			// check for agreement between clients
+			if channel != desiredChannel {
+				desiredChannel = ""
+				// two clients didn't agree for a same channel and no need to continue further
+				break
+			}
+		}
+	}
+
+	if desiredChannel != "" {
+		subscriptions := &opv1a1.SubscriptionList{}
+		err := c.list(
+			subscriptions,
+			client.InNamespace(c.OperatorNamespace),
+			client.MatchingLabels{subscriptionLabelKey: subscriptionLabelValue},
+			client.Limit(1),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to list subscription for ocs-client-operator using labels: %v", err)
+		}
+
+		if len(subscriptions.Items) == 1 {
+			clientSubscription := &subscriptions.Items[0]
+			if desiredChannel != clientSubscription.Spec.Channel {
+				clientSubscription.Spec.Channel = desiredChannel
+				// TODO: https://github.com/red-hat-storage/ocs-client-operator/issues/130
+				// there can be a possibility that platform is behind, even then updating the channel will only make subscription to be in upgrading state
+				// without any side effects for already running workloads. However, this will be a silent failure and need to be fixed via above TODO issue.
+				if err := c.update(clientSubscription); err != nil {
+					return fmt.Errorf("failed to update subscription channel to %v: %v", desiredChannel, err)
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func (c *OperatorConfigMapReconciler) list(obj client.ObjectList, opts ...client.ListOption) error {
+	return c.List(c.ctx, obj, opts...)
+}
+
+func (c *OperatorConfigMapReconciler) update(obj client.Object, opts ...client.UpdateOption) error {
+	return c.Update(c.ctx, obj, opts...)
 }


leelavg:
review only last commit.

leelavg:
/hold

after ff.

controllers/clusterversion_controller.go
```diff
@@ -127,12 +129,25 @@ func (c *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		predicate.NewPredicateFuncs(
+			func(obj client.Object) bool {
+				annotations := obj.GetAnnotations()
+				if desiredChannel, ok := annotations[utils.DesiredSubscriptionChannelAnnotationKey]; ok {
+					return c.subscriptionChannel != desiredChannel
+				}
+				return false
+			},
+		),
+	)
```
nb-ohad, line 142: Having a predicate based on an internal (not external) reconciler state will not work the way you expect. It will filter events as they queue with unpredictable results. Why do we need to filter reconciliation on clients?

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
 		}
 	}
 
+	c.subscriptionChannel = sub.Spec.Channel
```
nb-ohad, line 627: This function is supposed to have no side-effect as it is a pure add-label operation. It is a bad add to have invisible side-effects to an establish flow

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
...
+	for idx := range storageClients.Items {
+		storageClient := &storageClients.Items[idx]
+		annotations := storageClient.GetAnnotations()
+		if annotatedChannel, ok := annotations[utils.DesiredSubscriptionChannelAnnotationKey]; ok {
+			if desiredChannel != "" && desiredChannel != annotatedChannel {
+				// not all storageclients desiring to be on same channel
+				return nil
+			}
+			desiredChannel = annotatedChannel
+		}
```
nb-ohad, line 650: The intent here is to skip empty annotation and extract a single value that meets all non-empty annotations. They way the code is written does not reflect the intent:

```suggestion
		// guarantee to return a string even if annotations is a nil map
		channel := storageClients.
			Items[idx].
			GetAnnotations()[utils.DesiredSubscriptionChannelAnnotationKey]

		// Skip clients with no/empty desired channel annotation
		if channel != "" {
			// Check if we already established a desired channel
			if desiredChannel == "" {
				desiredChannel = channel
			}

			// Check for agreement between clients	
			if channel != desiredChannel {
					desiredChannel = ""
					break
				} 
			}
		}
```

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
...
+	}
+
+	if desiredChannel == "" {
+		// no storageclients mention a desired channel, user can bump the channel manually if required
```
nb-ohad, line 654:
```suggestion
		// No desired channel specified or no agreement between clients
```

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
...
+		return nil
+	}
+
+	subscriptions := &opv1a1.SubscriptionList{}
+	// we are already setting this label during reconcile phases
```
nb-ohad, line 659: Not sure I understand this comment

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
...
+	err := c.list(subscriptions, client.InNamespace(c.OperatorNamespace), client.MatchingLabels{subscriptionLabelKey: subscriptionLabelValue})
+	if err != nil {
+		return fmt.Errorf("failed to list subscriptions with known labels: %v", err)
```
nb-ohad, line 662:
```suggestion
		return fmt.Errorf("failed to load subscription for ocs-client-operator%v", err)
```

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
...
+	}
+	if len(subscriptions.Items) != 1 {
+		return fmt.Errorf("failed to validate the presence and uniqueness of ocs-client-operator subscription")
+	}
```
nb-ohad, line 666: Why check and fail for that? what does it give us?

controllers/clusterversion_controller.go
```diff
@@ -604,6 +624,63 @@ func labelClientOperatorSubscription(c *ClusterVersionReconciler) error {
...
+
+	if desiredChannel == "" {
+		// no storageclients mention a desired channel, user can bump the channel manually if required
+		return nil
+	}
+
+	subscriptions := &opv1a1.SubscriptionList{}
+	// we are already setting this label during reconcile phases
+	err := c.list(subscriptions, client.InNamespace(c.OperatorNamespace), client.MatchingLabels{subscriptionLabelKey: subscriptionLabelValue})
```
nb-ohad, line 660: Add a count=1 to the list options, we assume a single subscription

nb-ohad (changes requested):

controllers/clusterversion_controller.go
leelavg: sry, I was also a bit skeptical to use it. My intention is to trigger a reconcile when the existing subscription channel is different compared to storageclient desired channel annotation.

I don't want to send reconciles every minute as job updates the annotations value of storageclient, what I could do is queue an event if annotation value changes which isn't fool proof though.

controllers/clusterversion_controller.go
leelavg: agreed! I was also doubtful.

controllers/clusterversion_controller.go
nb-ohad: This seems like a bug in the job. The actual API update call should only happen if the annotation has changed
We have a util that sets the annotation and returns a boolean to indicate if a change actually happened (in memory) and in most cases, we call `client.Update` only if that return value is true.

controllers/clusterversion_controller.go
leelavg: we want to ensure there's only one subscription for ocs-client-operator w/ the label set, if there are more than one, something from outside changed it and so failing to update it.

controllers/clusterversion_controller.go
leelavg: didn't add because of https://github.com/red-hat-storage/ocs-client-operator/pull/120#discussion_r1549652298

controllers/clusterversion_controller.go
nb-ohad: Again. I am not sure what the comment means

controllers/clusterversion_controller.go
nb-ohad: What? OCS can only be deployed once on a cluster. This is a basic assumption. You don't need to test for it

controllers/clusterversion_controller.go
leelavg: The label that I'm using here to list the ocs-client-operator subscription is set in another reconciler method and I'm simply stating that in the comment.

leelavg:
rebased and addressed comments.

controllers/clusterversion_controller.go
leelavg: done.

controllers/clusterversion_controller.go
leelavg: done.

controllers/clusterversion_controller.go
leelavg: done.

controllers/clusterversion_controller.go
leelavg: done.

controllers/clusterversion_controller.go
leelavg: done.

controllers/clusterversion_controller.go
leelavg: done.

controllers/clusterversion_controller.go
leelavg: done.

controllers/operatorconfigmap_controller.go
```diff
@@ -591,3 +598,62 @@ func labelClientOperatorSubscription(c *OperatorConfigMapReconciler) error {
...
+
+	storageClients := &v1alpha1.StorageClientList{}
+	if err := c.list(storageClients); err != nil {
+		return fmt.Errorf("failed to list storageclients: %v", err)
+	}
+
+	var desiredChannel string
+	for idx := range storageClients.Items {
+		// empty if annotation doesn't exist or else gets desired channel
+		channel := storageClients.Items[idx].GetAnnotations()[utils.DesiredSubscriptionChannelAnnotationKey]
```
nb-ohad, line 612: This is a long line, can you please split it?
```suggestion
		channel := storageClients.
			Items[idx].
			GetAnnotations()[utils.DesiredSubscriptionChannelAnnotationKey]
```

controllers/operatorconfigmap_controller.go
```diff
@@ -591,3 +598,62 @@ func labelClientOperatorSubscription(c *OperatorConfigMapReconciler) error {
...
+			}
+		}
+	}
+	if desiredChannel == "" {
+		// no desired channel specified or no agreement between clients
+		return nil
+	}
+
+	subscriptions := &opv1a1.SubscriptionList{}
+	err := c.list(subscriptions, client.InNamespace(c.OperatorNamespace), client.MatchingLabels{subscriptionLabelKey: subscriptionLabelValue}, client.Limit(1))
```
nb-ohad, line 633: Long line please split
```suggestion
	err := c.list(
		subscriptions, 
		client.InNamespace(c.OperatorNamespace), 
		client.MatchingLabels{subscriptionLabelKey: subscriptionLabelValue}, 
		client.Limit(1),
	)
```

controllers/operatorconfigmap_controller.go
```diff
@@ -591,3 +598,62 @@ func labelClientOperatorSubscription(c *OperatorConfigMapReconciler) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to list subscription for ocs-client-operator using labels: %v", err)
+	}
+	if len(subscriptions.Items) != 1 {
+		return fmt.Errorf("failed to load subscription for ocs-client-operator")
+	}
```
nb-ohad, line 639: Although the result is identical (because we are limiting the query to only 1 result), the intent is clearer if you use the `<` operator and not the `!=` operator.

controllers/operatorconfigmap_controller.go
```diff
@@ -591,3 +598,62 @@ func labelClientOperatorSubscription(c *OperatorConfigMapReconciler) error {
...
+			if channel != desiredChannel {
+				desiredChannel = ""
+				// two clients didn't agree for a same channel and no need to continue further
+				break
+			}
+		}
+	}
+	if desiredChannel == "" {
+		// no desired channel specified or no agreement between clients
+		return nil
```
nb-ohad, line 629: You are returning nil in all cases, and this is a non-error return value. In this type of cases, I would prefer to consolidate the return branching to a single return at the end of the function. You can achieve this easily by reverting the condition and nesting the happy path inside the if

controllers/operatorconfigmap_controller.go
nb-ohad, line 639: Should we consider it an error and halt all other reconciliation functions because we didn't find a subscription? In my eyes, it seems like an overkill. If there is no subscription we can assume the operator is not upgradable easily or at least not using OLM which means it will be really hard to get the version to ignore the restrictions.

nb-ohad (changes requested):

controllers/operatorconfigmap_controller.go
leelavg: yes, makes sense, not having a subscription will not effect the running operator.

controllers/operatorconfigmap_controller.go
leelavg: done.

controllers/operatorconfigmap_controller.go
leelavg: done.

controllers/operatorconfigmap_controller.go
leelavg: changed as per other comment.

controllers/operatorconfigmap_controller.go
leelavg: done.

controllers/operatorconfigmap_controller.go
leelavg: done.

leelavg:
tech debt for a specific case will be tracked in https://github.com/red-hat-storage/ocs-client-operator/issues/130

PR #121, opened 2024-03-25
RDR: setting client info

raaizik:
Writes to the newly added fields of `ReportStatusRequest` which are then exported by means of the changes introduced in [RDR: exporting client info #2504 ](https://github.com/red-hat-storage/ocs-operator/pull/2504)


umangachapagain (commented):
Looks good to me.

Which is the PR that creates the final ConfigMap?

service/status-report/main.go
```diff
@@ -138,7 +145,9 @@ func main() {
 
 	status := providerclient.NewStorageClientStatus().
 		SetPlatformVersion(pltVersion).
-		SetOperatorVersion(oprVersion)
+		SetOperatorVersion(oprVersion).
+		SetClusterID(string(clusterID)).
```
Madhu-1, line 149: why are we sending the clusterID again? i believe it should be available on the storageConsumer CR. isn't that the case?

Can't we send this information when onboarding not each time during the status report?

raaizik: See [discussion](https://github.com/red-hat-storage/ocs-operator/pull/2504#discussion_r1524166253)

raaizik:
> Looks good to me.
> 
> Which is the PR that creates the final ConfigMap?

@umangachapagain it's still a [WIP](https://github.com/red-hat-storage/ocs-operator/pull/2440)

service/status-report/main.go
raaizik: > See discussions [(2)](https://github.com/red-hat-storage/ocs-operator/pull/2504#discussion_r1524179758)

Madhu-1: Resolving it as got the answer

PR #131, opened 2024-04-04
update bundle w/ bumping version

leelavg:
the generated bundle is out of sync w/ the config & manifests, even though they can be rebuilt we can have the bundle fully removed or keep it upto date from time to time, this PR does the latter.

for some of the controllers kubebuilder tags were added but it wasn't updated in the config and so syncing everything once.


leelavg:
probably the bundle was out of sync as Makefile targets weren't working in M1 and so as this was recently fixed, we can either have some CI actions to enforce an updated bundle if any rbac/api is changed or can remove (~.gitignore) the bundle folder altogether.

Madhu-1 (approved):
we need to catch these kind of problems in the CI.

PR #133, opened 2024-04-05
make storageclient api to clusterscope resource

leelavg:
as of now storageclient api is namespaced w/ no known benefits that could outweigh ease of operations performed in controllers if it can be cluster scoped.

this PR makes storageclient to clusterscope and majorly benefits from existing tracking of clusterscoped resource (storageclaim) by namespaced resource (existing storageclient) replaced w/ owner references.

design wise, no two storageclient even if namespaced shouldn't be allowed to connect to single provider and as such storageclient better be clusterscoped and only con is for user to choose explicit names to avoid collision and that'd be less of an issue when compared to all the plumbing that is being done in controllers as of now.


leelavg:
reviewing on per commit basis will be good.

/hold

testing in progress as this needs changes at provider end as well to test.

api/v1alpha1/storageclaim_types.go
```diff
@@ -52,10 +54,10 @@ type StorageClientNamespacedName struct {
...
 type StorageClaimSpec struct {
 	//+kubebuilder:validation:Enum=block;sharedfile
-	Type             string                       `json:"type"`
-	EncryptionMethod string                       `json:"encryptionMethod,omitempty"`
-	StorageProfile   string                       `json:"storageProfile,omitempty"`
-	StorageClient    *StorageClientNamespacedName `json:"storageClient"`
+	Type              string  `json:"type"`
+	EncryptionMethod  string  `json:"encryptionMethod,omitempty"`
+	StorageProfile    string  `json:"storageProfile,omitempty"`
+	StorageClientName *string `json:"storageClient"`
```
nb-ohad, line 60: Why the Name prefix? I understand we are changing the type. buy why do we need to change the name of the field?

controllers/storageclaim_controller.go
```diff
@@ -595,8 +596,11 @@ func (r *StorageClaimReconciler) delete(obj client.Object) error {
 }
 
 func (r *StorageClaimReconciler) own(resource metav1.Object) error {
-	// Ensure StorageClaim ownership on a resource
-	return controllerutil.SetOwnerReference(r.storageClaim, resource, r.Scheme)
+	return r.ownWithOwner(r.storageClaim, resource)
+}
+
+func (r *StorageClaimReconciler) ownWithOwner(owner, dependent metav1.Object) error {
+	return controllerutil.SetControllerReference(owner, dependent, r.Scheme)
```
nb-ohad, line 603: This util does not make sense? what is the value added by it? why can we call `controllerutil.SetControllerReference(owner, dependent, r.Scheme)` on the call site? where we know the owner and the dependent.

In contrast, `r.own` has the added value of simplifying the operation of adding the object pointed by the reconciled event as the owner which is a common use case.

controllers/storageclient_controller.go
```diff
@@ -76,58 +66,49 @@ const (
...
 			res = append(res, o.(*v1alpha1.StorageClient).Spec.StorageProviderEndpoint)
 		}
 		return res
-	})
-
-	if err := mgr.GetCache().IndexField(ctx, &v1alpha1.StorageClaim{}, storageClientAnnotationIndexName, func(obj client.Object) []string {
-		return []string{obj.GetAnnotations()[storageClientAnnotationKey]}
 	}); err != nil {
-		return fmt.Errorf("unable to set up FieldIndexer for storageclient annotation: %v", err)
+		return fmt.Errorf("unable to set up FieldIndexer for storageclient providerendpoint: %v", err)
```
nb-ohad, line 92:
```suggestion 
                return fmt.Errorf("unable to set up FieldIndexer for StorageClient's ProviderEndpoint field: %v", err)
```

controllers/storageclient_controller.go
```diff
@@ -76,58 +66,49 @@ const (
...
+	if err := mgr.GetCache().IndexField(ctx, &v1alpha1.StorageClaim{}, ownerIndexName, func(obj client.Object) []string {
+		refs := obj.GetOwnerReferences()
+		var owners []string
+		for i := range refs {
+			owners = append(owners, string(refs[i].UID))
+		}
+		return owners
 	}); err != nil {
-		return fmt.Errorf("unable to set up FieldIndexer for storageclient owner annotation: %v", err)
+		return fmt.Errorf("unable to set up FieldIndexer for storageclaim owner: %v", err)
```
nb-ohad, line 103:
```suggestion
                return fmt.Errorf("unable to set up FieldIndexer for StorageClaim's owner uid: %v", err)
```

controllers/storageclient_controller.go
```diff
@@ -177,110 +155,95 @@ func (s *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if err := r.Client.List(r.ctx, storageClientList, storageClientListOption...); err != nil {
+		r.Log.Error(err, "unable to list storage clients")
 		return ctrl.Result{}, err
 	}
 
 	if len(storageClientList.Items) > 1 {
-		s.Log.Info("one StorageClient is allowed per namespace but found more than one. Rejecting new request.")
-		instance.Status.Phase = v1alpha1.StorageClientFailed
+		r.Log.Info("one StorageClient is allowed with a unique storageproviderendpoint but found more than one. Rejecting new request.")
+		r.storageClient.Status.Phase = v1alpha1.StorageClientFailed
```
nb-ohad, line 172: Why are you trying to protect against 2 clients on the same cluster connecting to the same provider twice? I can see use cases for such configurations and I don't; see any reason to disallow it. 

The only limitation should be that we should not allow two client CR having the same name. But that is already covered by k8s itself.

controllers/storageclient_controller.go
```diff
@@ -177,110 +155,95 @@ func (s *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 
 	// deletion phase
-	if !instance.GetDeletionTimestamp().IsZero() {
-		return s.deletionPhase(instance, externalClusterClient)
+	if !r.storageClient.GetDeletionTimestamp().IsZero() {
+		return r.deletionPhase(externalClusterClient)
 	}
 
 	// ensure finalizer
-	if !contains(instance.GetFinalizers(), storageClientFinalizer) {
```
nb-ohad, line 210: What is the reason for the removal of the contains check?

controllers/storageclient_controller.go
```diff
@@ -289,21 +252,22 @@ func (s *StorageClientReconciler) newExternalClusterClient(instance *v1alpha1.St
...
 
 	// TODO Need to find a way to get rid of ClusterVersion here as it is OCP
 	// specific one.
 	clusterVersion := &configv1.ClusterVersion{}
-	err := s.Client.Get(s.ctx, types.NamespacedName{Name: "version"}, clusterVersion)
+	clusterVersion.Name = "version"
+	err := r.get(clusterVersion)
 	if err != nil {
-		s.Log.Error(err, "failed to get the clusterVersion version of the OCP cluster")
+		r.Log.Error(err, "failed to get the clusterVersion version of the OCP cluster")
```
nb-ohad, line 263:
```suggestion
	if err := r.get(clusterVersion); err != nil {
		r.Log.Error(err, "failed to get the clusterVersion version of the OCP cluster")

File: controllers/storageclient_controller.go [hunk 8]
```diff
@@ -314,65 +278,65 @@ func (s *StorageClientReconciler) onboardConsumer(instance *v1alpha1.StorageClie
...
 
-func (s *StorageClientReconciler) verifyNoStorageClaimsExist(instance *v1alpha1.StorageClient) error {
+func (r *StorageClientReconciler) verifyNoStorageClaimsExist() error {
 
 	storageClaims := &v1alpha1.StorageClaimList{}
-	err := s.Client.List(s.ctx,
+	err := r.Client.List(r.ctx,
 		storageClaims,
-		client.MatchingFields{storageClientAnnotationIndexName: client.ObjectKeyFromObject(instance).String()},
+		client.MatchingFields{ownerIndexName: r.storageClient.Name},
```
nb-ohad, line 339: You should track by owner UID, not owner name. OwnerRefernace might be stale and pointing to a previous storage client that had a similar name.

controllers/storageclient_controller.go
```diff
@@ -526,98 +463,61 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
...
-
-	if len(claimOwners.Items) == 0 {
-		// no other storageclient claims as an owner and take responsibility of creating the default claims by becoming owner
-		if utils.AddAnnotation(instance, defaultClaimsOwnerAnnotationKey, "true") {
-			if err := s.update(instance); err != nil {
-				return fmt.Errorf("not able to claim ownership of creating default storageclaims: %v", err)
-			}
-		}
-	}
+func (r *StorageClientReconciler) reconcileStorageClaims() error {
```
nb-ohad, line 470: not that this func became trivial, can we please move the 2 calls into the call site and rename them from `create` to `reconcile`?

api/v1alpha1/storageclaim_types.go
```diff
@@ -44,6 +44,8 @@ type StorageClaimStatus struct {
 	Phase storageClaimState `json:"phase,omitempty"`
 }
 
+// TODO: remove after storageclassclaim migration controller and storageclassclaim is deleted
+
```
nb-ohad, line 48: This makes no sense, as this type is not used anymore in the code base. Why do we need to keep it?

controllers/storageclaim_controller.go
```diff
@@ -152,7 +152,7 @@ func (r *StorageClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	r.storageClaimHash = getMD5Hash(r.storageClaim.Name)
 	r.storageClaim.Status.Phase = v1alpha1.StorageClaimInitializing
 
-	if r.storageClaim.Spec.StorageClient == nil {
+	if r.storageClaim.Spec.StorageClientName == nil {
```
nb-ohad, line 155: When working with string it is preferable to check against the empty string
```suggestion
        if r.storageClaim.Spec.StorageClientName == "" {
```

controllers/storageclaim_controller.go
```diff
@@ -267,12 +266,15 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
-			r.log.Info("Finalizer not found for StorageClaim. Adding finalizer.", "StorageClaim", r.storageClaim.Name)
-			r.storageClaim.SetFinalizers(append(r.storageClaim.GetFinalizers(), storageClaimFinalizer))
+		if controllerutil.AddFinalizer(r.storageClaim, storageClaimFinalizer) {
 			updateStorageClaim = true
 		}
-		if utils.AddAnnotation(r.storageClaim, storageClientAnnotationKey, client.ObjectKeyFromObject(r.storageClient).String()) {
+
+		// if claim is created by default we'll already have a controller reference, however set the owner if it is explictly created by user
+		if !controllerutil.HasControllerReference(r.storageClaim) {
+			if err := r.ownWithOwner(r.storageClient, r.storageClaim); err != nil {
```
nb-ohad, line 275: Claims created by the user are owned by the user, not by the client. We should not add an owner reference in that case

api/v1alpha1/storageclaim_types.go
nb-ohad, line 60: Strings can be empty and nil. There should be a really good reason having an API field be defined as a pointer to a string. I don't see any reason to do it here.

controllers/storageclassclaim_migration_controller.go
```diff
@@ -82,7 +82,7 @@ func (r *StorageClassClaimMigrationReconciler) Reconcile(ctx context.Context, re
 
 	storageClaim.Spec.EncryptionMethod = storageClassClaim.Spec.EncryptionMethod
 	storageClaim.Spec.StorageProfile = storageClassClaim.Spec.StorageProfile
-	storageClaim.Spec.StorageClient = storageClassClaim.Spec.StorageClient.DeepCopy()
```
nb-ohad, line 85: What about name collision? shouldn't we compose a new name based on both the namespace and the name?

service/status-report/main.go
```diff
@@ -146,7 +135,7 @@ func main() {
 		SetPlatformVersion(pltVersion).
 		SetOperatorVersion(oprVersion).
 		SetClusterID(string(clusterID)).
-		SetNamespacedName(namespacedName.String())
+		SetNamespacedName(storageClientName)
```
nb-ohad, line 138: Is that an upcoming change to rename the API correctly?

nb-ohad (changes requested):

service/status-report/main.go
leelavg: Yes, #134

controllers/storageclassclaim_migration_controller.go
leelavg: I believe this whole migration controller should go away.

api/v1alpha1/storageclaim_types.go
leelavg: I didn't change any, https://ibm.enterprise.slack.com/archives/C065Q8HCSSC/p1709787993883729
tl:dr; version is, if client field is empty we pick existing client, however this field can't be empty due to crd cel rule and didn't touch due to reason mentioned in above thread.

controllers/storageclaim_controller.go
leelavg: Ack, so during deletion of client I'll check for all claims referring to that client then.

controllers/storageclaim_controller.go
leelavg: It is a pointer atm.

api/v1alpha1/storageclaim_types.go
leelavg: It is used in storageclassclaim, todo mentions that.

controllers/storageclient_controller.go
leelavg: Yes, actual bug in my code!

controllers/storageclient_controller.go
leelavg: I didn't impose the restriction, I added a comment to surface the obvious code surrounding it. Need to check why it was imposed in the first place.

controllers/storageclaim_controller.go
leelavg: Will be removed as asked in some other comment.

api/v1alpha1/storageclaim_types.go
leelavg: No legit reason, only explicit mention.

controllers/storageclassclaim_migration_controller.go
leelavg: I don't have any info when this migratiom controller is useful.

Simply migrating the CR from storageclassclaim to storageclaim may create new resources (rns & svg) which are fundamentally diff and so why can't the migration be carried out during at one place ie, migration script.

service/status-report/main.go
nb-ohad: ok

api/v1alpha1/storageclaim_types.go
nb-ohad: Let's remove it. This is part of a k8s API we need to be consistent. 
(e.g we have `StoageProfile` and not  `StoageProfileName`, etc.)

controllers/storageclient_controller.go
nb-ohad: It was imposed for 2 clients on the same namespace as we wanted a namespace to govern a single client. That is not the case anymore.

api/v1alpha1/storageclaim_types.go
nb-ohad: Ha. ok. I understand now. Resolving.

controllers/storageclaim_controller.go
nb-ohad: This is confusing. Is it the pointer that is nill or the string that is nil (the storage container inside the string)? 

Of course there is only one correct answer:
`r.storageClaim.Spec.StorageClientName == nil` the pointer is nil
`*r.storageClaim.Spec.StorageClientName == nil` the string is nil

But I hope you do understand why it is so confusing. As mentioned in another comment, I don't see the value of having a pointer to a string in the first place, and this is a good example why we should not add unnecessary complexity

controllers/storageclaim_controller.go
nb-ohad: Yes

api/v1alpha1/storageclaim_types.go
nb-ohad: But you did. Before the change, it was an embedded structure. That meant that in order to make it optional we needed to use a pointer to a struct. But now it is a plain string, and strings can be nil (or empty) without us needing to add a supporting pointer.

controllers/storageclassclaim_migration_controller.go
nb-ohad: Ok. Can you open an issue about the necessity/lack of necessity of the entire migration controller? link it here and I will resolve this.

api/v1alpha1/storageclaim_types.go
leelavg: string can't be nil, it has an empty value, isn't it?

controllers/storageclaim_controller.go
leelavg: I don't understand, a pointer to string can be nil but a string can't be nil.

controllers/storageclaim_controller.go
leelavg: Ack, if client is not owning claim then I need to add anno/label on claim to track it.

However, if I can make use of CRD cel rule which already exist I can make use of every client having client name in its spec which also removes the other pointer to string business.

controllers/storageclient_controller.go
nb-ohad: It is, but why does that matter?

controllers/storageclaim_controller.go
nb-ohad: You are correct a string var cannot be assigned nil, I was thinking about slices. But regardless empty string will serve us exactly like nil. And as empty string is the default value for a string it means that if the user does no provide a value we will get an empty string in the field.

controllers/storageclaim_controller.go
nb-ohad: Why? the claim already has it in the spec. Why do we need to add an additional ref?

api/v1alpha1/storageclaim_types.go
nb-ohad: Yes, you are correct (regarding string and ni values), it was a mistake on my side. But regardless, a string empty value has the same meaning in this case. If the user does not provide a claim name or provide an empty claim name we should react in exactly the same way.  And, assuming we defined the type correctly, omitted claim names will get deserialized into an empty string.

controllers/storageclient_controller.go
leelavg: It matters because off boarding is per storageconsumer, isn't it?

api/v1alpha1/storageclaim_types.go
leelavg: > ommited claim names
- that's the interesting part, as per cel rule on CRD there can never be a ommited claim name
- we weren't creating claims per client earlier however we are doing that now. Given that, should we provide an option to have empty client name in claim which wasn't possible upto now?

controllers/storageclaim_controller.go
leelavg: as stated in https://github.com/red-hat-storage/ocs-client-operator/pull/133#discussion_r1555153525, client field in claim spec can be empty.

leelavg: I see another issue, if we create claim we should be the one to delete it, if user creates it then user should delete it, however having two entities deleting similar resources may not be a good idea.

can we provide the whole deletion to user and not reconcile even if claim corresponding to a client is deleted by tracking via anno.

controllers/storageclassclaim_migration_controller.go
leelavg: done, https://github.com/red-hat-storage/ocs-client-operator/issues/135

leelavg:
/hold

There's one more point in my saved notes around claim deletion, will be posting that commit soon.

leelavg:
fixed commit lint.

api/v1alpha1/storageclaim_types.go
leelavg: done.

controllers/storageclaim_controller.go

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclient_controller.go
leelavg: raised #142

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclaim_controller.go
leelavg: done, change to string.

controllers/storageclaim_controller.go
leelavg: done, operator removes auto created claims during client offboard.

api/v1alpha1/storageclaim_types.go
leelavg: done.

controllers/storageclassclaim_migration_controller.go
```diff
@@ -82,7 +82,7 @@ func (r *StorageClassClaimMigrationReconciler) Reconcile(ctx context.Context, re
...
+	storageClaim.Spec.StorageClient = storageClassClaim.Spec.StorageClient.DeepCopy().Name
```
nb-ohad, line 85: DeepCopy to get a string field? is that necessary?

controllers/storageclient_controller.go
```diff
@@ -526,98 +456,63 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
...
 
-func (s *StorageClientReconciler) createDefaultBlockStorageClaim(instance *v1alpha1.StorageClient) error {
-	storageclaim := &v1alpha1.StorageClaim{}
-	storageclaim.Name = defaultBlockStorageClaim
-	storageclaim.Spec.Type = "block"
-	storageclaim.Spec.StorageClient = &v1alpha1.StorageClientNamespacedName{
-		Name:      instance.Name,
-		Namespace: instance.Namespace,
+func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
+	if r.storageClient.GetAnnotations()[storageClaimProcessedAnnotationKey] == "true" {
```
nb-ohad, line 498: The annotation concern is added at a upper level. I would expect the guard to be outside of the reconciler function and not inside it.

controllers/storageclient_controller.go
```diff
@@ -526,98 +456,63 @@ func (s *StorageClientReconciler) reconcileClientStatusReporterJob(instance *v1a
...
 
-func (s *StorageClientReconciler) reconcileDefaultStorageClaims(instance *v1alpha1.StorageClient) error {
-
-	if instance.GetAnnotations()[defaultClaimsProcessedAnnotationKey] == "true" {
-		// we already processed default claims for this client
+func (r *StorageClientReconciler) reconcileBlockStorageClaim() error {
+	if r.storageClient.GetAnnotations()[storageClaimProcessedAnnotationKey] == "true" {
+		// we already processed claim creation for this client
 		return nil
 	}
```
nb-ohad, line 483: The annotation concern is added at a upper level. I would expect the guard to be outside of the reconciler function and not inside it.

nb-ohad (changes requested):

controllers/storageclient_controller.go
leelavg: makes sense, will do.

controllers/storageclient_controller.go
leelavg: makes sense, will do.

leelavg: done.

controllers/storageclient_controller.go
leelavg: done.

controllers/storageclassclaim_migration_controller.go
leelavg: done.

controllers/storageclient_controller.go
```diff
@@ -177,110 +140,108 @@ func (s *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		if err := r.reconcileSharedfileStorageClaim(); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	if utils.AddAnnotation(r.storageClient, storageClaimProcessedAnnotationKey, "true") {
+		if err := r.update(r.storageClient); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with claim processed annotation: %v", err)
+		}
 	}
```
nb-ohad, line 205:
```suggestion
	if r.storageClient.GetAnnotations()[storageClaimProcessedAnnotationKey] != "true" {
		if err := r.reconcileBlockStorageClaim(); err != nil {
			return reconcile.Result{}, err
		}

		if err := r.reconcileSharedfileStorageClaim(); err != nil {
			return reconcile.Result{}, err
		}
		
		utils.AddAnnotation(r.storageClient, storageClaimProcessedAnnotationKey, "true")
		if err := r.update(r.storageClient); err != nil {
			return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with claim processed annotation: %v", err)
		}
	}
```

nb-ohad (changes requested):

controllers/storageclient_controller.go
leelavg: done.

Issue #82, opened 2024-02-20
RBAC permissions in ClusterVersion controllers


bernerhat:
RBAC permissions seem excessive in the ClusterVersion controller and need to be revisited to make sure that all of them are being used and are necessary.

leelavg:
already done in #260

PR #141, opened 2024-04-15
RBAC permissions alignment

bernerhat:
addition and removale of rbac permissions based on operations handled by each controller, also handles issue #82


leelavg (commented):
could you pls update the rbac in the bundle as well?

leelavg:
closing as these are already done as part of #260

PR #143, opened 2024-04-16
Refactors status Setters and restricts RBAC

raaizik:
## Changes

- Completion of the debts from the TODOs in #127
   - Revises permissions to include all CMs accessed by `status-reporter`

[RHSTOR-5500](https://issues.redhat.com/browse/RHSTOR-5500)

### DIFF
--- a/service/status-report/main.go
+++ b/service/status-report/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"github.com/red-hat-storage/ocs-operator/v4/services/provider/interfaces"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"os"
@@ -90,64 +91,6 @@ func main() {
 		klog.Exitf("Failed to get storageClient %q/%q: %v", storageClient.Namespace, storageClient.Name, err)
 	}
 
-	var oprVersion string
-	csvList := opv1a1.ClusterServiceVersionList{}
-	if err = cl.List(ctx, &csvList, client.InNamespace(operatorNamespace)); err != nil {
-		klog.Warningf("Failed to list csv resources: %v", err)
-	} else {
-		item := utils.Find(csvList.Items, func(csv *opv1a1.ClusterServiceVersion) bool {
-			return strings.HasPrefix(csv.Name, csvPrefix)
-		})
-		if item != nil {
-			oprVersion = item.Spec.Version.String()
-		}
-	}
-	if oprVersion == "" {
-		klog.Warningf("Unable to find csv with prefix %q", csvPrefix)
-	}
-
-	var pltVersion string
-	var clusterID configv1.ClusterID
-	clusterVersion := &configv1.ClusterVersion{}
-	clusterVersion.Name = "version"
-	if err = cl.Get(ctx, types.NamespacedName{Name: clusterVersion.Name}, clusterVersion); err != nil {
-		klog.Warningf("Failed to get clusterVersion: %v", err)
-	} else {
-		item := utils.Find(clusterVersion.Status.History, func(record *configv1.UpdateHistory) bool {
-			return record.State == configv1.CompletedUpdate
-		})
-		if item != nil {
-			pltVersion = item.Version
-		}
-		clusterID = clusterVersion.Spec.ClusterID
-	}
-	if pltVersion == "" {
-		klog.Warningf("Unable to find ocp version with completed update")
-	}
-
-	clusterConfig := &corev1.ConfigMap{}
-	clusterConfig.Name = clusterConfigName
-	clusterConfig.Namespace = clusterConfigNamespace
-
-	if err = cl.Get(ctx, client.ObjectKeyFromObject(clusterConfig), clusterConfig); err != nil {
-		klog.Warningf("Failed to get clusterConfig %q/%q: %v", clusterConfig.Namespace, clusterConfig.Name, err)
-	}
-
-	clusterMetadataYAML := clusterConfig.Data["install-config"]
-	clusterMetadata := struct {
-		Metadata struct {
-			Name string `yaml:"name"`
-		} `yaml:"metadata"`
-	}{}
-	err = yaml.Unmarshal([]byte(clusterMetadataYAML), &clusterMetadata)
-	if err != nil {
-		klog.Warningf("Fatal error, %v", err)
-	}
-	clusterName := ""
-	if len(clusterMetadata.Metadata.Name) > 0 {
-		clusterName = clusterMetadata.Metadata.Name
-	}
-
 	providerClient, err := providerclient.NewProviderClient(
 		ctx,
 		storageClient.Spec.StorageProviderEndpoint,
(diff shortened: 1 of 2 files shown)


leelavg:
/hold
May not be able to review now, holding as I believe change in rbac creates regression as you can't get csi-config anymore

raaizik:
> /hold May not be able to review now, holding as I believe change in rbac creates regression as you can't get csi-config anymore

@leelavg what if I remove that and instead add a `kubebuilder` marker above the function that needs to read the resource?

leelavg:
No, kube builder tags will not work for custom sa that aren't part of controllers.

Most straightforward is to include all the resources that status reporter is interested in across all namespaces or create a role & role binding in target ns but referencing sa of status reporter in operator ns.

service/status-report/main.go
```diff
@@ -202,3 +144,79 @@ func main() {
...
+
+	clusterMetadataYAML := clusterConfig.Data["install-config"]
+	clusterMetadata := struct {
+		Metadata struct {
+			Name string `yaml:"name"`
+		} `yaml:"metadata"`
+	}{}
+	err := yaml.Unmarshal([]byte(clusterMetadataYAML), &clusterMetadata)
+	if err != nil {
+		klog.Warningf("Fatal error, %v", err)
```
leelavg, line 174: I know there was no change in this sentence, this now looks a bit odd. We are logging the message which has `Fatal` at `Warning` level, since our intention is to find the name if it exists, could you pls change errors to reflect failure in unmarshal?

service/status-report/main.go
```diff
@@ -202,3 +144,79 @@ func main() {
...
+	}
+	clusterName := ""
+	if len(clusterMetadata.Metadata.Name) > 0 {
+		clusterName = clusterMetadata.Metadata.Name
```
leelavg, line 178: following similar approach in this file, use the condition to only log the warning.

service/status-report/main.go
```diff
@@ -202,3 +144,79 @@ func main() {
...
+	}
+	status.SetClusterID(string(clusterID)).SetClusterName(clusterName)
```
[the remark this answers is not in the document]
raaizik: See [comment](https://github.com/red-hat-storage/ocs-client-operator/pull/127#issuecomment-2042299805)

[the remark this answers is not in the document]
leelavg: I've asked to set the values in the same func but at different places wherever they are found. Here you are setting everything at once, at the end.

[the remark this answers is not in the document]
raaizik: I misread that, thanks for clarifying ✅

service/status-report/main.go
leelavg: not changed, I meant
```
if empty {
 log a warning that name is not found
}
```

leelavg: @raaizik could you pls flip the condition as I mentioned in my comment and I'd be able to merge. 

From
```
var name string
if not empty {
  set name from meta
} else {
  log warning
}
setStatus(name)
```

To
```
If empty {
  log warning
}
setStatus(name from meta) // no need for extra var
```

raaizik: Your suggestion doesn't capture the case of a nil `clusterMetadata.Metadata` while mine does.

leelavg: why would `clusterMetadata.Metadata` be nil as I see no field in the struct (or struct itself) will have a zero value of `nil`? Not related, just a point, if `clusterMetadata.Metadata` can be `nil` then existing `len(clusterMetadata.Metadata.Name)` would panic already.

I'm not able to see how my suggestion will fail known cases or it is incorrect, let me know otherwise, thanks.

``` go
package main

import (
	"fmt"

	"k8s.io/apimachinery/pkg/util/yaml"
)

type meta struct {
	Metadata struct {
		Name string `yaml:"name"`
	} `yaml:"metadata"`
}

var noData string

var noMeta = `
---
apiVersion: v1
controlPlane:
  replicas: 1
networking:
  machineNetwork:
platform:
  none: {}
`

var noName = `
---
additionalTrustBundlePolicy: Proxyonly
apiVersion: v1
baseDomain: example.com
metadata:
  creationTimestamp: null
networking:
  clusterNetwork:
  - cidr: 10.0.0.0/14
    hostPrefix: 23
`

var withName = `
---
additionalTrustBundlePolicy: Proxyonly
apiVersion: v1
baseDomain: example.com
metadata:
  creationTimestamp: null
  name: clusterName
networking:
  clusterNetwork:
  - cidr: 10.0.0.0/14
    hostPrefix: 23
`

func main() {
	var err error

	noData_ := &meta{}
	if err = yaml.Unmarshal([]byte(noData), noData_); err != nil {
		panic("failed unmarshal")
	}
	fmt.Printf("noData: %q\n", noData_.Metadata.Name)

	noMeta_ := &meta{}
	if err = yaml.Unmarshal([]byte(noMeta), noMeta_); err != nil {
		panic("failed unmarshal")
	}
	fmt.Printf("noMeta: %q\n", noMeta_.Metadata.Name)

	noName_ := &meta{}
	if err = yaml.Unmarshal([]byte(noName), noName_); err != nil {
		panic("failed unmarshal")
	}
	fmt.Printf("noName: %q\n", noName_.Metadata.Name)

	withName_ := &meta{}
	if err = yaml.Unmarshal([]byte(withName), withName_); err != nil {
		panic("failed unmarshal")
	}
	fmt.Printf("withName: %q\n", withName_.Metadata.Name)

	// $ go run main.go
	// noData: ""
	// noMeta: ""
	// noName: ""
	// withName: "clusterName"
}
```

raaizik: > why would

Not about why, it's about what if. 

> would panic already

And that is my concern, that the flow would resume to the line of setting the status even if, for example, `Name` is nil (condition doesn't panic in this case).
```go
	clusterName := ""
	if len(clusterMetadata.Metadata.Name) == 0 {
		klog.Warningf("Cluster name is empty, %v", err)
	}
	status.SetClusterName(clusterMetadata.Metadata.Name)
```

leelavg: > it's about what if.
- still, why would the zero value of a string (here `clusterMetadataYAML`) be `nil` and none of the fields of `clusterMetadata` struct can have a zero value of `nil` the field types are `strings` only?

nb-ohad: @raaizik @leelavg 
`Name` is of type `string` not `*string` so it cannot accept a `nil` value only an empty string value. That means you will never get to a point where accessing `.Name` will panic.  The only question left is what should we do in case we have an empty string for a name but I think that is already answered above with the discussion about having an warning at that case

raaizik: Changed

leelavg: as discussed offline, I hope below helps
``` go
package main

import (
	"fmt"

	"k8s.io/apimachinery/pkg/util/yaml"
)

type pointerMeta struct {
	Metadata *struct {                // here the type is a pointer
		Name string `yaml:"name"`
	} `yaml:"metadata"`
}

type valueMeta struct {
	Metadata struct {                // here the type is a struct
		Name string `yaml:"name"`
	} `yaml:"metadata"`
}

var noMeta = `
---
apiVersion: v1
controlPlane:
  replicas: 1
networking:
  machineNetwork:
platform:
  none: {}
`

func main() {
	var err error

	noMetaValue_ := &valueMeta{}
	if err = yaml.Unmarshal([]byte(noMeta), noMetaValue_); err != nil {
		panic("failed unmarshal")
	}
	fmt.Printf("noMetaValue_: %q\n\n", noMetaValue_.Metadata.Name)

	noMetaPtr_ := &pointerMeta{}
	if err = yaml.Unmarshal([]byte(noMeta), noMetaPtr_); err != nil {
		panic("failed unmarshal")
	}
	fmt.Printf("noMetaPtr_: %q\n", noMetaPtr_.Metadata.Name)

	// $ go run main.go
	// noMetaValue_: ""

	// panic: runtime error: invalid memory address or nil pointer dereference
	// [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4fbf78]

	// goroutine 1 [running]:
	// main.main()
	//         /tmp/meta/main.go:45 +0xf8
	// exit status 2
}
```

leelavg (approved):

PR #145, opened 2024-04-19
make value of storageclaim type case insensitive and consistent

leelavg:
value of storageclaim.spec.type can be `sharedfile` or `block` and should be case insensitive for user to choose from.

for gRPC comms we are using `sharedFile` and `block` as those are the correct abstracting names for the underlying storage.

Depends on red-hat-storage/ocs-operator#2570


leelavg:
/hold

for dependency to be merged.

api/v1alpha1/storageclaim_types.go
```diff
@@ -53,7 +53,7 @@ type StorageClientNamespacedName struct {
 
 // StorageClaimSpec defines the desired state of StorageClaim
 type StorageClaimSpec struct {
-	//+kubebuilder:validation:Enum=block;sharedfile
+	//+kubebuilder:validation:XValidation:rule="self.lowerAscii()=='block'||self.lowerAscii()=='sharedfile'",message="value should be either 'sharedfile' or 'block' (case insensitive)"
```
Madhu-1, line 56:
```suggestion
//+kubebuilder:validation:Enum=block;sharedFile
```
@leelavg i dont see a reason to change the CEL validation due to the internal details. if there is a problem with `shreadfile`, we can change it to `sharedFile` but we need to have strict validation for the user input.

leelavg: we shouldn't be enforcing the user to choose b/n letters casing, CEL rule allows casing but still enforces the values. I'm not creating rule for internal details, it's providing way for user not to think about what the case should be.

I could mention all the variations in enum as done in ocs for some different field https://github.com/red-hat-storage/ocs-operator/blob/cc58086ab8ed14965e6277c865d6aaa34c8cc49f/api/v1/storagecluster_types.go#L52
 but having CEL rule greatly simplifies the requirement and indicates the intention.

@Madhu-1 ^^

leelavg: > SHaReDFiLE BLOck etc
- I know, should that effect on how we reconcile, is it changing the intent on what we need to do beside the point user is really doing that?

Madhu-1: It wont but we also need to think about how to keep our CRD as we want, not giving more than expected flexibility to users where CR's looks ugly. i have never seen that kube objects, do you have any example for it?

Madhu-1: i think you are missing the point, this is not some user-chosen name like CR names, this is the field we are enforcing how it looks like and how it should be, we don't need to give this much flexibility to the user.

IMHO we should keep it as `sharedFile` and `block` nothing else. @nb-ohad thought?

nb-ohad: As this is a customer-facing CR,  I would like to remove the unnecessary strictness a user needs to consider when filling in the CR. 

A user-friendliness consideration will dictate that all of the values below should be ok:
- sharedfile
- SharedFile
- Sharedfile
- sharedFile

I am not concerned about a user abusing the system to have values like sHaReDfIle.  If someone set the value to be something like that then it was done on purpose, and as a whole, the system is resilient enough to handle such values.

nb-ohad, line 56: Let's not confuse the user with unnecessary options in our messages, it will just land a hand for abuse.  The fact that it is case insensitive does not mean we need to publicly announce it as such

```suggestion
	//+kubebuilder:validation:XValidation:rule="self.lowerAscii()=='block'||self.lowerAscii()=='sharedfile'",message="value should be either 'sharedfile' or 'block'"
```

controllers/storageclaim_controller.go
```diff
@@ -274,13 +274,13 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 
 		// storageClaimStorageType is the storage type of the StorageClaim
 		var storageClaimStorageType providerclient.StorageType
-		switch r.storageClaim.Spec.Type {
+		switch strings.ToLower(r.storageClaim.Spec.Type) {
```
nb-ohad, line 277: Please calculate the lowered case value only once and use it in both places.

nb-ohad (changes requested):

api/v1alpha1/storageclaim_types.go
leelavg: makes sense, done.

controllers/storageclaim_controller.go

leelavg:
do not review last commit, it was a bundle update w/ all missing changes. CI is yet to be improved to have bundle check w/ each PR.

controllers/storageclaim_controller.go
```diff
@@ -274,13 +274,14 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 
 		// storageClaimStorageType is the storage type of the StorageClaim
 		var storageClaimStorageType providerclient.StorageType
-		switch r.storageClaim.Spec.Type {
+		claimType := strings.ToLower(r.storageClaim.Spec.Type)
```
nb-ohad, line 277: This was already calculated above as sccType. Please calculate in a single place and reuse

api/v1alpha1/storageclaim_types.go
nb-ohad: @Madhu-1 Please confirm that you are ok with the above so we can merge

Madhu-1: i have not seen anything like this and if others are agreeing to have this i don't have any objection.

controllers/storageclaim_controller.go

leelavg:
/unhold

pulled latest proto change.

Madhu-1 (approved):

PR #148, opened 2024-04-23
controllers: define claimType before getting storageclass

leelavg:
during first reconcile of storageclaim, storageclass will not exist and so `claimType` will have zero value which is being set inside a conditional and fails type check in switch statement as this var was reused.


leelavg:
Tested locally.

Madhu-1 (approved):

PR #151, opened 2024-05-16
Setting basedomain as cluster name

rchikatw:
On hosted cluster value of 
`k get cm cluster-config-v1 -nkube-system -ojsonpath='{.data.install\-config}' | yq -r .metadata.name` is empty.  
Hence, It has been decided to display the base domain regardless of whether it's an HCP or OCP cluster.

### DIFF
--- a/config/rbac/status-reporter-clusterrole.yaml
+++ b/config/rbac/status-reporter-clusterrole.yaml
@@ -22,19 +22,10 @@ rules:
       - list
       - update
   - apiGroups:
-      - ""
-    resources:
-      - configmaps
-    resourceNames:
-      - cluster-config-v1
-    verbs:
-      - get
-      - list
-  - apiGroups:
       - config.openshift.io
     resources:
       - clusterversions
+      - dns
     verbs:
       - get
       - list
-      - watch
--- a/service/status-report/main.go
+++ b/service/status-report/main.go
@@ -18,12 +18,11 @@ package main
 
 import (
 	"context"
-	"github.com/red-hat-storage/ocs-operator/v4/services/provider/interfaces"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/yaml"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/red-hat-storage/ocs-operator/v4/services/provider/interfaces"
 
 	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/csi"
@@ -42,8 +41,7 @@ import (
 
 const (
 	csvPrefix              = "ocs-client-operator"
-	clusterConfigNamespace = "kube-system"
-	clusterConfigName      = "cluster-config-v1"
+	clusterDNSResourceName = "cluster"
 )
 
 func main() {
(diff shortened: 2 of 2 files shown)


raaizik:
@rchikatw I think a more correct approach would be to concatenate the cluster's name and the said domain value. That way we know it's never empty (as per your claim) and actually embeds a cluster's name.

rchikatw:
> @rchikatw I think a more correct approach would be to concatenate the cluster's name and the said domain value. That way we know it's never empty (as per your claim) and actually embeds a cluster's name.

Cluster name is included in the base domain. We don’t need to worry about it

leelavg (commented):
LG, pls add new RBAC and remove old RBAC, do a simple test as well.

service/status-report/main.go
```diff
@@ -156,28 +154,16 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
-	if err != nil {
-		klog.Warningf("Failed to unmarshal cluster metadata, %v", err)
-	}
-	if len(clusterMetadata.Metadata.Name) == 0 {
-		klog.Warningf("Cluster name is empty, %v", err)
+	if len(clusterDNS.Spec.BaseDomain) == 0 {
+		klog.Warningf("Cluster Base Domain is empty.")
 	}
-	status.SetClusterName(clusterMetadata.Metadata.Name)
+	status.SetClusterName(clusterDNS.Spec.BaseDomain)
```
nb-ohad, line 166: @rchikatw does the BaseDomain string includes the cluster's domain name or only the ORG's domain name (the suffix)?

rchikatw: I am not 100% sure about this, the Base domain is the API hostname From what I understood, @leelavg Can you correct if I am wrong.

rchikatw: Checked with cnv team: 
in the DNS CR's `.spec.baseDomain`, it's the cluster name + org domain name
in `install-config.yaml`, it is separated:
`.baseDomain` is only the org's domain name
`.metadata.name` is the cluster name
however, the second option is not available in HCP clusters

config/rbac/status-reporter-clusterrole.yaml
```diff
@@ -22,11 +22,11 @@ rules:
       - list
       - update
   - apiGroups:
```
leelavg, line 24: pls move this rbac to next existing stanza and just add `dns` and also remove `watch` verb on that stanza.

rchikatw: Done

leelavg:
/hold

Can unhold after resolving conversation.

leelavg (approved):

service/status-report/main.go
nb-ohad: Ok, so we are good. I have advocated using the full name anyways. This will enable whoever consumes the name to make their own decision on which part is important to them.

service/status-report/main.go
```diff
@@ -41,9 +40,8 @@ import (
 )
 
 const (
-	csvPrefix              = "ocs-client-operator"
-	clusterConfigNamespace = "kube-system"
-	clusterConfigName      = "cluster-config-v1"
+	csvPrefix      = "ocs-client-operator"
+	clusterDNSName = "cluster"
```
nb-ohad, line 44: If this is supposed to be a placeholder name, please use "N/A" or an empty string. 
And you don't need a constant for it.

nb-ohad (changes requested):

service/status-report/main.go
rchikatw: This is the DNS resource name we need to examine for cluster domain information.

nb-ohad: In that case, please give it a name that makes sense. Maybe `clusterDNSResourceName`. The current name suggest this is a DNS name value.

rchikatw: Done.

leelavg:
@rchikatw could you pls raise a PR cherrypicking commits from #143 & from current PR to `release-4.16` branch?

PR #155, opened 2024-05-21
Bug 2278593: Setting basedomain as cluster name

rchikatw:
Rasing PR for 4.16, for Ref: https://github.com/red-hat-storage/ocs-client-operator/pull/151


rchikatw:
closing this because i cherrypicked the commits in single command that leads to single commit.

PR #161, opened 2024-05-27
bundle: add noobaa-operator dependency

bernerhat:
add the noobaa operator to the list of dependencies as part of [RHSTOR-5187 epic](https://issues.redhat.com/browse/RHSTOR-5187)


leelavg:
We might clash b/n both odf-op & ocs-client-op bringing noobaa-op as a dep, we are facing the same w/ csi-addons, however noobaa could be a bigger piece and need to be coordinate w/ other repos as well.

High level question is, who will update the noobaa dep that we bring?

bernerhat:
> We might clash b/n both odf-op & ocs-client-op bringing noobaa-op as a dep, we are facing the same w/ csi-addons, however noobaa could be a bigger piece and need to be coordinate w/ other repos as well.
I can see the concern but how does it still has an effect while being under a different namespace? in any case i'll consult with the noobaa team to see if they can predict any issues
> High level question is, who will update the noobaa dep that we bring?
i assume our team will need to update that dependency same as with csi-addons

leelavg:
> under a different namespace
- No, odf-op brings ocs-client-op as olm dep which implicitly mean same ns.

ezio-auditore:
This looks good to me. But please hold on this PR from merging until a descision has been made regarding different noobaa operator versions in provider and client clusters.
We might need some more changes in this PR then

leelavg:
> i assume our team will need to update that dependency same as with csi-addons
- yes, I don't see this in the PR, planning for another?

bernerhat:
> > i assume our team will need to update that dependency same as with csi-addons
> 
> * yes, I don't see this in the PR, planning for another?

sorry about that, yes our team will be responsible for updating that dependency. what do you miss in this pr in that regard? @leelavg

ezio-auditore (changes requested):
You can use the util function from k8sutils.go
`func GetClusterID(ctx context.Context, kubeClient client.Client, logger *logr.Logger) string` to fetch the cluster ID instead of writing your own

controllers/operatorconfigmap_controller.go
```diff
@@ -613,6 +623,25 @@ func (c *OperatorConfigMapReconciler) getDeployCSIConfig() (bool, error) {
 	return deployCSI, nil
 }
 
+func (c *OperatorConfigMapReconciler) checkForRemoteClient() (bool,error) {
+	clusterVersion := &configv1.ClusterVersion{}
+	clusterVersion.Name = clusterVersionName
```
ezio-auditore, line 628: From where are you setting the `clusterVersionName` value?

controllers/operatorconfigmap_controller.go
```diff
@@ -613,6 +623,25 @@ func (c *OperatorConfigMapReconciler) getDeployCSIConfig() (bool, error) {
...
+	if err := c.get(clusterVersion); err != nil {
+		c.log.Error(err, "failed to get the clusterVersion version of the OCP cluster")
+		return false, err
+	}
+	storageClients := &v1alpha1.StorageClientList{}
+	if err := c.list(storageClients); err != nil {
+		return false, fmt.Errorf("failed to list storageclients: %v", err)
+	}
+	for idx := range storageClients.Items {
```
ezio-auditore, line 637: You should use the util.Find() to search through the loop.
https://github.com/red-hat-storage/ocs-client-operator/blob/ca0c95c9c552a7d0e1c60e592f036d6dad05546e/pkg/utils/slices.go#L20

controllers/operatorconfigmap_controller.go
ezio-auditore:
```suggestion
	clusterVersion.Name = "version"
```

ezio-auditore: Fetching the clusterID is a useful utility, please add it as a util function

ezio-auditore (changes requested):

controllers/operatorconfigmap_controller.go
bernerhat: not sure what is the added value for that is

leelavg:
@bernerhat if it is tested and working, pls remove the [wip] from the title

@nb-ohad need your opinion, the intention is client-op needs to reconcile noobaa if it is running standalone, this is similar to deploy CSI or not, can we use the same mechanism, like let ocs-op explicitly state client-op not to reconcile noobaa via the existing configmap?

bundle/metadata/dependencies.yaml
```diff
@@ -3,3 +3,7 @@ dependencies:
     value:
       packageName: csi-addons
       version: 0.8.0
+  - type: olm.package
+    value:
+      packageName: noobaa-operator
```
rchikatw, line 8: odf/ocs-operator also brings the noobaa operator. When odf is deployed in provider mode, I think we will have multiple noobaa operators. Will this cause any problem or we are okay with it?

rchikatw: i see this question already being discussed you can resolve this conv.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -613,6 +623,25 @@ func (c *OperatorConfigMapReconciler) getDeployCSIConfig() (bool, error) {
...
+	storageClients := &v1alpha1.StorageClientList{}
+	if err := c.list(storageClients); err != nil {
+		return false, fmt.Errorf("failed to list storageclients: %v", err)
+	}
+	for idx := range storageClients.Items {
+		if storageClients.Items[idx].Status.ConsumerID == string(clusterVersion.Spec.ClusterID){
+			return false, nil
+		}
+	}
+	return true, nil
```
nb-ohad, line 642: This is a bad indicator for running locally. Provider should be the source of truth for this kind of info.
In the case of subscriptions we can have 1 of 2 options:
- Annotation/label the subscription on the ODF side to indicate quasi-ownership, and look for that annotation here.
- More general annotation approach where an annotation indicates that the client op should not touch the subscription regardless of the reason (I like this more than the first one)
- Have some information passed from the provider to the client as part of the onboarding process to indicate that this client should not bother with noobaa concerns. 

Out of all of the options, I like option number 2 the most. It will give us a mechinizem to disable channel changes for a specific sub-operator. And is flexible enough to allow us in the future to remove the annotation and regain ownership

internal/controller/operatorconfigmap_controller.go
```diff
@@ -694,6 +723,20 @@ func (c *OperatorConfigMapReconciler) reconcileClientOperatorSubscription() erro
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() error {
+	noobaaSubscription, err := c.getSubscriptionByPackageName("noobaa-operator")
```
nb-ohad, line 727: Not sure how this calculates the correct subscription when the client is not on the same cluster as the provider. Can you please provide an explanation?

nb-ohad (changes requested):

nb-ohad:
> @bernerhat if it is tested and working, pls remove the [wip] from the title
> 
> @nb-ohad need your opinion, the intention is client-op needs to reconcile noobaa if it is running standalone, this is similar to deploy CSI or not, can we use the same mechanism, like let ocs-op explicitly state client-op not to reconcile noobaa via the existing configmap?

Please see my comment here: https://github.com/red-hat-storage/ocs-client-operator/pull/161#discussion_r1694219671

internal/controller/operatorconfigmap_controller.go
bernerhat: as decided offline, we are going with the approach of a config in the operator's configmap which indicates to the client-operator not to manage noobaa-operator in the case that odf is deployed and the one deploying client-operator and noobaa.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -613,6 +624,21 @@ func (c *OperatorConfigMapReconciler) getDeployCSIConfig() (bool, error) {
...
+	if data != nil {
+		var err error
+		if value, ok := data[manageNoobaaSubKey]; ok {
+			manageNoobaaSub, err = strconv.ParseBool(value)
+			if err != nil {
+				return false, fmt.Errorf("failed to parse value for %q in operator configmap as a boolean: %v", manageNoobaaSubKey, err)
+			}
+		}
+	}
+	return manageNoobaaSub, nil
```
nb-ohad, line 639: Failure of a configuration value should not result in a failure of the reconciliation. Assume that any value that cannot be lowercased to "true", as false. 
This entire function could be replaced with a two-liner on the call site. 
```

manageNoobaaSubAsString := cmp.Or(
    c.operatorConfigMap.Data[manageNoobaaSubKey], 
    strconv.FormatBool(true),
) 
manageNooBaaSub: = strings.ToLower(manageNoobaaSubAsString) != strconv.FormatBool(true)
    
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -694,6 +720,20 @@ func (c *OperatorConfigMapReconciler) reconcileClientOperatorSubscription() erro
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() error {
+	noobaaSubscription, err := c.getSubscriptionByPackageName("noobaa-operator")
+	if err != nil {
+		return err
+	}
+	if c.subscriptionChannel != "" && c.subscriptionChannel != noobaaSubscription.Spec.Channel {
```
nb-ohad, line 728: This code is still pointing to the wrong subscription channel. The noobaa subscription channel is NOT the client subscription channel.

nb-ohad (changes requested):

config/metadata/dependencies.yaml
```diff
@@ -3,3 +3,7 @@ dependencies:
     value:
       packageName: csi-addons
       version: 0.8.0
+  - type: olm.package
+    value:
+      packageName: noobaa-operator
+      version: 5.17.0
```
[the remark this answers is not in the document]
bernerhat: i'm unsure if there needs to be a range here, i believe we don't. but it can be changed, can you weigh in on this? @leelavg

internal/controller/operatorconfigmap_controller.go
bernerhat: @leelavg has assured me in regards of downstream the channel should be the same as client-operator and in regards to upstream we should open a tech-debt issue.

internal/controller/operatorconfigmap_controller.go
bernerhat: i understand what you mean by that but if we are supposed to manage noobaa and fail to do so because of a typo shouldn't we fail the reconciliation? how would we indicate to the user that something is problematic in his configuration? (the user is not creating this configuration as well it is us who set it up so the typo should not happen in any case)

nb-ohad: No, reconciliation failure is the worst result because it stops any and all reconciliation processes, related and unrelated to the noobaa subscription.  
When taking a value from an untyped value (like a config map value) consider that any value is deviation from the default behavior. That means that a type from a valid value should be treated as no change from the default behavior. 
As noobaa subscription reconciliation is the default we need to consider that any value that is not specifically false (ignoring case) means we should reconcile noobaa subscription. 

P.S
I have updated the suggestion in the original comment to check for non-equality with `false`

internal/controller/operatorconfigmap_controller.go
nb-ohad: Any Engineer (even a maintainer) assurance is not enough. We should code defensibly. 
The point is that these 2 channels are conceptually different things and the code should reflect that. 

As I already mentioned offline, the solution is simple. 
An env var to the operator that specifies the desired channel with a panic in case it is not set.  
Then we can set this by default on the operator deployment (as a default value) to have this injected via the downstream build process.

config/metadata/dependencies.yaml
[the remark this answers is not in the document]
leelavg: 1. There is no strict requirement as in d/s it'll be replaced w/ value `> 4.2 <= release-version` when we inform build team.
2. We need to align w/ odf-op in u/s on this as both depend on noobaa.
3. Amit, Ohad mentioned the reason why we need to provide a range, ie, to facilitate upgrades. Pls note that.

If you are installing alongside odf-op, align w/ the range in u/s w/ odf-op for testing both fresh install & upgrades.

internal/controller/operatorconfigmap_controller.go
leelavg: @nb-ohad we agreed on raising an issue and raise a follow-up PR for this, isn't it?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -801,6 +811,14 @@ func (c *OperatorConfigMapReconciler) getDeployCSIConfig() (bool, error) {
 	return deployCSI, nil
 }
 
+func (c *OperatorConfigMapReconciler) getNoobaaSubManagementConfig() bool {
+	manageNoobaaSubAsString := cmp.Or(
+		c.operatorConfigMap.Data[manageNoobaaSubKey],
+		strconv.FormatBool(true),
+	)
+	return strings.ToLower(manageNoobaaSubAsString) != strconv.FormatBool(false)
```
nb-ohad, line 819: 1. There is value in loging an error in case user provided a value in to config map but it does not match an expected value.
2. You can eliminate a lof of the conversions here between string and bools

```suggestion
	valAsString, ok := manageNoobaaSubAsString[manageNoobaaSubKey]
     	if !ok {
        	return true
     	}

	if val, err := strconv.ParseBool(valAsString); err != nil {
		c.Log.Error(
			err, 
			"Unsupported value under manageNoobaaSubscription key", 
		)
		return true
	} else {
		return val
	}
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -882,6 +900,18 @@ func (c *OperatorConfigMapReconciler) reconcileClientOperatorSubscription() erro
...
 
+func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() error {
+	noobaaSubscription, err := c.getSubscriptionByPackageName("noobaa-operator")
+	if err != nil {
+		return err
+	}
+	noobaaSubscription.Spec.Channel = utils.GetNoobaaOperatorChannel()
+	if err := c.update(noobaaSubscription); err != nil {
+		return fmt.Errorf("failed to update subscription channel of 'noobaa-operator' to %v: %v", c.subscriptionChannel, err)
+	}
```
nb-ohad, line 911: This will cause an update on the resource even if the value never changes, which in turn will trigger the reconcile on every controller that watches this resource

nb-ohad (changes requested):

pkg/utils/subscription.go
```diff
@@ -0,0 +1,30 @@
...
+
+package utils
+
+import (
+	"cmp"
+	"os"
+)
+
+var (
+	noobaaOperatorChannel = cmp.Or(os.Getenv("NOOBAA_OPERATOR_CHANNEL"), "latest")
```
nb-ohad, line 25: Does it make sense to have 'latest' as the default? I am not sure there is a channel can even be called latest

internal/controller/operatorconfigmap_controller.go
```diff
@@ -882,6 +907,21 @@ func (c *OperatorConfigMapReconciler) reconcileClientOperatorSubscription() erro
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() error {
+	noobaaSubscription, err := c.getSubscriptionByPackageName("noobaa-operator")
+	if err != nil {
+		return err
+	}
+	desiredSubChannel := utils.GetNoobaaOperatorChannel()
+	if desiredSubChannel != "" && desiredSubChannel != noobaaSubscription.Spec.Channel {
```
nb-ohad, line 916: Looking at the impl of `GetNoobaaOperatorChannel`, the function cannon return "". So why are we testing for that?

nb-ohad (changes requested):

pkg/utils/subscription.go
bernerhat: there is no `latest` channel. it was a suggestion by you previously as a default value indicator for now until we can better address channeling for the dependency.

internal/controller/operatorconfigmap_controller.go
bernerhat: you are correct, changed.

PR #162, opened 2024-05-27
action: update docker-push action for pushing images on demand

leelavg:
relates to red-hat-storage/ocs-operator#2616, when ocs or odf operator is deployed by u/s CI they try to pull a preconfigured bundle of ocs-client-operator for deployment however that bundle is being created on demand manually.

Having a working gh-action for performing that op reduces manual work, at the same time trying to keep the action as short as possible we should set the vars correctly that corresponds to a particular release in Makefile.

Below is an example on how it was done manually
```
export VERSION=4.16.0
export IMG=quay.io/ocs-dev/ocs-client-operator:v${VERSION}
export OCS_CLIENT_CONSOLE_IMG=\
quay.io/ocs-dev/ocs-client-console:v4.15.0
export BUNDLE_IMG=quay.io/ocs-dev/ocs-client-operator-bundle:v${VERSION}
```

Apart from above, client-op depends on csi-addons and for u/s, again we are creating & pushing a bundle for that operator and use the relavent image tags of csi-addons, which will still be manual op atm.

### DIFF
--- a/.github/workflows/docker-push.yaml
+++ b/.github/workflows/docker-push.yaml
@@ -7,8 +7,8 @@ on:
         required: true
         default: ocs-dev
       IMAGE_TAG:
-        required: true
-        default: latest
+        required: false
+        default: ""
 
 jobs:
   docker-push:
@@ -16,8 +16,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
 
     - uses: actions/setup-go@v5
       with:
@@ -34,7 +32,10 @@ jobs:
       env:
         REGISTRY_NAMESPACE: ${{ github.event.inputs.REGISTRY_NAMESPACE }}
         IMAGE_TAG: ${{ github.event.inputs.IMAGE_TAG }}
+        IMAGE_TAG_SHA: ${{ github.ref_name }}-${{ github.sha }}
       run: |
-        make docker-build docker-push
+        # Set IMAGE_TAG to a calculated <branch>-<sha> if not specified
+        export IMAGE_TAG=${IMAGE_TAG:-${IMAGE_TAG_SHA:0:-33}}
+        make container-build container-push
         make bundle-build bundle-push
         make catalog-build catalog-push


leelavg:
@Madhu-1 could you pls check, secrets mentioned in this PR are already defined in the repo? I don't have perms to view that. @iamniting could you pls provide secrets to @Madhu-1? thanks.

iamniting:
> @Madhu-1 could you pls check, secrets mentioned in this PR are already defined in the repo? I don't have perms to view that. @iamniting could you pls provide secrets to @Madhu-1? thanks.

Even I can not see the settings of this repo. @Madhu-1 do let me know if the repo does not have secrets.

.github/workflows/docker-push.yaml
```diff
@@ -34,7 +32,10 @@ jobs:
       env:
         REGISTRY_NAMESPACE: ${{ github.event.inputs.REGISTRY_NAMESPACE }}
         IMAGE_TAG: ${{ github.event.inputs.IMAGE_TAG }}
+        IMAGE_TAG_SHA: ${{ github.ref_name }}-${{ github.sha }}
       run: |
-        make docker-build docker-push
+        # Set IMAGE_TAG to a calculated <branch>-<sha> if not specified
+        export IMAGE_TAG=${IMAGE_TAG:-${IMAGE_TAG_SHA:0:-33}}
```
Madhu-1, line 38: if the TAG is not provided can we push with branch name or use the latest for main branch and branch name for the released branch, this ensures that we don't push to many images in long term

iamniting: @Madhu-1 This action will not push images automatically, It will only push if someone invokes it. So we should be good.

Madhu-1: i still like to change that as we are making the TAG as optional in this PR.

Madhu-1:
> > @Madhu-1 could you pls check, secrets mentioned in this PR are already defined in the repo? I don't have perms to view that. @iamniting could you pls provide secrets to @Madhu-1? thanks.
> 
> Even I can not see the settings of this repo. @Madhu-1 do let me know if the repo does not have secrets.

strangely I don't have all/admin access for this repo. @nb-ohad @jarrpa can help you.

.github/workflows/docker-push.yaml
nb-ohad: @Madhu-1 What exactly are you suggesting, if we put a specific tag it will become mandatory

Madhu-1: instead of Hash/SHA i would vote to tag as on the branch name to avoid having many tags in the registry.

iamniting: @Madhu-1 We need unique images for the odf and ocs upstream ci, If we push the images with same tag we may see a lot of failures in the odf and ocs CI.

Madhu-1: can you please explain it a bit why we need to have unique names?

iamniting: I am not talking about the unique names, I am talking about the unique tags.

client-operator and ocs-operator are tightly coupled with each other. We should stick to one image unless the required changes have merged in both the repos together. Moreover, why are we worried about having a lot of images? That is how the whole industry is doing. for example rook.

Madhu-1: >why are we worried about having a lot of images? 

This will a concern about maintenance and cleanup why we need to have lot of images without a need is still the question.

sorry that was a typo, i meant unique tag. i don't see a reason to take example from the whole industry or Rook and try to impose it here without a proper reason. Even if they are tight-coupled, could you please explain why the tags need to be unique here?

iamniting: Lets say something has merged in the ocs-client-operator but we have not made the required changes to the ocs-operator. In this case ocs-operator CI will pick the latest changes from the ocs-client-operator and will fail the CI.

It will put us at risk and all the PR's will be blocked due to the failing CI. Moreover, it will take a lot of cycles to dig into the CI and look at failures.

leelavg: PR is still required, any resolution can be reached?

iamniting: Yes we do need a PR or someone has to manually create the images everytime

Madhu-1: okay, merging this one as its a manual workflow trigger not the automated one, but now its the admin/maintainer responsibility to ensure they set the required details and avoid pushing more images to the quay

PR #168, opened 2024-06-03
Generates and labels `VolumeGroupSnapshotClass`

raaizik:
## Changes

- Have storage claim controller own `VolumeGroupSnapshotClass` with the required labels for RDR added to both VGSC and VSC. 
- Adds Available CRDs functionality to k8sutils
  - Resolves [compatibility issue](https://github.com/red-hat-storage/ocs-client-operator/pull/168#pullrequestreview-2115248107)
- Upgrades [k8s-csi ](https://github.com/red-hat-storage/ocs-client-operator/pull/168#discussion_r1637880958)
  - Requires go >= 1.22.0 which uses client-go v0.30 which [works](https://github.com/kubernetes-sigs/controller-runtime/issues/2788#issuecomment-2063935394) with CR v0.18 [and](https://github.com/kubernetes-sigs/controller-tools/issues/880) CG v0.14

[RHSTOR-5795](https://issues.redhat.com/browse/RHSTOR-5795)

## TODO

- [x] Change to v1beta1 once available for `VolumeGroupSnapshotClass` [#1150](https://github.com/kubernetes-csi/external-snapshotter/pull/1150)
- [ ] Test complete workflow


controllers/storageclaim_controller.go
```diff
@@ -651,6 +677,68 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
...
+
+	if err := r.get(existing); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to get VolumeGroupSnapshotClass: %v", err)
+	}
+
+	// If present then compare the existing VolumeGroupSnapshotClass parameters with
+	// the received VolumeGroupSnapshotClass parameters, and only proceed if they differ.
+	if reflect.DeepEqual(existing.Parameters, volumeGroupSnapshotClass.Parameters) {
+		return nil
+	}
```
rewantsoni, line 722: Should we also check for labels?

rewantsoni (changes requested):

controllers/storageclaim_controller.go
```diff
@@ -398,15 +404,33 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
 				utils.AddAnnotation(volumeSnapshotClass, storageClaimAnnotation, r.storageClaim.Name)
 				if err := r.createOrReplaceVolumeSnapshotClass(volumeSnapshotClass); err != nil {
 					return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeSnapshotClass: %s", err)
 				}
+			case "VolumeGroupSnapshotClass":
+				var volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass
+				data["csi.storage.k8s.io/snapshotter-secret-namespace"] = r.OperatorNamespace
+				// generate a new clusterID for cephfs subvolumegroup, as
+				// storageclaim is clusterscoped resources using its
+				// hash as the clusterID
```
leelavg, line 422: infact you could delete the existing comment rather than copying it.

controllers/storageclaim_controller.go
```diff
@@ -398,15 +404,33 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 				// storageclaim is clusterscoped resources using its
 				// hash as the clusterID
 				data["clusterID"] = r.storageClaimHash
+				labels["ramendr.openshift.io/storageID"] = storageID
 				if resource.Name == "cephfs" {
-					volumeSnapshotClass = r.getCephFSVolumeSnapshotClass(data)
+					volumeSnapshotClass = r.getCephFSVolumeSnapshotClass(data, labels)
```
leelavg, line 409: data is specific and labels are generic, no need to club them in here as I believe we already have a util which adds labels to objects, pls use that.

controllers/storageclaim_controller.go
```diff
@@ -327,11 +330,14 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
 		for _, resource := range resources {
 			data := map[string]string{}
+			labels := map[string]string{}
 			err = json.Unmarshal(resource.Data, &data)
 			if err != nil {
 				return reconcile.Result{}, fmt.Errorf("failed to unmarshal StorageClaim configuration response: %v", err)
 			}
 
+			// SID for RamenDR
+			storageID := getMD5Hash(string(r.storageClaim.UID)) + r.storageClaimHash
```
leelavg, line 340: May I know who decided this? Because UID is already unique and not sure why we require hash(uid) + hash.

leelavg (commented):
Content wise this PR doesn't depend on PR at ocs-operator end right?

controllers/storageclaim_controller.go
raaizik: Data and Labels are assigned to completely different fields in the CR, therefore they must be sent over as separate arguments.

controllers/storageclaim_controller.go
raaizik: I did. The point is to have distinct and consistent across clusters identifiers. Replication ID will store `r.storageClaimHash` and SID must differ from it and be unique, hence why I concatenated in this way.

raaizik:
> Content wise this PR doesn't depend on PR at ocs-operator end right?

See [#2661](https://github.com/red-hat-storage/ocs-operator/pull/2661)

controllers/storageclaim_controller.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 
 	"github.com/go-logr/logr"
+	groupsnapapi "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumegroupsnapshot/v1alpha1"
```
Madhu-1, line 35: use `v8` client

controllers/storageclaim_controller.go
```diff
@@ -398,15 +404,33 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
-					volumeSnapshotClass = r.getCephRBDVolumeSnapshotClass(data)
+					volumeSnapshotClass = r.getCephRBDVolumeSnapshotClass(data, labels)
 				}
 				utils.AddAnnotation(volumeSnapshotClass, storageClaimAnnotation, r.storageClaim.Name)
 				if err := r.createOrReplaceVolumeSnapshotClass(volumeSnapshotClass); err != nil {
 					return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeSnapshotClass: %s", err)
 				}
+			case "VolumeGroupSnapshotClass":
+				var volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass
+				data["csi.storage.k8s.io/snapshotter-secret-namespace"] = r.OperatorNamespace
```
Madhu-1, line 419: This secret is not correct, can you please check [the](https://github.com/kubernetes-csi/external-snapshotter/blob/d505406912a47db92f433b07de03c7c77bc76318/pkg/utils/util.go#L62-L63)

controllers/storageclaim_controller.go
```diff
@@ -599,9 +625,10 @@ func (r *StorageClaimReconciler) createOrReplaceVolumeSnapshotClass(volumeSnapsh
 		return fmt.Errorf("failed to get VolumeSnapshotClass: %v", err)
 	}
 
-	// If present then compare the existing VolumeSnapshotClass parameters with
-	// the received VolumeSnapshotClass parameters, and only proceed if they differ.
-	if reflect.DeepEqual(existing.Parameters, volumeSnapshotClass.Parameters) {
+	// If present then compare the existing VolumeSnapshotClass parameters and labels with
+	// the received VolumeSnapshotClass parameters and labels, and only proceed if they differ.
+	if reflect.DeepEqual(existing.Parameters, volumeSnapshotClass.Parameters) && reflect.DeepEqual(existing.Labels,
```
Madhu-1, line 630: if the labels are not matching we should not delete the resource and recreate it we just need to update it, can you rework this logic?

controllers/storageclaim_controller.go
```diff
@@ -651,6 +678,69 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
...
+	if err := r.get(existing); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to get VolumeGroupSnapshotClass: %v", err)
+	}
+
+	// If present then compare the existing VolumeGroupSnapshotClass parameters and labels with
+	// the received VolumeGroupSnapshotClass parameters and labels, and only proceed if they differ.
+	if reflect.DeepEqual(existing.Parameters, volumeGroupSnapshotClass.Parameters) && reflect.DeepEqual(existing.Labels,
+		volumeGroupSnapshotClass.Labels) {
+		return nil
+	}
```
Madhu-1, line 724: same here as well.

controllers/storageclaim_controller.go
```diff
@@ -651,6 +678,69 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   r.storageClaim.Name,
+			Labels: labels,
+		},
+		Driver:         csi.GetRBDDriverName(),
+		DeletionPolicy: snapapi.VolumeSnapshotContentDelete,
+		Parameters:     data,
+	}
+	return volumegroupsnapshotclass
+}
```
Madhu-1, line 705: can we rework on this and existing code for separate RBD and cephfs by just passing the driver name so that we can get rid of duplicate functions?

Madhu-1 (changes requested):
@raaizik As ocs-client-operator need to runs on different OCP versions where we don't have volumegroupsnapshot support how we are going to handle this case?

lets keep this PR on hold until the voluemgroupsnapshot goes to beta or else whoever is testing ocs-client-operator need to do extra steps in OCP to have volumegroupsnapshot support and we need to have one more PR to change v1alpav1 to v1beta1.

controllers/storageclaim_controller.go
leelavg: Labels always go into object meta on any resource.

raaizik: And? This doesn't contradict what I wrote.

controllers/storageclaim_controller.go
raaizik: It requires go  >= 1.22.0. Should I go ahead with the upgrade?

Makefile
raaizik, line 1: `yq` isn't recognized anymore.

controllers/storageclaim_controller.go
raaizik: Why? It is still relevant.

controllers/storageclaim_controller.go
nb-ohad: Storage ID does not need to be distinct across clusters, only the Replication ID needs to be.
But even if the requirement was true for Storage ID, a UUID is assumed to be globally unique (by definition).

```suggestion
			storageID := r.storageClaim.UID
```

And please move the definition of the storage ID var outside of the for, there is no need to redefine it on every iteration as it is constant for the entire reconciliation.

controllers/storageclaim_controller.go
```diff
@@ -398,15 +404,27 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
 				// hash as the clusterID
 				data["clusterID"] = r.storageClaimHash
-				if resource.Name == "cephfs" {
-					volumeSnapshotClass = r.getCephFSVolumeSnapshotClass(data)
-				} else if resource.Name == "ceph-rbd" {
-					volumeSnapshotClass = r.getCephRBDVolumeSnapshotClass(data)
-				}
+				labels["ramendr.openshift.io/storageID"] = storageID
+				volumeSnapshotClass = r.getCephDriverVolumeSnapshotClass(resource.Name, data)
+				utils.AddLabels(volumeSnapshotClass, labels)
```
nb-ohad, line 409: If we are only concerned with a single-label why do we need the labels map var in the first place
```suggestion
				volumeSnapshotClass = r.getCephDriverVolumeSnapshotClass(resource.Name, data)
				utils.AddLabel(volumeSnapshotClass, "ramendr.openshift.io/storageID", storageID)

File: controllers/storageclaim_controller.go [hunk 11]
```diff
@@ -398,15 +404,27 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
+			case "VolumeGroupSnapshotClass":
+				var volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass
+				data["csi.storage.k8s.io/group-snapshotter-secret-namespace"] = r.OperatorNamespace
+				// generate a new clusterID for cephfs subvolumegroup, as
+				// storageclaim is clusterscoped resources using its
+				// hash as the clusterID
+				data["clusterID"] = r.storageClaimHash
+				labels["ramendr.openshift.io/storageID"] = storageID
+				volumeGroupSnapshotClass = r.getCephDriverVolumeGroupSnapshotClass(resource.Name, data)
+				utils.AddLabels(volumeGroupSnapshotClass, labels)
```
nb-ohad, line 423: Same comment as the one above regarding `AddLabels`

controllers/storageclaim_controller.go
```diff
@@ -398,15 +404,27 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
 				if err := r.createOrReplaceVolumeSnapshotClass(volumeSnapshotClass); err != nil {
 					return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeSnapshotClass: %s", err)
 				}
+			case "VolumeGroupSnapshotClass":
+				var volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass
+				data["csi.storage.k8s.io/group-snapshotter-secret-namespace"] = r.OperatorNamespace
+				// generate a new clusterID for cephfs subvolumegroup, as
+				// storageclaim is clusterscoped resources using its
+				// hash as the clusterID
+				data["clusterID"] = r.storageClaimHash
```
nb-ohad, line 420: @raaizik I am confused.
 If we can decide on a cluster ID locally why do we need to send it from the provider (as implemented in this PR: https://github.com/red-hat-storage/ocs-operator/pull/2661)?

controllers/storageclaim_controller.go
```diff
@@ -599,27 +612,36 @@ func (r *StorageClaimReconciler) createOrReplaceVolumeSnapshotClass(volumeSnapsh
 		return fmt.Errorf("failed to get VolumeSnapshotClass: %v", err)
 	}
 
-	// If present then compare the existing VolumeSnapshotClass parameters with
-	// the received VolumeSnapshotClass parameters, and only proceed if they differ.
-	if reflect.DeepEqual(existing.Parameters, volumeSnapshotClass.Parameters) {
+	// If present then compare the existing VolumeSnapshotClass parameters and labels with
+	// the received VolumeSnapshotClass parameters and labels, and only proceed if they differ.
+	labelsEqual := reflect.DeepEqual(existing.Labels, volumeSnapshotClass.Labels)
+	if reflect.DeepEqual(existing.Parameters, volumeSnapshotClass.Parameters) && labelsEqual {
```
nb-ohad, line 618: Let's make the code more readable by aligning both the params and label equality checks
Also, we are setting annotations so we should also consider checking equality there as well

``suggestion
	areLabelsEqual := eflect.DeepEqual(existing.Labels, volumeSnapshotClass.Labels)
	areAnnotationsEqual := eflect.DeepEqual(existing.Annotations, volumeSnapshotClass.Annotations)
	areParamsEqual := reflect.DeepEqual(existing.Parameters, volumeSnapshotClass.Parameters)
	if areLabelsEqual && areAnnotationsEqual && areParamsEqual {

controllers/storageclaim_controller.go
```diff
@@ -599,27 +612,36 @@ func (r *StorageClaimReconciler) createOrReplaceVolumeSnapshotClass(volumeSnapsh
...
 		return nil
 	}
 
-	// VolumeSnapshotClass already exists, but parameters have changed. Delete the existing VolumeSnapshotClass and create a new one.
-	if existing.UID != "" {
-		// Since we have to update the existing VolumeSnapshotClass, so we will delete the existing VolumeSnapshotClass and create a new one.
-		r.log.Info("VolumeSnapshotClass needs to be updated, deleting it.", "Name", existing.Name)
+	if labelsEqual {
```
nb-ohad, line 622: This check does not make sense. 
At this point in the code, it is already established that an update is needed (because of label changes or params changes)

The only concern left is if we need to delete the existing one or not!

Your current approach has a couple of bugs:
1. If the VolumeSnapshotClass didn't exist then your labelsEqual will be true and you will jump to the lease which will try to update instead of create
2. if both labels and params changed you will jump to the else and try to update instead of deleting and creating 

With that in mind, I would prefer to keep the code simple and always delete and recreate, even in cases where we don't have to delete. Trying to do it differently will just lead to many edge cases that we can easily miss and the overhead in deleting and recreating is negligible as we don't assume many label changes

controllers/storageclaim_controller.go
```diff
@@ -651,6 +673,70 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
 	return false, nil
 }
 
+func (r *StorageClaimReconciler) getCephDriverVolumeGroupSnapshotClass(driverType string,
```
nb-ohad, line 676: Can we please move this function near the `getCephDriverVolumeSnapshotClass` function? both of them have similar concerns and I would prefer to have both of them at a similar location in the file

controllers/storageclaim_controller.go
```diff
@@ -651,6 +673,70 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
...
+	return volumegroupsnapshotclass
+}
+
+func (r *StorageClaimReconciler) createOrReplaceVolumeGroupSnapshotClass(volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass) error {
+	existing := &groupsnapapi.VolumeGroupSnapshotClass{}
+	existing.Name = r.storageClaim.Name
+
+	if err := r.own(volumeGroupSnapshotClass); err != nil {
+		return fmt.Errorf("failed to own volumegroupsnapshotclass: %v", err)
+	}
```
nb-ohad, line 701: Isn't volumeGroupSnapshotClass (and volumeSnapshotClass) a cluster-wide resource? In that case, you cannot own it by a namespace scoped resource (StorageClaim resource). We are already working around that using the storage claim annotation. So let's remove this `own` call.

controllers/storageclaim_controller.go
```diff
@@ -651,6 +673,70 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
...
+		}
+	} else {
+		r.log.Info("Updating VolumeGroupSnapshotClass", "Name", existing.Name)
+		if err := r.Client.Update(r.ctx, volumeGroupSnapshotClass); err != nil {
+			return fmt.Errorf("failed to update VolumeGroupSnapshotClass: %v", err)
+		}
+	}
+
+	return nil
+}
```
nb-ohad, line 738: This function is almost an identical duplication of the `createOrReplaceVolumeSnapshotClass`. 
Instead of creating a duplicate where Bugz might be repeating, can we please generalize to a 
func (r *StorageClaimReconciler) createOrReplace(client object.Client) error

nb-ohad (changes requested):

controllers/storageclaim_controller.go
raaizik: As are we ➡️ [discussion](https://github.com/red-hat-storage/ocs-operator/pull/2661#discussion_r1642481467)

controllers/storageclaim_controller.go
raaizik: Looks like your opinions are conflicting ➡️ [discussion](https://github.com/red-hat-storage/ocs-client-operator/pull/168#discussion_r1637891065)

controllers/storageclaim_controller.go
raaizik: For now. Who knows what will be added next

controllers/storageclaim_controller.go
raaizik: @rewantsoni WDYT?

rewantsoni: Yes we can use storageClaim UID as the storageID

controllers/storageclaim_controller.go
Madhu-1: @nb-ohad i dont agree to delete the resource when its not required, we do have update for that case, these resources have special meaning we should use what is required when instead of having a common case to delete and recreate always if there is a change but it doesn't require to be deleted.

nb-ohad: @Madhu-1 This is going to be an evergrowing pit of bugs, even the current implementation is missing very important edge cases. I am not predicting too many label, and annotation changes over the lifetime of these resources and want to avoid very common eng mistakes that result in non-so-simple bugs when an Eng touches this code.

Madhu-1: I believe it's pretty straightforward because we already know what will change and what will not change. However, I haven't reviewed this part of the code thoroughly, so I may be missing some edge cases. My suggestion is to update the metadata if it changes, and to use delete/create for anything else. If you believe there are potential issues, we can resort to using delete/create, but I still prefer updating in certain cases.

raaizik: Changed back to no update.

controllers/storageclaim_controller.go
raaizik: Changed

controllers/storageclaim_controller.go
```diff
@@ -207,6 +219,20 @@ func (r *StorageClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return result, nil
 }
 
+func (r *StorageClaimReconciler) mapCRDAvailability(crdNames ...string) map[string]bool {
+	CRDExist := map[string]bool{}
```
Madhu-1, line 223: use `crdExists`

controllers/storageclaim_controller.go
```diff
@@ -70,6 +75,7 @@ type StorageClaimReconciler struct {
 	storageClient    *v1alpha1.StorageClient
 	storageClaim     *v1alpha1.StorageClaim
 	storageClaimHash string
+	AvailableCRDs    map[string]bool
```
Madhu-1, line 78: use `availableCRDs`, lets not expose the member if its not required.

controllers/storageclaim_controller.go
```diff
@@ -501,28 +544,42 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
 	return storageClass
 }
 
-func (r *StorageClaimReconciler) getCephFSVolumeSnapshotClass(data map[string]string) *snapapi.VolumeSnapshotClass {
+func (r *StorageClaimReconciler) getCephDriverVolumeSnapshotClass(driverType string,
+	data map[string]string) *snapapi.VolumeSnapshotClass {
+	var driverName string
+	if driverType == "cephfs" {
+		driverName = csi.GetCephFSDriverName()
+	} else {
```
Madhu-1, line 552: can we use `} else if driverType == "rbd" ` ? we might get nfs later

controllers/storageclaim_controller.go
```diff
@@ -563,6 +620,63 @@ func (r *StorageClaimReconciler) createOrReplaceStorageClass(storageClass *stora
 	return nil
 }
 
+func (r *StorageClaimReconciler) createOrReplaceClass(class client.Object) error {
```
Madhu-1, line 623: can we update this to cover storageclass as well?

controllers/storageclaim_controller.go
```diff
@@ -207,6 +219,20 @@ func (r *StorageClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return result, nil
 }
 
+func (r *StorageClaimReconciler) mapCRDAvailability(crdNames ...string) map[string]bool {
```
Madhu-1, line 222: Do we need to make crdNames as input? instead of that can we have it as var inside this method only?

controllers/storageclaim_controller.go
```diff
@@ -501,64 +544,111 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
...
 }
 
-func (r *StorageClaimReconciler) getCephRBDVolumeSnapshotClass(data map[string]string) *snapapi.VolumeSnapshotClass {
-	volumesnapshotclass := &snapapi.VolumeSnapshotClass{
+func (r *StorageClaimReconciler) getCephDriverVolumeGroupSnapshotClass(driverType string,
+	data map[string]string) *groupsnapapi.VolumeGroupSnapshotClass {
+	var driverName string
+	if driverType == "cephfs" {
+		driverName = csi.GetCephFSDriverName()
+	} else {
```
Madhu-1, line 572: can you please add the case of rbd here?

controllers/storageclaim_controller.go
```diff
@@ -501,64 +544,111 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
...
+			}.String())
+			return fmt.Errorf("failed to delete "+classKind+": %v", err)
 		}
 	}
-	r.log.Info("Creating StorageClass.", "StorageClass", klog.KRef(storageClass.Namespace, existing.Name))
-	err := r.Client.Create(r.ctx, storageClass)
-	if err != nil {
-		return fmt.Errorf("failed to create StorageClass: %v", err)
+	r.log.Info("Creating "+classKind+".", "Name", types.NamespacedName{
+		Namespace: class.GetNamespace(),
```
Madhu-1, line 647: namespace is not required as its clusterScoped resouces

Madhu-1 (commented):
some nits, LGTM

controllers/storageclaim_controller.go
raaizik: What do you mean? We need to be able to send to this function all the CRD names that need to go through this flow path.

controllers/storageclaim_controller.go
raaizik: And how about for `StoageClass`? Seems like it did have its namespace mentioned in the logs:
```go
r.log.Info("StorageClass needs to be updated, deleting it.", "StorageClass", klog.KRef(storageClass.Namespace, existing.Name))
```

controllers/storageclaim_controller.go
```diff
@@ -501,64 +543,103 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
...
 		Parameters:     data,
 	}
-	return volumesnapshotclass
+	return volumegroupsnapshotclass
 }
 
-func (r *StorageClaimReconciler) createOrReplaceStorageClass(storageClass *storagev1.StorageClass) error {
-	existing := &storagev1.StorageClass{}
-	existing.Name = r.storageClaim.Name
+func (r *StorageClaimReconciler) createOrReplaceClass(class client.Object) error {
```
rewantsoni, line 586: Shouldn't we get the existing before the switch case? existingParameters will always be empty in this case. 
We should, 
1. set the name, to existing, get it based on the object kind
2. compare the existing parameter/labels/annotations to the desired
3. if equal break
4. own the desired
5. if UID is not nil delete it
6. create the desired

Could we also rename class client.Object to a better name? maybe to desired?

rewantsoni (changes requested):

Madhu-1 (changes requested):
@raaizik i totally forgot about one flag and RBAC https://github.com/kubernetes-csi/external-snapshotter/blob/c72f087f751abb285a90c4f5bb02df9014d2bc19/cmd/csi-snapshotter/main.go#L88 this need to be set to true and we need to have required RBAC's for csi driver as well to make use of this whole feature? Have you tested the complete end to end workflow?

controllers/storageclaim_controller.go
Madhu-1: it looks like that is also not required but it was present

controllers/storageclaim_controller.go
```diff
@@ -207,6 +218,20 @@ func (r *StorageClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return result, nil
 }
 
+func (r *StorageClaimReconciler) mapCRDAvailability(crdNames ...string) map[string]bool {
```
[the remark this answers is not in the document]
raaizik: Done

controllers/storageclaim_controller.go
raaizik: OK now that I converted it to a util function it might make more sense to you

controllers/storageclaim_controller.go
```diff
@@ -398,15 +411,28 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
+					var volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass
+					data["csi.storage.k8s.io/group-snapshotter-secret-namespace"] = r.OperatorNamespace
+					// generate a new clusterID for cephfs subvolumegroup, as
+					// storageclaim is clusterscoped resources using its
+					// hash as the clusterID
+					data["clusterID"] = r.storageClaimHash
+					volumeGroupSnapshotClass = r.getCephDriverVolumeGroupSnapshotClass(resource.Name, data)
+					utils.AddLabel(volumeGroupSnapshotClass, storageIDRamenLabel, storageID)
+					utils.AddAnnotation(volumeGroupSnapshotClass, storageClaimAnnotation, r.storageClaim.Name)
+					if err := r.createOrReplaceClass(volumeGroupSnapshotClass); err != nil {
```
Madhu-1, line 432: when the storageclaim gets deleted we need to ensure that no volumegroupsnapshotcontent exists before deleting the storageclaim

pkg/templates/csisidecars.go
```diff
@@ -91,6 +91,7 @@ var SnapshotterContainer = corev1.Container{
 		"--timeout=150s",
 		"--leader-election=true",
 		"--extra-create-metadata=true",
+		"--enable-volume-group-snapshots=true",
```
Madhu-1, line 94: This flag doesnt exists on all snapshotter sidecar versions. this exists in only ODF 4.16 and ODF 4.17 not below that and also if the CRD's doesn't exists we should not set this to true, because the snapshotter controller will crash.

config/rbac/csi_rbd_provisioner_clusterrole.yaml
```diff
@@ -45,6 +45,9 @@ rules:
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshotcontents/status"]
     verbs: ["update", "patch"]
+  - apiGroups: ["groupsnapshot.storage.k8s.io"]
+    resources: ["volumegroupsnapshotclasses"]
+    verbs: ["create", "delete", "get", "list", "watch"]
```
Madhu-1, line 50: Need to update RBAC for both cephfs and rbd like we did in this PR https://github.com/rook/rook/pull/13832/files. we don't need create/delete access for cephcsi

Madhu-1 (commented):
@raaizik Can you please test it out and see if works as expected.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/red-hat-storage/ocs-client-operator
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.4
```
nb-ohad, line 5: What is the reason we need to add toolchain to our go.mod file?

main.go
```diff
@@ -144,11 +144,12 @@ func main() {
 	setupLog.Info("setting up webhook server")
 	hookServer := mgr.GetWebhookServer()
 
+	decoder := admission.NewDecoder(mgr.GetScheme())
 	setupLog.Info("registering Subscription Channel validating webhook endpoint")
 	hookServer.Register("/validate-subscription", &webhook.Admission{
 		Handler: &admwebhook.SubscriptionAdmission{
 			Client:  mgr.GetClient(),
-			Decoder: admission.NewDecoder(mgr.GetScheme()),
+			Decoder: &decoder,
```
nb-ohad, line 152: Why this change? what was wrong with the old code? Even after the change the decoder is used only once

pkg/utils/k8sutils.go
```diff
@@ -101,3 +106,17 @@ func AddAnnotation(obj metav1.Object, key string, value string) bool {
 	}
 	return false
 }
+
+func MapCRDAvailability(ctx context.Context, clnt client.Client, log logr.Logger, crdNames ...string) map[string]bool {
+	crdExist := map[string]bool{}
+	for _, crdName := range crdNames {
+		crd := &apiextensions.CustomResourceDefinition{}
+		crd.Name = crdName
+		if err := clnt.Get(ctx, client.ObjectKeyFromObject(crd), crd); err != nil && !errors.IsNotFound(err) {
```
nb-ohad, line 115:
```suggestion
		if err := clnt.Get(ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
```

pkg/utils/k8sutils.go
```diff
@@ -101,3 +106,17 @@ func AddAnnotation(obj metav1.Object, key string, value string) bool {
...
+			log.Error(err, fmt.Sprintf("Error getting CRD  for %s", crdName))
+			os.Exit(1)
```
nb-ohad, line 117: There should be no reason to exit the process from a util func. 
Return an error instead and exit in the call site if needed.

pkg/webhook/subscription.go
```diff
@@ -24,7 +24,7 @@ func (s *SubscriptionAdmission) Handle(ctx context.Context, req admission.Reques
 
 	// review should be for a subscription
 	subscription := &opv1a1.Subscription{}
-	if err := s.Decoder.Decode(req, subscription); err != nil {
+	if err := (*s.Decoder).Decode(req, subscription); err != nil {
```
nb-ohad, line 27: Why dereference to get a copy? If `s.Decoder` is a pointer then `s.Decode.Decode` should work

nb-ohad (changes requested):

go.mod
raaizik: The toolchain directive gets automatically added upon running go commands (e.g., `go mod tidy`) for versions starting in 1.21.

main.go
raaizik: In [CR v0.18](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission@v0.18.0#NewDecoder), `NewDecoder` returns argument of type `Decoder`, while what is required here is what used to be returned from [CR v0.17](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission@v0.17.2#NewDecoder) -- a pointer to one.

pkg/webhook/subscription.go
raaizik: In [CR v0.18](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission@v0.18.0#Decoder), `Decoder` is an interface and not a struct as it was in [CR v0.17.2](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/webhook/admission@v0.17.2#Decoder) -- this basically renders `s.Decoder.Decode` undefined.

go.mod
nb-ohad: ok, thanks for the explanation

pkg/webhook/subscription.go
nb-ohad: This means something else is wrong. An interface does not replace an implementation it just captures a contract and described on the link you attached above, that contract includes a Decode function. This means that you should be able to call decode on the Decoder interface which will invoke the decode function on the actual instance. 

If `s.Decoder.Decode` returns a different behavior than `(*s.Decoder).Decode` it is a very strong indication that something else is wrong here, which we need to resolve before allowing this to merge, or at least until we fully understand why it behaves this way

main.go
nb-ohad: A pointer to an interface does not make sense in almost all cases. 
When moving to an interface usually all places accepting a pointer to a struct will now accept an interface instead. 
I have a feeling that we have updated some libs but not all of the necessary libs, as part of the framework upgrade. 

Have we followed the upgrade notes? Sometimes there are manual changes that need to be applied

pkg/webhook/subscription.go
leelavg: seems there is a breaking change as per https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.18.0

nb-ohad: A breaking change is ok as long as we are following the migration guide. I don't believe the migration guide wants us to use a pointer to an interface, but I might be wrong

main.go
raaizik: Rebase allowed reverting this change

pkg/webhook/subscription.go
raaizik: Rebase allowed reverting this change

controllers/storageclaim_controller.go
```diff
@@ -501,64 +539,111 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
...
-	// StorageClass already exists, but parameters have changed. Delete the existing StorageClass and create a new one.
-	if existing.UID != "" {
+	// If present then compare the existing Storage(Volume*Snapshot)Class parameters and labels with
+	// the received Storage(Volume*Snapshot)Class parameters and labels, and only proceed if they differ.
+	labelsEqual := reflect.DeepEqual(existing.GetLabels(), class.GetLabels())
+	parametersEqual := reflect.DeepEqual(existingParameters, classParameters)
+	annotationsEqual := reflect.DeepEqual(existing.GetAnnotations(), class.GetAnnotations())
+	if labelsEqual && parametersEqual && annotationsEqual {
+		return nil
+	}
```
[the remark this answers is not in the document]
raaizik: You're describing the implementation of `CreateOrUpdate`, yet what is required is to Recreate (delete and create) in case of inequality. Also, it's not exactly an ever-growing list per se as these would be the only sections of the CR that need comparing.

controllers/storageclaim_controller.go
```diff
@@ -53,9 +54,13 @@ import (
 const (
 	storageClaimFinalizer  = "storageclaim.ocs.openshift.io"
 	storageClaimAnnotation = "ocs.openshift.io/storageclaim"
+	storageIDRamenLabel    = "ramendr.openshift.io/storageID"
 
-	pvClusterIDIndexName  = "index:persistentVolumeClusterID"
-	vscClusterIDIndexName = "index:volumeSnapshotContentCSIDriver"
+	pvClusterIDIndexName         = "index:persistentVolumeClusterID"
+	vscClusterIDIndexName        = "index:volumeSnapshotContentCSIDriver"
+	volumeGroupSnapshotClassCRD  = "groupsnapshot.storage.k8s.io/volumegroupsnapshotclass"
```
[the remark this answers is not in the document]
raaizik: Linting rules don't permit `Id`, it demands `ID`.

controllers/storageclaim_controller.go
```diff
@@ -106,11 +112,16 @@ func (r *StorageClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
-	return ctrl.NewControllerManagedBy(mgr).
+	r.availableCRDs = utils.MapCRDAvailability(r.ctx, r.Client, r.log, volumeGroupSnapshotClassCRD)
+
+	storageClaimController := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.StorageClaim{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&storagev1.StorageClass{}).
-		Owns(&snapapi.VolumeSnapshotClass{}).
-		Complete(r)
+		Owns(&snapapi.VolumeSnapshotClass{})
```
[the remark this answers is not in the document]
raaizik: That's blocked by [#2745](https://github.com/red-hat-storage/ocs-operator/pull/2745)

internal/controller/operatorconfigmap_controller.go
```diff
@@ -323,6 +324,15 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
...
 
+	crd := &metav1.PartialObjectMetadata{}
+	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+	crd.Name = volumeGroupSnapshotClassCrd
+	if err := c.Client.Get(c.ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
+		c.log.Error(err, "Failed to get CRD", "CRD", volumeGroupSnapshotClassCrd)
+		return err
+	}
+	utils.AssertEqual(c.AvailCrds[volumeGroupSnapshotClassCrd], crd.UID != "", utils.ExitCodeThatShouldRestartTheProcess)
+
```
rewantsoni, line 335: This should move to storageClaim controller, where we are having a similar check for drClusterconfigs

internal/controller/storageclaim_controller.go
```diff
@@ -56,11 +58,12 @@ const (
 	storageClaimFinalizer  = "storageclaim.ocs.openshift.io"
 	storageClaimAnnotation = "ocs.openshift.io/storageclaim"
 	keyRotationAnnotation  = "keyrotation.csiaddons.openshift.io/schedule"
+	storageIDRamenLabel    = "ramendr.openshift.io/storageID"
```
rewantsoni, line 61: We don't need this anymore right?

internal/controller/storageclaim_controller.go
```diff
@@ -443,15 +461,27 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
 				utils.AddAnnotation(volumeSnapshotClass, storageClaimAnnotation, r.storageClaim.Name)
-				if err := r.createOrReplaceVolumeSnapshotClass(volumeSnapshotClass); err != nil {
+				if err := r.createOrReplaceClass(volumeSnapshotClass); err != nil {
 					return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeSnapshotClass: %s", err)
 				}
+			case "VolumeGroupSnapshotClass":
+				// check for CRD availability
+				if r.AvailableCrds[("VolumeGroupSnapshotClass")] {
+					var volumeGroupSnapshotClass *groupsnapapi.VolumeGroupSnapshotClass
+					data := map[string]string{}
```
rewantsoni, line 473: We need to unmarshall the data we receive from provider here,
```
				data := map[string]string{}
				err = json.Unmarshal(resource.Data, &data)
				if err != nil {
					return reconcile.Result{}, fmt.Errorf("failed to unmarshal StorageClaim configuration response: %v", err)
				}

**Review by rewantsoni — CHANGES_REQUESTED** [review]

**Comment by rewantsoni** [comment]
adding a soft LGTM and waiting for others to review

File: controllers/storageclaim_controller.go (hunk 35 above)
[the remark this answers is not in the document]
Comment (nb-ohad) [reply controllers/storageclaim_controller.go]: > You're describing the implementation of CreateOrUpdate

Yes, think about it as a `recreateInChanged` that accepts a resource and a mutate function. It internally creates a copy of the given resource, runs the mutate function on top of it, and then compares the original object with the new mutated one.  if they differ it calls Delete on the resource and then Create on the same resource with the mutated payload.

File: internal/controller/storageclaim_controller.go [hunk 41]
```diff
@@ -141,8 +154,10 @@ func (r *StorageClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotclasses,verbs=get;list;watch;create;delete
+//+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotclasses,verbs=get;list;watch
```
Madhu-1, line 157: we need delete access as well, can you please check

raaizik: And create

internal/controller/storageclaim_controller.go
```diff
@@ -170,6 +185,14 @@ func (r *StorageClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	utils.AssertEqual(r.AvailableCrds[drClusterConfigCRDName], crd.UID != "", utils.ExitCodeThatShouldRestartTheProcess)
 
+	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+	crd.Name = volumeGroupSnapshotClassCrd
+	if err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
+		r.log.Error(err, "Failed to get CRD", "CRD", volumeGroupSnapshotClassCrd)
+		return reconcile.Result{}, err
+	}
+	utils.AssertEqual(r.AvailableCrds[volumeGroupSnapshotClassCrd], crd.UID != "", utils.ExitCodeThatShouldRestartTheProcess)
+
```
rewantsoni, line 195: We can put this watch and the previous one under a loop,
```
	for _, crdName := range []string{volumeGroupSnapshotClassCrd, drClusterConfigCRDName} {
		crd := &metav1.PartialObjectMetadata{}
		crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
		crd.Name = crdName
		if err := r.Client.Get(ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
			r.Log.Error(err, "Failed to get CRD", "CRD", crd.Name)
		}
		util.AssertEqual(r.AvailableCrds[crd.Name], crd.UID != "", util.ExitCodeThatShouldRestartTheProcess)
	}
```

internal/controller/storageclaim_controller.go
```diff
@@ -126,7 +127,19 @@ func (r *StorageClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Owns(&csiopv1a1.ClientProfile{}, builder.WithPredicates(generationChangePredicate)).
+		Watches(
+			&extv1.CustomResourceDefinition{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				utils.NamePredicate(volumeGroupSnapshotClassCrd),
+				utils.CrdCreateAndDeletePredicate(&r.log, volumeGroupSnapshotClassCrd, r.AvailableCrds[volumeGroupSnapshotClassCrd]),
+			),
+			builder.OnlyMetadata,
+		)
```
rewantsoni, line 139: Add this predicate in the previous watch of the CRD
```suggestion
		Watches(
			&extv1.CustomResourceDefinition{},
			&handler.EnqueueRequestForObject{},
			builder.WithPredicates(
				utils.NamePredicate(volumeGroupSnapshotClassCrd),
				utils.CrdCreateAndDeletePredicate(&r.log, volumeGroupSnapshotClassCrd, r.AvailableCrds[volumeGroupSnapshotClassCrd]),
			),
			builder.WithPredicates(
				utils.NamePredicate(drClusterConfigCRDName),
				utils.CrdCreateAndDeletePredicate(&r.log, drClusterConfigCRDName, r.AvailableCrds[drClusterConfigCRDName]),
			),
			builder.OnlyMetadata,
		)
```

internal/controller/storageclaim_controller.go
```diff
@@ -392,7 +415,13 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					storageClass = r.getCephRBDStorageClass(data)
 				}
 				utils.AddAnnotation(storageClass, storageClaimAnnotation, r.storageClaim.Name)
-				err = r.createOrReplaceStorageClass(storageClass)
+				err = r.createOrReplace(storageClass, func() error {
+					if storageClass.Parameters == nil {
+						storageClass.Parameters = map[string]string{}
+					}
+					storageClass.Parameters = data
```
rewantsoni, line 422: We don't need to set to map[string]string{} if we are setting to be equal to data in next line

internal/controller/storageclaim_controller.go
```diff
@@ -535,52 +602,65 @@ func (r *StorageClaimReconciler) getCephFSVolumeSnapshotClass(data map[string]st
...
+	if reflect.DeepEqual(existing, obj) {
 		return nil
 	}
 
-	// StorageClass already exists, but parameters have changed. Delete the existing StorageClass and create a new one.
-	if existing.UID != "" {
+	if err := r.delete(obj); err != nil {
+		return err
+	}
+	fmt.Println("deleted")
```
rewantsoni, line 648: please remove the print statement

internal/controller/storageclaim_controller.go
```diff
@@ -535,52 +602,65 @@ func (r *StorageClaimReconciler) getCephFSVolumeSnapshotClass(data map[string]st
...
 
-		// Since we have to update the existing StorageClass, so we will delete the existing StorageClass and create a new one.
-		r.log.Info("StorageClass needs to be updated, deleting it.", "StorageClass", klog.KRef(storageClass.Namespace, existing.Name))
+	if err := r.create(obj); err != nil {
+		return err
+	}
+	fmt.Println("created again")
```
rewantsoni, line 653: please remove the create statement

internal/controller/storageclaim_controller.go
```diff
@@ -535,52 +602,65 @@ func (r *StorageClaimReconciler) getCephFSVolumeSnapshotClass(data map[string]st
...
+	return volumegroupsnapshotclass
 }
 
-func (r *StorageClaimReconciler) createOrReplaceStorageClass(storageClass *storagev1.StorageClass) error {
-	existing := &storagev1.StorageClass{}
-	existing.Name = r.storageClaim.Name
-
-	if err := r.own(storageClass); err != nil {
-		return fmt.Errorf("failed to own storageclass: %v", err)
+func (r *StorageClaimReconciler) createOrReplace(obj client.Object, f controllerutil.MutateFn) error {
```
rewantsoni, line 620: This is a generic function can you move this to pkg/utils?

internal/controller/storageclaim_controller.go
```diff
@@ -598,6 +678,13 @@ func (r *StorageClaimReconciler) list(obj client.ObjectList, listOptions ...clie
 	return r.Client.List(r.ctx, obj, listOptions...)
 }
 
+func (r *StorageClaimReconciler) create(obj client.Object) error {
+	if err := r.Client.Create(r.ctx, obj); err != nil && !errors.IsNotFound(err) {
```
rewantsoni, line 682: can create return a notFound error?

rewantsoni (changes requested):

internal/controller/storageclaim_controller.go
raaizik: Removed both `create` and `delete` aux funcs post move to k8sutil

pkg/utils/k8sutils.go
```diff
@@ -113,3 +118,49 @@ func GetMD5Hash(text string) string {
...
+
+	if reflect.DeepEqual(existing, obj) {
+		return nil
+	}
+
+	if err := c.Delete(ctx, obj); err != nil {
+		return err
+	}
+
+	obj.SetResourceVersion("")
```
rewantsoni, line 150: Could you add a comment reg why this is required?
something like:
```
k8s doesn't allow us to create objects when resourceVersion is set, as we are DeepCopying the object, the resource version also gets copied, hence we need to set it to empty before creating it
```

internal/controller/storageclaim_controller.go
```diff
@@ -408,14 +422,38 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 				// storageclaim is clusterscoped resources using its
 				// hash as the clusterID
 				data["clusterID"] = r.storageClaimHash
-				if resource.Name == "cephfs" {
-					volumeSnapshotClass = r.getCephFSVolumeSnapshotClass(data)
-				} else if resource.Name == "ceph-rbd" {
-					volumeSnapshotClass = r.getCephRBDVolumeSnapshotClass(data)
-				}
+				volumeSnapshotClass = r.getCephDriverVolumeSnapshotClass(resource.Labels, resource.Annotations)
```
rewantsoni, line 425: Let's also set the Labels and annotation in CreateOrReplace for all resources

rewantsoni (changes requested):

Madhu-1:
@raaizik can we split this in to Two PR's one for volumesnapshotclass and another one for volumegroupsnapshotclass? we can get the volumesnapshotclass PR merged and we can visit the volumegroupsnapshotclass PR once we have the feature in betav1 state.

internal/controller/storageclaim_controller.go
```diff
@@ -594,3 +648,17 @@ func (r *StorageClaimReconciler) hasVolumeSnapshotContents() (bool, error) {
 
 	return false, nil
 }
+
+func (r *StorageClaimReconciler) hasVolumeGroupSnapshotContents() (bool, error) {
+	vscList := &groupsnapapi.VolumeGroupSnapshotContentList{}
+	if err := r.list(vscList, client.MatchingFields{vscClusterIDIndexName: r.storageClaimHash}); err != nil {
```
leelavg, line 654: Pls add IndexField on this object as well while setting manager similar to VOlumeSnapshotContent.

internal/controller/storageclaim_controller.go
```diff
@@ -141,6 +184,17 @@ func (r *StorageClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	r.ctx = ctrllog.IntoContext(ctx, r.log)
 	r.log.Info("Reconciling StorageClaim.")
 
+	crd := &metav1.PartialObjectMetadata{}
+	for _, crdName := range []string{volumeGroupSnapshotClassCrd, drClusterConfigCRDName} {
```
Madhu-1, line 188: why do we have `drClusterConfigCRDName` isnt this PR only for VGSC?

internal/controller/storageclaim_controller.go
```diff
@@ -401,7 +455,31 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
+					// hash as the clusterID
+					data["clusterID"] = r.storageClaimHash
+					volumeGroupSnapshotClass = r.getCephDriverVolumeGroupSnapshotClass(resource.Labels, resource.Annotations)
+					utils.AddAnnotation(volumeGroupSnapshotClass, storageClaimAnnotation, r.storageClaim.Name)
+					err = utils.CreateOrReplace(r.ctx, r.Client, volumeGroupSnapshotClass, func() error {
+						volumeGroupSnapshotClass.Parameters = data
+						return nil
+					})
+					if err != nil {
+						return reconcile.Result{}, fmt.Errorf("failed to create or update StorageClass: %s", err)
```
Madhu-1, line 481: `failed to create or update StorageClass` to `failed to create or update VolumeGroupSnapshotClass`

internal/controller/storageclaim_controller.go
```diff
@@ -550,6 +633,20 @@ func (r *StorageClaimReconciler) getCephRBDVolumeSnapshotClass() *snapapi.Volume
...
 
+func (r *StorageClaimReconciler) getCephDriverVolumeGroupSnapshotClass(
+	labels map[string]string, annotations map[string]string) *groupsnapapi.VolumeGroupSnapshotClass {
+	volumegroupsnapshotclass := &groupsnapapi.VolumeGroupSnapshotClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        r.storageClaim.Name,
+			Labels:      labels,
+			Annotations: annotations,
+		},
+		Driver:         templates.RBDDriverName,
```
Madhu-1, line 644: This should be configurable as we need to support for both cephfs and rbd

internal/controller/storageclaim_controller.go
```diff
@@ -401,7 +454,35 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					return nil
 				})
 				if err != nil {
-					return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeSnapshotClass: %s", err)
+					return reconcile.Result{}, fmt.Errorf("failed to create or update StorageClass: %s", err)
```
rewantsoni, line 457:
```suggestion
					return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeSnapshotClass: %s", err)
```
You are under the VolumeSnapshotClass case here not StorageClass

internal/controller/storageclaim_controller.go
```diff
@@ -401,7 +454,35 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
+					// generate a new clusterID for cephfs subvolumegroup, as
+					// storageclaim is clusterscoped resources using its
+					// hash as the clusterID
+					data["clusterID"] = r.storageClaimHash
+					driverName := templates.CephFsDriverName
+					if strings.Contains(strings.ToLower(resource.Name), "rbd") {
+						driverName = templates.RBDDriverName
+					}
+					volumeGroupSnapshotClass = r.getCephDriverVolumeGroupSnapshotClass(resource.Labels, resource.Annotations, driverName)
+					utils.AddAnnotation(volumeGroupSnapshotClass, storageClaimAnnotation, r.storageClaim.Name)
```
rewantsoni, line 478: Let's move everything under CreateOrReplace

internal/controller/storageclaim_controller.go
```diff
@@ -550,6 +636,20 @@ func (r *StorageClaimReconciler) getCephRBDVolumeSnapshotClass() *snapapi.Volume
 	return volumesnapshotclass
 }
 
+func (r *StorageClaimReconciler) getCephDriverVolumeGroupSnapshotClass(
+	labels map[string]string, annotations map[string]string, driver string) *groupsnapapi.VolumeGroupSnapshotClass {
```
rewantsoni, line 640: We could add the labels/annotations directly in the CreateOrReplace

rewantsoni: We might not need this function, we could have everything under CreateOrReplace.

rewantsoni (commented):
Could also test this PR on a cluster to see if the VGSC is created with the required fields

internal/controller/storageclaim_controller.go
raaizik: The `CreateOrReplace` body should contain only part of this snippet, as seen in current code

internal/controller/storageclaim_controller.go
raaizik: We do need it to create the VGSC resource prior to applying its labels, annotations, etc.

rewantsoni:
@raaizik Are the changes tested?

internal/controller/storageclaim_controller.go
```diff
@@ -550,6 +641,18 @@ func (r *StorageClaimReconciler) getCephRBDVolumeSnapshotClass() *snapapi.Volume
 	return volumesnapshotclass
 }
 
+func (r *StorageClaimReconciler) getCephVolumeGroupSnapshotClassByDriver(
+	driver string) *groupsnapapi.VolumeGroupSnapshotClass {
+	volumegroupsnapshotclass := &groupsnapapi.VolumeGroupSnapshotClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: r.storageClaim.Name,
```
Madhu-1, line 648: please make sure we use the same class names as internal mode

raaizik: I will do that, keeping track on [2859](https://github.com/red-hat-storage/ocs-operator/pull/2859)

Madhu-1 (commented):
/lgtm

Madhu-1:
Added hold, @rewantsoni any final comments, if changes looks good, please remove the hold

raaizik:
> Added hold, @rewantsoni any final comments, if changes looks good, please remove the hold

Local testing still WIP

Madhu-1:
> > Added hold, @rewantsoni any final comments, if changes looks good, please remove the hold
> 
> Local testing still WIP @Madhu-1

Thank you , Lets do a final testing and update the results.

PR #152, opened 2024-05-17
BUG 2276565: match csi-addons subscription channel to ocs-client-operator

rohan47:
this change will match csi-addons subscription channel with ocs-client-operator in 4.15

PR #173, opened 2024-06-21
BUG 2276565: controllers: fixed typo subscription to subscriptions

rohan47:
rbac generation was affected because of this typo
The bug was introduced in this [PR](https://github.com/red-hat-storage/ocs-client-operator/pull/152)

### DIFF
--- a/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
@@ -225,6 +225,15 @@ spec:
           - list
           - watch
         - apiGroups:
+          - operators.coreos.com
+          resources:
+          - subscriptions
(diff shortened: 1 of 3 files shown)


leelavg:
Pls add the BZ to the description.

leelavg:
Pls reflect on why this issue was caused, missing bundle generation and so not to repeat it again, generate the bundle as well.

rohan47:
> Pls add the BZ to the description.

Do I need to create a new bug or the original BZ should do?

leelavg:
> Do I need to create a new bug or the original BZ(removing Modified) should do?
- Move it to POST and do a refresh here
- It's also good to have a ref to previous PR which introduced the change.

leelavg (approved):

PR #176, opened 2024-07-02
Add support for noobaa on client clusters

ezio-auditore:
1.Fetch noobaa remote join secret `(noobaa-remote-join-secret-{clusterID})` from the provider server via `GetStorageConfig` Ref: https://github.com/red-hat-storage/ocs-operator/pull/2680
    1. `auth_token` to be used for authentication with provider noobaa
    2. `mgmt_addr` to be use for getting the address of the provider noobaa server

2. Creation of the join secret in client clusters `noobaa-remote-join-secret`
3. Creation of a remote Noobaa CR referencing the join secret
Part of [RHSTOR-5187](https://issues.redhat.com//browse/RHSTOR-5187)


service/status-report/main.go
```diff
@@ -206,3 +226,21 @@ func setPlatformInformation(ctx context.Context, cl client.Client, status interf
...
+		return "", "", fmt.Errorf("failed to unmarshall data: %v", err)
+	}
+	noobaaAuthToken, ok := data["auth_token"]
+	if !ok {
+		return "", "", fmt.Errorf("noobaa auth token not found")
+	}
+	noobaaMgmtAddress, ok := data["mgmt_addr"]
+	if !ok {
+		return "", "", fmt.Errorf("noobaa mgmt address not found")
+	}
```
rewantsoni, line 243: Isn't this a redundant check? We already check if the values are empty on the ocs-operator side

service/status-report/main.go
```diff
@@ -141,6 +151,16 @@ func main() {
...
+	joinSecret := &corev1.Secret{}
+	joinSecret.Name = "noobaa-remote-join-secret"
+	joinSecret.Namespace = operatorNamespace
+	joinSecret.Data = map[string][]byte{
+		"auth_token": []byte(noobaaAuthToken),
+		"mgmt_addr":  []byte(noobaaMgmtAddress),
+	}
+	if err = cl.Create(ctx, joinSecret); err != nil {
+		klog.Exitf("Failed to create join secret %s in namespace %s: %v", joinSecret.Name, joinSecret.Namespace, err)
+	}
```
rewantsoni, line 163: - We can create it where we find the secret like we are doing in other cases 
- Create will fail on the second reconcile. Should we use CreateOrUpdate instead?
- We should probably set owner ref on this as well

rewantsoni (changes requested):

service/status-report/main.go
ezio-auditore: I think it is ok, since I am extracting and checking at the same point

service/status-report/main.go
rchikatw: yes true either use CreateOrUpdate or handle already exist case.

service/status-report/main.go
```diff
@@ -141,6 +151,16 @@ func main() {
...
 		klog.Exitf("Failed to update mon configmap for storageClient %v: %v", storageClient.Status.ConsumerID, err)
 	}
+	joinSecret := &corev1.Secret{}
+	joinSecret.Name = "noobaa-remote-join-secret"
+	joinSecret.Namespace = operatorNamespace
+	joinSecret.Data = map[string][]byte{
+		"auth_token": []byte(noobaaAuthToken),
+		"mgmt_addr":  []byte(noobaaMgmtAddress),
+	}
+	if err = cl.Create(ctx, joinSecret); err != nil {
```
rchikatw, line 161: 2 things:
1. I believe we need to register the scheme here as it is part of the status reporter's job. We might also need back for it.
2. I am not sure this is the correct place to create the secret Because status reporters should report only reporting should not create things if this is decided then I am ok with. 
3. Would you like to try creating a secret every minute?

rchikatw:
It would be beneficial to include testing results and a PR description to avoid the need to navigate through the epic.

controllers/storageclient_controller.go
```diff
@@ -505,6 +517,32 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
 	return nil
 }
 
+func (r *StorageClientReconciler) reconcileRemoteNoobaa() error {
```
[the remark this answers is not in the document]
ezio-auditore: at lineNo 237 in controllers/storageclient_controller.go

service/status-report/main.go
ezio-auditore: Added to scheme and added rbac
Can you suggest a better way or place I can add for the secret creation?

controllers/storageclient_controller.go
```diff
@@ -227,6 +233,12 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 			return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with claim processed annotation: %v", err)
 		}
 	}
+	// reconcile  remote noobaa only if it is not a local client
+	if _, exists := r.storageClient.GetAnnotations()["is-local-client"]; !exists {
+		if err := r.reconcileRemoteNoobaa(); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
```
nb-ohad, line 241: Please move the reconciliation of noobaa above the code reconciling the default storage classes. We kept them last for a reason.

controllers/storageclient_controller.go
```diff
@@ -549,6 +561,32 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
 }
 
+func (r *StorageClientReconciler) reconcileRemoteNoobaa() error {
+	joinSecret := &corev1.Secret{}
+	joinSecret.Name = joinSecretName
+	joinSecret.Namespace = r.OperatorNamespace
+	if err := r.get(joinSecret); err != nil {
+		r.Log.Error(err, "failed to get the joinSecret for remote noobaa")
+		return err
+	}
```
nb-ohad, line 571: This is unnecessary.  Getting to this part of the code means that you already called GetConfig from the provider, which guarantees a joint secret was created. 

Let's remove this check

controllers/storageclient_controller.go
```diff
@@ -549,6 +561,32 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+		Namespace: r.OperatorNamespace,
+	}
+	utils.AddAnnotation(nb, "remote-client-noobaa", "true")
+	if err := r.own(nb); err != nil {
+		return fmt.Errorf("failed to own remote noobaa: %v", err)
+	}
+	if err := r.create(nb); err != nil && !kerrors.IsAlreadyExists(err) {
+		return fmt.Errorf("failed to create remote noobaa: %v", err)
+	}
+	return nil
```
nb-ohad, line 587: Please use CreateOrUpdate to reconcile the noobaa CR, without it all update operations will go directly into the kube API server and will cause a generation change will in turn requeue another event resulting in an infinite reconcile loop

service/status-report/main.go
nb-ohad, line 1: @ezio-auditore The reporter is not a reconciler and does not implement reconcile loop mechanics. 
The creation of the secret needs to happen as part of the reconciler and the information itself should come from the GetStorageConfig operation and not from the heartbeat operation. 

Please revert these changes and re-implement using the established guidelines around sending resources from provider to  client

controllers/storageclient_controller.go
```diff
@@ -98,6 +101,7 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&v1alpha1.StorageClaim{}).
 		Owns(&batchv1.CronJob{}).
 		Owns(&quotav1.ClusterResourceQuota{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&nbv1.NooBaa{}).
```
nb-ohad, line 104: Please add a generationChangedPredicate to this call, we are not interested in status or metadata changes.

nb-ohad (changes requested):

service/status-report/main.go
ezio-auditore: removed secret creation in status-reporter

service/status-report/main.go
ezio-auditore: @rchikatw 
removed secret creation from status-reporter and added it to storageclient_reconciler

controllers/storageclient_controller.go
ezio-auditore: Removed fetching joined secret again

controllers/storageclient_controller.go
ezio-auditore: Updated to use CreateOrUpdate

controllers/storageclient_controller.go
ezio-auditore: Added generationChangedPredicate call

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,56 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+	nb.Namespace = r.OperatorNamespace
+	nb.Spec.JoinSecret = &corev1.SecretReference{
+		Name:      joinSecretName,
+		Namespace: r.OperatorNamespace,
+	}
+	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, nb, func() error {
+		utils.AddAnnotation(nb, "remote-client-noobaa", "true")
+		if err := r.own(nb); err != nil {
+			return err
+		}
```
rchikatw, line 588: please set the spec after owning the secret. it should be inside CreateOrUpdate something like this.


```
_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, nb, func() error {
		utils.AddAnnotation(nb, "remote-client-noobaa", "true")
		if err := r.own(nb); err != nil {
			return err
		}
nb.Spec.JoinSecret = &corev1.SecretReference{
		Name:      joinSecretName,
		Namespace: r.OperatorNamespace,
	}
return nil
	})
```

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,56 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+	joinSecret.Namespace = r.OperatorNamespace
+	joinSecret.Data = map[string][]byte{
+		"auth_token": []byte(noobaaAuthToken),
+		"mgmt_addr":  []byte(noobaaMgmtAddress),
+	}
+	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, joinSecret, func() error {
+		if err := r.own(joinSecret); err != nil {
+			return err
+		}
+		return nil
```
rchikatw, line 617: same here setting spec inside CreateOrUpdate.

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,56 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
 	return nil
 }
 
+func (r *StorageClientReconciler) reconcileRemoteNoobaa() error {
+	// Create noobaa CR referencing the join secret
+	nb := &nbv1.NooBaa{}
+	nb.Name = "noobaa-remote"
+	nb.Namespace = r.OperatorNamespace
+	nb.Spec.JoinSecret = &corev1.SecretReference{
+		Name:      joinSecretName,
```
rewantsoni, line 581: Could we append the storageClient hash to noobaa and joinSecret, this will ensure if we move to support multiple storageClients in the same cluster we won't run into name collision

controllers/storageclient_controller.go
ezio-auditore: Updated

controllers/storageclient_controller.go
ezio-auditore: Updated

controllers/storageclient_controller.go
ezio-auditore: Sure, I will take it up in a different PR
I have raised an issue for the same
https://github.com/red-hat-storage/ocs-client-operator/issues/189

rchikatw:
+1

service/status-report/main.go
ezio-auditore: Moved creation of secret inside storageClient Reconciler and removed action from status reporter

main.go
```diff
@@ -71,6 +73,8 @@ func init() {
 	utilruntime.Must(opv1a1.AddToScheme(scheme))
 	utilruntime.Must(extv1.AddToScheme(scheme))
 	utilruntime.Must(quotav1.AddToScheme(scheme))
+	utilruntime.Must(nbapis.AddToScheme(scheme))
+	utilruntime.Must(corev1.AddToScheme(scheme))
```
leelavg, line 77: doesn't clientgoscheme already registers kube native APIs?

controllers/storageclient_controller.go
```diff
@@ -210,9 +217,25 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+				}
+
+				if err := r.reconcileNoobaaJoinSecret(data); err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+		}
+	}
+	// reconcile  remote noobaa only if it is not a local client
+	if _, exists := r.storageClient.GetAnnotations()["is-local-client"]; !exists {
```
leelavg, line 234: could you pls explain why we need to differentiate b/n native vs remote client?

ezio-auditore: We don't want to create a client noobaa on the same cluster where the noobaa system is present to which all other remote clients will connect to, hence the annotation to differentiate
Ref: https://github.com/red-hat-storage/ocs-operator/pull/2680/files#diff-3a3043bcbec115677bf1cd899c5c562a02eb74b2c874f80b42ff8bd6880c41f4R43

main.go
ezio-auditore: yes, it does removed corev1 from scheme in new commit

controllers/storageclient_controller.go
ezio-auditore: No it won't. The nooba on the provider side is the one which consists of the main noobaa system, similar to what we have today, no change in existing functionality to which other clients would connect to

leelavg:
I believe #161 need to be in before this PR could go as this PR depends on nb CRDs, isn't it?

ezio-auditore:
> I believe #161 need to be in before this PR could go as this PR depends on nb CRDs, isn't it?

Yes correct, without noobaa on the client cluster it wont work

controllers/storageclient_controller.go
```diff
@@ -210,9 +217,25 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 			if err := r.reconcileClusterResourceQuota(clusterResourceQuotaSpec); err != nil {
 				return reconcile.Result{}, err
 			}
+		case "Secret":
+			if eResource.Name == fmt.Sprintf("noobaa-remote-join-secret-%s", r.storageClient.Status.ConsumerID) {
+				data := map[string]string{}
+				if err := json.Unmarshal(eResource.Data, &data); err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to unmarshall noobaa join secret data: %v", err)
+				}
```
nb-ohad, line 225: Can you please move the unmarshaling above the condition? All secrets will need unmarshaling of their data.

controllers/storageclient_controller.go
```diff
@@ -210,9 +217,25 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 			if err := r.reconcileClusterResourceQuota(clusterResourceQuotaSpec); err != nil {
 				return reconcile.Result{}, err
 			}
+		case "Secret":
+			if eResource.Name == fmt.Sprintf("noobaa-remote-join-secret-%s", r.storageClient.Status.ConsumerID) {
```
nb-ohad, line 221: Why match the exact name, use `strings.hasPrefix` instead.
But also, is there a reason we need the suffix at all? there is only going to be a single join secret on this cluster

controllers/storageclient_controller.go
```diff
@@ -210,9 +217,25 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+				if err := r.reconcileNoobaaJoinSecret(data); err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+		}
+	}
+	// reconcile  remote noobaa only if it is not a local client
+	if _, exists := r.storageClient.GetAnnotations()["is-local-client"]; !exists {
+		if err := r.reconcileRemoteNoobaa(); err != nil {
+			return reconcile.Result{}, err
```
nb-ohad, line 236: 1. who will put the annotation on the client?
2. Doesn't noobaa reconciliation here depend on the existing of the secret? Aren't they coupled and if so why is this not dependent on the existence of the secret instead of an annotation?

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,57 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create remote noobaa: %v", err)
+	}
+	return nil
+}
+
+func (r *StorageClientReconciler) reconcileNoobaaJoinSecret(noobaaData map[string]string) error {
+	noobaaAuthToken, ok := noobaaData["auth_token"]
+	if !ok {
```
nb-ohad, line 600: What if it is found but have an empty value?

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,57 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+		return fmt.Errorf("noobaa auth token not found")
+	}
+	noobaaMgmtAddress, ok := noobaaData["mgmt_addr"]
+	if !ok {
```
nb-ohad, line 604: What if it is found but have an empty value?

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,57 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+		return fmt.Errorf("noobaa mgmt address not found")
+	}
```
nb-ohad, line 606: Why do we need to validate this when noobaa operator has the concern to validate this?

controllers/storageclient_controller.go
```diff
@@ -549,6 +572,57 @@ func (r *StorageClientReconciler) reconcileSharedfileStorageClaim() error {
...
+	joinSecret.Name = joinSecretName
+	joinSecret.Namespace = r.OperatorNamespace
+	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, joinSecret, func() error {
+		if err := r.own(joinSecret); err != nil {
+			return err
+		}
+		joinSecret.Data = map[string][]byte{
+			"auth_token": []byte(noobaaAuthToken),
+			"mgmt_addr":  []byte(noobaaMgmtAddress),
+		}
```
nb-ohad, line 617: The entire point of sending the secret data is that you don't need to rebuild it yourself! You don't need a special joined secret reconcile function just a general reconcile func for secrets!

nb-ohad (changes requested):

controllers/storageclient_controller.go
ezio-auditore: 1. ocs storageclient controller is going to put this annotation on the local client.
ref: https://github.com/red-hat-storage/ocs-operator/pull/2680/files#diff-3a3043bcbec115677bf1cd899c5c562a02eb74b2c874f80b42ff8bd6880c41f4R43
2. Yes but ocs-client needs to know if it will create the client noobaa CR, for provider clusters it will not create the same. Also local ocs-client will not have a join secret

controllers/storageclient_controller.go
ezio-auditore: These values will anyway be validated by remote noobaa. What I am doing here is extracting the data and creating a join secret

controllers/storageclient_controller.go
ezio-auditore: same as previous comment

controllers/storageclient_controller.go
ezio-auditore: But the reconciler will still have to unmarshall the data accordingly from GetStorageConfig, extract it and create a secret specific for the client

controllers/storageclient_controller.go
ezio-auditore: The condition is required since hub cluster won't be having the join secret. Unmarshalling is only necessary if it's a remote storage client

controllers/storageclient_controller.go
ezio-auditore: moved

controllers/storageclient_controller.go
ezio-auditore: Wil change the prefix matching in updated commit
This is to keep the code viable for multiple noobaa clients in one application cluster
Ref: https://github.com/red-hat-storage/ocs-client-operator/issues/189

controllers/storageclient_controller.go
ezio-auditore: Will remove validation in next updated commit

controllers/storageclient_controller.go
ezio-auditore: Also there might be future implementations where other secrets are fetched from provider.
We will need to differentiate between these too.

controllers/storageclient_controller.go
ezio-auditore: As discussed, not depending on the annotation, only existence of join secret

controllers/storageclient_controller.go
```diff
@@ -210,9 +215,50 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			secret.Name = eResource.Name
+			secret.Namespace = r.storageClient.Namespace
+			_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, secret, func() error {
+				if err := r.own(secret); err != nil {
+					return err
+				}
+				if secret.Data == nil {
+					secret.Data = map[string][]byte{}
+				}
+				secret.Data = data.Data
```
nb-ohad, line 233:
```suggestion
				secret.Data =  data.Data
```

controllers/storageclient_controller.go
```diff
@@ -210,9 +215,50 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+				return nil
+			})
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to create or update secret %v", err)
```
nb-ohad, line 237: Add the secret name to the error

nb-ohad (changes requested):

controllers/storageclient_controller.go
ezio-auditore: Updated

controllers/storageclient_controller.go
ezio-auditore: Added secret name to error

controllers/storageclient_controller.go
```diff
@@ -210,9 +215,47 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			secret.Namespace = r.storageClient.Namespace
+			_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, secret, func() error {
+				if err := r.own(secret); err != nil {
+					return err
+				}
+				secret.Data = data.Data
+				return nil
+			})
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to create or update secret %s :%v", secret.Name, err)
```
nb-ohad, line 234: 1. You have an issue with the position of the char ":"
2. Please capture the namespaced name 
2. Consider splitting it into a multiline statement (As of the length of the current line) 

```suggestion
				return reconcile.Result{}, fmt.Errorf(
					"failed to create or update secret %v: %v", 
					client.ObjectKeyFromObject(secret) , 
					err,
				)
```

nb-ohad (changes requested):

controllers/storageclient_controller.go
ezio-auditore: Updated

PR #178, opened 2024-07-08
Changes to read quota and created

rchikatw:
- Create the ClusterResourceQuota if spec for it is received as part of RPC call of storage client config
- Not yet verified initial PR
- tested creating sample application with more then available storage it failed to create PVC with error

`Error from server (Forbidden): error when creating "sampledeployment.yaml": persistentvolumeclaims "pvc-test" is forbidden: exceeded quota: ocs-storagecluster-quota, requested: requests.storage=50Gi, used: requests.storage=10Gi, limited: requests.storage=20Gi`

### DIFF
--- a/controllers/storageclaim_controller.go
+++ b/controllers/storageclaim_controller.go
@@ -491,6 +491,7 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string) 
 			Name: r.storageClaim.Name,
 			Annotations: map[string]string{
 				"description": "Provides RWO Filesystem volumes, and RWO and RWX Block volumes",
+				"reclaimspace.csiaddons.openshift.io/schedule": "@weekly",
 			},
 		},
 		ReclaimPolicy:        &pvReclaimPolicy,


rchikatw:
@leelavg

controllers/storageclient_controller.go
```diff
@@ -203,9 +205,57 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+
+			err = json.Unmarshal([]byte(clusterResourceQuotaSpec), &clusterResourceQuota)
+			if err != nil {
+				return fmt.Errorf("error: %v", err)
+			}
+			fmt.Printf("%+v\n", clusterResourceQuota)
+
+			_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, &clusterResourceQuota, func() error {
+				// cluster scoped resource owning namespace scoped resource which allows garbage collection
+				if err := r.own(&clusterResourceQuota); err != nil {
```
leelavg, line 244: miss leading comment, isn't clusterresourcequota a cluster scoped resource?

rchikatw: I copied this from some other controller and missed removing this comment. Thank you for noticing

controllers/storageclient_controller.go
```diff
@@ -203,9 +205,54 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
 	}
 
+	if err := r.reconcileClusterResourceQuota(externalClusterClient); err != nil {
+		return reconcile.Result{}, err
+	}
+
 	return reconcile.Result{}, nil
 }
 
+func (r *StorageClientReconciler) reconcileClusterResourceQuota(externalClusterClient *providerClient.OCSProviderClient) error {
```
leelavg, line 215: not sure this is enough, there could be more than 1 storageclient and you can't scope a single clusterquota resource to entire cluster but scope it to sum of all storageclasses that reference this storageclient and does this quota also takes snapshots into account?

controllers/storageclient_controller.go
```diff
@@ -203,9 +205,54 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			}
+			clusterResourceQuotaResource, ok := data["clusterResourceQuota"]
+			if !ok {
+				return fmt.Errorf("no key found in the map")
+			}
+			var clusterResourceQuota quotav1.ClusterResourceQuota
+			err = json.Unmarshal([]byte(clusterResourceQuotaResource), &clusterResourceQuota)
+			if err != nil {
+				return fmt.Errorf("error: %v", err)
+			}
```
nb-ohad, line 237: All of this will have to change after addressing the comments here: 
https://github.com/red-hat-storage/ocs-operator/pull/2685#pullrequestreview-2176607686

controllers/storageclient_controller.go
rchikatw: it won't consider volume snapshot because Snapshot YAML has the source field(inside spec section) pointing to PVC.

```
spec:
  source:
    persistentVolumeClaimName: pvc-test
```
and PVC spec has pointing `requests.storage`
```
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 20Gi
```
In CRC we are mentioning `requests.storage` so it will consider only PVC which has `request.storage` in it.

rchikatw: > not sure this is enough, there could be more than 1 storageclient and you can't scope a single clusterquota resource to entire cluster but scope it to sum of all storageclasses that reference this storageclient

Regarding this this is the limitation using ClusterResourceQuota and Eran is aware of it. @nb-ohad please correct me if I am wrong.

leelavg: iirc, the limitation that's discussed was more than one storageclaim referencing a single storageclient but not more than one storageclient is delimited by a single clusterresourcequota.

imho, from the start we are taking a holistic approach for accommodating more than one storageclient and deviating from that might create problems if we use multiple storageclients.

PR #183, opened 2024-07-18
RHSTOR-5883: Annotate RBD storageclasses for EncryptionKeyRotation

black-dragon74:
This PR adds the required EncrpytionKeyRotation annotations on RBD StorageClasses to automate key rotation operations.

The default schedule is: weekly

### DIFF
--- a/internal/controller/storageclaim_controller.go
+++ b/internal/controller/storageclaim_controller.go
@@ -51,6 +51,7 @@ import (
 const (
 	storageClaimFinalizer  = "storageclaim.ocs.openshift.io"
 	storageClaimAnnotation = "ocs.openshift.io/storageclaim"
+	keyRotationAnnotation  = "keyrotation.csiaddons.openshift.io/schedule"
 
 	pvClusterIDIndexName  = "index:persistentVolumeClusterID"
 	vscClusterIDIndexName = "index:volumeSnapshotContentCSIDriver"
@@ -496,6 +497,10 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string) 
 		AllowVolumeExpansion: &allowVolumeExpansion,
 		Provisioner:          csi.GetRBDDriverName(),
 		Parameters:           data,
+	}
+
+	if r.storageClaim.Spec.EncryptionMethod != "" {
+		utils.AddAnnotation(storageClass, keyRotationAnnotation, utils.CronScheduleWeekly)
 	}
 	return storageClass
 }
--- a/pkg/utils/k8sutils.go
+++ b/pkg/utils/k8sutils.go
@@ -43,6 +43,8 @@ const DesiredSubscriptionChannelAnnotationKey = "ocs.openshift.io/subscription.c
 
 const runCSIDaemonsetOnMaster = "RUN_CSI_DAEMONSET_ON_MASTER"
 
+const CronScheduleWeekly = "@weekly"
+
 // GetOperatorNamespace returns the namespace where the operator is deployed.
 func GetOperatorNamespace() string {
 	return os.Getenv(OperatorNamespaceEnvVar)


controllers/storageclaim_controller.go
```diff
@@ -499,6 +504,10 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
 		Provisioner:          csi.GetRBDDriverName(),
 		Parameters:           data,
 	}
+
+	if isEncryptionRequired {
+		storageClass.ObjectMeta.Annotations["keyrotation.csiaddons.openshift.io/schedule"] = "@weekly"
```
Madhu-1, line 509: this should be set only if https://github.com/red-hat-storage/ocs-client-operator/blob/9dea589a7a1f6cbeff1b2ddceb3ceae8e24037f3/api/v1alpha1/storageclaim_types.go#L51 is set.

black-dragon74:
@rchikatw PTAL.

rchikatw:
> @rchikatw PTAL.

I am not sure about this part. I think you may have tagged me incorrectly.

controllers/storageclaim_controller.go
```diff
@@ -499,6 +500,10 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
 		Provisioner:          csi.GetRBDDriverName(),
 		Parameters:           data,
 	}
+
+	if r.storageClaim.Spec.EncryptionMethod != "" {
+		storageClass.ObjectMeta.Annotations["keyrotation.csiaddons.openshift.io/schedule"] = "@weekly"
```
nb-ohad, line 505: Can you please move the annotation name onto a const at the top of the file, as well as the value? 
Also, why was weekly chosen over other periodicity?

nb-ohad (changes requested):

controllers/storageclaim_controller.go
black-dragon74: > Also, why was weekly chosen over other periodicity?

Values less than weekly would be too soon while values more than it would be too late. Weekly justifies itself as the sweet spot. This decision was also influenced by the fact that ReclaimSpace has its default schedule as weekly.

leelavg:
@black-dragon74 what is the pending work on this PR? only resolution of comments?

black-dragon74:
> @black-dragon74 what is the pending work on this PR? only resolution of comments?

None. Yes, the comments have been addressed :)

Madhu-1:
@black-dragon74 can you please resolve the conflict?

Madhu-1:
@black-dragon74 please fix the commitlint.

LGTM, @nb-ohad PTAL for final approval

internal/controller/storageclaim_controller.go
```diff
@@ -484,6 +487,7 @@ func (r *StorageClaimReconciler) getCephFSStorageClass(data map[string]string) *
 func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string) *storagev1.StorageClass {
 	pvReclaimPolicy := corev1.PersistentVolumeReclaimDelete
 	allowVolumeExpansion := true
+
```
nb-ohad, line 490:
```suggestion
```

internal/controller/storageclaim_controller.go
```diff
@@ -497,6 +501,10 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
 		Provisioner:          csi.GetRBDDriverName(),
 		Parameters:           data,
 	}
+
+	if r.storageClaim.Spec.EncryptionMethod != "" {
+		storageClass.ObjectMeta.Annotations[keyRotationAnnotation] = cronScheduleWeekly
```
nb-ohad, line 506: 1. ObjectMeta is an embedded struct, you don't need to explicitly specify it
2. You cannot be sure that the annotations map is not nil. you first need to check it

```suggestion
		if storageClass.Annotations == nil {
			storageClass.Annotations = map[string]string{}
		}
		storageClass.Annotations[keyRotationAnnotation] = cronScheduleWeekly
```

But you don't need to be explicit here because we already have a util that implements the above snippet
```suggestion
		utils.AddAnnotation(&storageClass, keyRotationAnnotation, cronScheduleWeekly)
```

nb-ohad (changes requested):

internal/controller/storageclaim_controller.go
black-dragon74: In this case, annotations are explicitly set at line 493 and if empty they are initialized first by `utils.AddAnnotation`. Am I missing something?

internal/controller/storageclaim_controller.go
```diff
@@ -51,9 +51,12 @@ import (
 const (
 	storageClaimFinalizer  = "storageclaim.ocs.openshift.io"
 	storageClaimAnnotation = "ocs.openshift.io/storageclaim"
+	keyRotationAnnotation  = "keyrotation.csiaddons.openshift.io/schedule"
 
 	pvClusterIDIndexName  = "index:persistentVolumeClusterID"
 	vscClusterIDIndexName = "index:volumeSnapshotContentCSIDriver"
+
+	cronScheduleWeekly = "@weekly"
```
rchikatw, line 59: It would be good if you could find a suitable place to add it globally because it is being used in many places now.

black-dragon74: How about I take it in a separate PR?

rchikatw: In this PR, you can move and use it. In a different PR, you can make changes to the existing code.

black-dragon74: Done

rchikatw: yes LGTM.

internal/controller/storageclaim_controller.go
nb-ohad: @black-dragon74 As a general remake, please do not mark a conversation provided as review comments as resolved. 

The decision on resolution needs to be decided after input from the person who submitted the review. It makes it very hard to track on re-review what was addressed correctly and what still needs discussion/additional fixes if someone who is not the review owner resolves the comment.

When you fix an issue leave a "fixed" or "done" reply on the comment thread. If the reviewer is a maintainer they will then resolve after verifying, if the reviewer is a non-maintainer (which means they do not have permission to resolve comments on other contributors` PRs), they will leave their ack comment or a thumbs-up emoji to indicate that the fix is accepted then you can resolve

internal/controller/storageclaim_controller.go
```diff
@@ -497,6 +498,10 @@ func (r *StorageClaimReconciler) getCephRBDStorageClass(data map[string]string)
 		Provisioner:          csi.GetRBDDriverName(),
 		Parameters:           data,
 	}
+
+	if r.storageClaim.Spec.EncryptionMethod != "" {
+		utils.AddAnnotation(storageClass, keyRotationAnnotation, utils.CronScheduleWeekly)
+	}
```
nb-ohad, line 504: I might be missing something, but it seems that this annotation is already present at line 493. What is the point of adding it again here? if the intent was to add a condition then it should not be present on line 493 (which I am not sure is the intent). They were the code is written right now, the condition itself is irrelevant and the annotation will always be present on the returned storage class.

There are 2 options here:
1. If having the annotation on the storageclass has no side-effects in cases where encryption is not enabled then these lines need to be removed.
2 If, on the other hand, having the annotation always set has side effects in non-encrypted cases, then you should delete line 493

nb-ohad (changes requested):

internal/controller/storageclaim_controller.go
black-dragon74: The annotation on line 493 pertains to `reclaim space` while the current one refers to `key rotation`.

nb-ohad: Checking...

nb-ohad:
@black-dragon74 
I gave an LGTM which should trigger merge after you remove the hold.
Is there a reason to keep the hold?

black-dragon74:
> Is there a reason to keep the hold?

Yes. We are leaning towards not adding the annotation by default (in this release) due to additional validations required to ensure the annotation is present only with SCs having Vault as the KMS backend.

I will update once we have a final say on this.

Regards

nb-ohad:
@black-dragon74 I am not sure this is even relevant here! client op does not support KMS integration. In that case, we need to drop this PR.

black-dragon74:
> client op does not support KMS integration.

I agree. Moreover, the only way to identify underlying KMS is to parse the related configmap, which does not seem ideal to me. Let's wait until we have a decision? I will drop the PR accordingly.

leelavg:
@black-dragon74 I remember seeing a revert in ocs-op wrt key rotation, is this the same PR and need a revert?

black-dragon74:
> Need a revert?

@nb-ohad ^^

Madhu-1:
> @black-dragon74 I remember seeing a revert in ocs-op wrt key rotation, is this the same PR and need a revert?

@leelavg the revert is required in ocs-operator/UI as we support different KMS in internal/external mode but with proper only Amazon KMS or vault will be support, that's why its not reverted here, we have plan to support key rotation for these 2 KMS

PR #185, opened 2024-07-19
controllers: kernel mount options for cephfs Encryption in transit

rohan47:
support encryption in transit for cephfs when kernel mount option is set to secure. This information is recieved from provider and processed during storage claim creation.

The PR on OCS operator side is: https://github.com/red-hat-storage/ocs-operator/pull/2707

### DIFF
--- a/controllers/storageclaim_controller.go
+++ b/controllers/storageclaim_controller.go
@@ -54,6 +54,8 @@ const (
 
 	pvClusterIDIndexName  = "index:persistentVolumeClusterID"
 	vscClusterIDIndexName = "index:volumeSnapshotContentCSIDriver"
+
+	kernelMountOptionsKey = "kernelmountoptions"
 )
 
 // StorageClaimReconciler reconciles a StorageClaim object
@@ -380,6 +382,9 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					csiClusterConfigEntry.CephFS.SubvolumeGroup = data["subvolumegroupname"]
 					// delete groupname from data as its not required in storageclass
 					delete(data, "subvolumegroupname")
+					csiClusterConfigEntry.CephFS.KernelMountOptions = data[kernelMountOptionsKey]
+					// delete kernelmountoptions from data as its not required in storageclass
+					delete(data, kernelMountOptionsKey)
 					storageClass = r.getCephFSStorageClass(data)
 				} else if resource.Name == "ceph-rbd" {
 					storageClass = r.getCephRBDStorageClass(data)
--- a/pkg/csi/monconfigmap.go
+++ b/pkg/csi/monconfigmap.go
@@ -50,7 +50,8 @@ type CephRBDSpec struct {
 }
 
 type CephFSSpec struct {
-	SubvolumeGroup string `json:"subvolumeGroup,omitempty"`
+	SubvolumeGroup     string `json:"subvolumeGroup,omitempty"`
+	KernelMountOptions string `json:"kernelMountOptions,omitempty"`
 }
 
 type ClusterConfig struct {


controllers/storageclaim_controller.go
```diff
@@ -382,6 +382,9 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					csiClusterConfigEntry.CephFS.SubvolumeGroup = data["subvolumegroupname"]
 					// delete groupname from data as its not required in storageclass
 					delete(data, "subvolumegroupname")
+					csiClusterConfigEntry.CephFS.KernelMountOptions = data["kernelmountoptions"]
```
Madhu-1, line 385: @rohan47 can you please define a const for `kernelmountoptions` and please link the ocs-operator PR to give context there is already PR which is doing this change in other places.

rohan47: Made the change, Updated PR description with ocs-operator PR link

Madhu-1:
LGTM

rohan47:
Tested the code changes

Csi config after the changes

```yaml
oc get cm ceph-csi-configs -o yaml
apiVersion: v1
data:
  config.json: '[{"clusterID":"81fd58def37a35d1b4d4bee7ba35c3c3","storageClientID":"fb3e4f11-a38b-4859-ad18-1e8d579eb72b",
"monitors":["10.0.16.90:3300","10.0.61.203:3300","10.0.59.136:3300"],"rbd":{"radosNamespace":"cephradosnamespace-f6e94668817e4eefeb19b7a402fd947e"}},{"clusterID":"fe09c94a50c72d033da43478c21b8314",
"storageClientID":"fb3e4f11-a38b-4859-ad18-1e8d579eb72b","monitors":["10.0.16.90:3300","10.0.61.203:3300","10.0.59.136:3300"],
"cephFS":{"subvolumeGroup":"cephfilesystemsubvolumegroup-38beca789f5bc10e556b9832fd1070f6",
"kernelMountOptions":"ms_mode=secure"}}]'
kind: ConfigMap
metadata:
  creationTimestamp: "2024-07-22T19:24:10Z"
  name: ceph-csi-configs
  namespace: openshift-storage
  ownerReferences:
  - apiVersion: v1
    blockOwnerDeletion: true
    controller: true
    kind: ConfigMap
    name: ocs-client-operator-config
    uid: 664e804f-fc77-4aef-9387-a3c39dcb0b00
  resourceVersion: "3135818"
  uid: 7dee1733-974c-48f3-9cc7-7dbab17aee62
```

controllers/storageclaim_controller.go
```diff
@@ -380,6 +382,9 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					csiClusterConfigEntry.CephFS.SubvolumeGroup = data["subvolumegroupname"]
 					// delete groupname from data as its not required in storageclass
 					delete(data, "subvolumegroupname")
+					csiClusterConfigEntry.CephFS.KernelMountOptions = data[kernelMountOptionsKey]
+					// delete kernelmountoptions from data as its not required in storageclass
+					delete(data, kernelMountOptionsKey)
```
nb-ohad, line 387: We should not piggyback information destination for one resource (csi configmap) as part of another resource (storage class) to send CSI config data. The two should not be coupled and the fact that you need to delete pieces of information from the incoming data proves you are using the wrong approach. 

Please send all of the CSI config information as its own resource (a configmap with a `config.json` key) and handle it in a separate switch case! (similar to other resources we are sending from the provider). As a rule of thumb, each switch case need to be able to act independently!

@leelavg please make sure @rohan47's contribution here aligns with the proper way we set to send information from provider to client.

nb-ohad (changes requested):

controllers/storageclaim_controller.go
leelavg: @nb-ohad the existing mechanism is also sending data as part of storageclass that corresponds to it and we pick the interested items to configmap and this is also following the same.

rohan47:
The feature of this PR is to be delivered as part of the csi-operator changes and is no longer required.

PR #190, opened 2024-07-25
Update controller-runtime to latest release

leelavg:
Multiple upcoming PRs are going to update the controller-runtime to latest which could potentially mix a lot vendoring w/ business logic and this PR updates it before hand exclusively.

Among other breaking changes in controller-runtime v0.18.0 we are effected only w/ the change of Decoder from struct to interface and so we no longer need to get a pointer to it.


leelavg:
comprehensive testing will be done/clubbed in another PR and rudimentary tests pass.

PR #194, opened 2024-07-25
hack: add verification of go deps in gh action

rewantsoni:
update Make file to add new targets to verify go deps and add it in the github action

### DIFF
--- a/.github/workflows/verify-generated.yaml
+++ b/.github/workflows/verify-generated.yaml
@@ -8,16 +8,22 @@ on:
       - hack/**
       - Makefile
       - config/manager/csi-images.yaml
+      - go.mod
+      - go.sum
+      - vendor
   pull_request:
     branches: ['*']
     paths:
       - hack/**
       - Makefile
       - config/manager/csi-images.yaml
+      - go.mod
+      - go.sum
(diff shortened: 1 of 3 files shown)


.github/workflows/verify-generated.yaml
```diff
@@ -17,13 +17,20 @@ on:
...
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
       - name: Verify changes to CSI images manifest
         run: make verify-csi-images-manifest
+
+      - name: Verify go deps
```
leelavg, line 35: just a thought, seeing the actions log it seems we are going to download deps everytime and when you run the same in your machine it'll get copied from global cache afaik.

do you think it's worth to cache the dependencies as well?

leelavg: disregard, I just [saw](https://github.com/actions/setup-go?tab=readme-ov-file#caching-dependency-files-and-build-outputs) cache is on by default

rchikatw:
LGTM

leelavg (approved):

PR #196, opened 2024-07-29
init container to decide allowing reconciliation of manager

leelavg:
we want to run manager only when required, i.e, we are going to reconcile and manage CSI among other operands. we are achieving that by adding an init container which only get successful if client-operator is deployed standalone or running in provider configured odf cluster.

Depends on: red-hat-storage/ocs-operator#2717
ref: red-hat-storage/odf-operator#450

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -177,6 +177,7 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=*
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=subscriptions,verbs=get;list;watch;update
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=get;list
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
--- a/service/deployment-guard/main.go
+++ b/service/deployment-guard/main.go
@@ -1,0 +1,3 @@
+package main
+
+func main() {}
--- a/service/deployment-guard/main.go
+++ b/service/deployment-guard/main.go
@@ -1,3 +1,98 @@
 package main
 
-func main() {}
+import (
+	"context"
+	"os"
+	"time"
+
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func main() {
+	// validations
+	operatorNamespace := os.Getenv(utils.OperatorNamespaceEnvVar)
+	if operatorNamespace == "" {
+		klog.Exitf("%s env var is empty", utils.OperatorNamespaceEnvVar)
+	}
+
+	// creation of kube client
+	scheme := runtime.NewScheme()
+	cfg, err := config.GetConfig()
+	if err != nil {
+		klog.Exitf("Failed to get config: %v", err)
+	}
+	cl, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Exitf("Failed to create controller runtime client: %v", err)
+	}
+	ctx := context.Background()
+
+	storageClusterCRD := &metav1.PartialObjectMetadata{}
+	storageClusterCRD.SetGroupVersionKind(
+		extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"),
+	)
+	storageClusterCRD.Name = "storageclusters.ocs.openshift.io"
+
+	// delay exponentially from half a sec and cap at 2 minutes
+	delayFunc := wait.Backoff{
+		Duration: 500 * time.Millisecond,
+		Factor:   2,
+		Jitter:   0.1,
+		Steps:    10,
+		Cap:      2 * time.Minute,
+	}.DelayFunc()
+
+	for !allowOperatorToRun(ctx, cl, operatorNamespace) {
+		time.Sleep(delayFunc())
+	}
+
+}
+
+func allowOperatorToRun(ctx context.Context, cl client.Client, namespace string) bool {
+	// verify presence of StorageCluster CRD
+	storageClusterCRD := &metav1.PartialObjectMetadata{}
+	storageClusterCRD.SetGroupVersionKind(
+		extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"),
+	)
+	storageClusterCRD.Name = "storageclusters.ocs.openshift.io"
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(storageClusterCRD), storageClusterCRD); client.IgnoreNotFound(err) != nil {
+		klog.Warning("Failed to find presence of StorageCluster CRD")
+		return false
+	}
+
+	if storageClusterCRD.UID != "" {
+		// StorageCluster CRD exists, wait till StorageCluster CR is configured in Provider mode
+		storageClusters := &metav1.PartialObjectMetadataList{}
+		storageClusters.SetGroupVersionKind(
+			schema.GroupVersionKind{
+				Group:   "ocs.openshift.io",
+				Version: "v1",
+				Kind:    "StorageCluster",
+			},
+		)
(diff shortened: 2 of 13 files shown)


internal/controller/operatorconfigmap_controller.go
```diff
@@ -241,6 +221,26 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			c.log.Error(err, "failed to perform precheck for deploying CSI")
 			return ctrl.Result{}, err
 		} else if deployCSI {
+			if err := c.reconcileWebhookService(); err != nil {
```
Madhu-1, line 224: should we move https://github.com/red-hat-storage/ocs-client-operator/blob/5e08b7357256da96ddb2448df9c81766ced21ed7/internal/controller/operatorconfigmap_controller.go#L215-L218 inside deployCSI as well? is there any case we cant to createConsolePlugin when we are not managing/deploying CSI?

leelavg: it isn't moved as that doesn't depend on CSI however console deployment is dependent on this reconcile result and so not moved.

Madhu-1: okay let me rephrase my question sorry for not being clear, do we need to deploy consolePlugin if we are not managing csi? i know both are not dependent but do we have any case like this?

leelavg: ack, I'll be reworking on this PR to shutdown unless explicitly asked to run as that should be the correct fix.

leelavg:
Testing:
```
- No StorageCluster CRD
I0729 14:00:24.636988 2124186 main.go:51] StorageCluster CRD not found and proceeding with reconciliation of operands

- CRD exists but not storagecluster
I0729 13:51:26.625930 2120304 main.go:77] Waiting for StorageCluster to be deployed
I0729 13:51:31.638621 2120304 main.go:77] Waiting for StorageCluster to be deployed
^Csignal: interrupt

- storagecluster exists but not in Provider mode
> k -nopenshift-storage get storagecluster/ocs-storagecluster -ojsonpath='{.metadata.annotations}' | jq 'del(."kubectl.kubernetes.io/last-applied-configuration")'
{
  "uninstall.ocs.openshift.io/cleanup-policy": "delete",
  "uninstall.ocs.openshift.io/mode": "graceful"
}
I0729 13:52:24.907032 2120630 main.go:79] Waiting for StorageCluster deployed mode to be 'Provider'
I0729 13:52:30.179057 2120630 main.go:79] Waiting for StorageCluster deployed mode to be 'Provider'
^Csignal: interrupt

> k -nopenshift-storage get storagecluster/ocs-storagecluster -ojsonpath='{.metadata.annotations}' | jq 'del(."kubectl.kubernetes.io/last-applied-configuration")'
{
  "mode": "provider",
  "uninstall.ocs.openshift.io/cleanup-policy": "delete",
  "uninstall.ocs.openshift.io/mode": "graceful"
}
I0729 13:56:41.060310 2123137 main.go:72] StorageCluster is configured to run in 'Provider' mode and proceeding with reconciliation of operands
```

update to the client-console deployment is yet to be tested.

service/reconcile-waiter/main.go
nb-ohad, line 1: 1. That is a very undescriptive name, in my opinion, for the functionality provided by this program. 
If you look at the definition of a waiter you will find the following:
`a man whose job is to serve customers at their tables in a restaurant.`



Let's rename this program to something like `deployment-guard`

service/reconcile-waiter/main.go
```diff
@@ -0,0 +1,84 @@
...
+				}
+			}
+			if len(storageClusters.Items) == 0 {
+				klog.Info("Waiting for StorageCluster to be deployed")
+			} else {
+				klog.Info("Waiting for StorageCluster deployed mode to be 'Provider'")
+			}
+			time.Sleep(time.Duration(intervalSeconds) * time.Second)
+		}
+	}
```
nb-ohad, line 83: This code should be organized for readability, which means we should minimize the exit flows
My suggestion is (in pseudo-code)
```
1. Read CRD into a var, 
   1.1 sleep then retry read (we should fail on error because it will lead us into a crush loop backoff)
2. If CRD exists 
   2.1 Read storageclsuter CR metadata
   2.2 if not exists or not provider sleep + loop
3. Return peacefully

File: service/reconcile-waiter/main.go [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const intervalSeconds = 5
```
nb-ohad, line 20: Fixed 5-second interval should not work here. Preferably to have an growing interval as this pod might be looping a lot

nb-ohad (changes requested):
Why do we need a new service account an a new RBAC? 
Why can't we just use the operators SA and RBAC?

service/reconcile-waiter/main.go
leelavg: > definition of a waiter
- no need to take the literal meaning of the used words, it is similar to existing golang interfaces, reader - one who reads, writer - one who writes, waiter - one who waits

> like deployment-guard
- we are not guarding against deployment, could possibly be reconcile-guard

service/reconcile-waiter/main.go
leelavg: since there is a wait loop (3) has to be in (1.2) and (2.3) which is what I'm doing above.

> will lead us into a crush loop backoff
- yes, I thought of CLBO but left it as kubelet will provide incremental backoff w/ CLBO and so didn't expand the code
- as you also want it to fold into code, I could do that, no qualms

service/reconcile-waiter/main.go
leelavg: will implement a backoff.

leelavg:
> Why do we need a new service account an a new RBAC? Why can't we just use the operators SA and RBAC?
1. We have two deployments in CSV, client-op (custom sa) and client-console (default sa)
2. We need to inject sidecar into both of these deployments, however custom sa in client-op is enough for it's init containers but using the same which has more privilege for init container in client-console didn't seem good
3. Reusing client-op sa for client-console reduces good chunk of rbac in this PR which I could do based on your comment, wdyt?

-----

I realized something when replying to the comments, CSV will only be successful if Pod is in Running state, with an init container, pod will only be in Running state after init which could potentially fail d/s CI. Overall we might need to rethink on this.

service/reconcile-waiter/main.go
nb-ohad: Sorry, but it is not the same in English:
- A reader is someone who reads
- A writer is someone who writes
- A waiter is someone who serves

service/reconcile-waiter/main.go
nb-ohad: >since there is a wait loop (3) has to be in (1.2) and (2.3) which is what I'm doing above.
No, you don't need to there should only be one wait loop and one return

nb-ohad:
> > Why do we need a new service account an a new RBAC? Why can't we just use the operators SA and RBAC?
> 
> 1. We have two deployments in CSV, client-op (custom sa) and client-console (default sa)
> 2. We need to inject sidecar into both of these deployments, however custom sa in client-op is enough for it's init containers but using the same which has more privilege for init container in client-console didn't seem good
> 3. Reusing client-op sa for client-console reduces good chunk of rbac in this PR which I could do based on your comment, wdyt?
> 
> I realized something when replying to the comments, CSV will only be successful if Pod is in Running state, with an init container, pod will only be in Running state after init which could potentially fail d/s CI. Overall we might need to rethink on this.

CI should take into account that this pod is waiting on installation, not the other way around

service/reconcile-waiter/main.go
leelavg: that's fine, my main question is `reconcile-guard` vs `deployment-guard`

leelavg: done.

service/reconcile-waiter/main.go
leelavg: done.

service/reconcile-waiter/main.go
leelavg: done.

internal/controller/operatorconfigmap_controller.go
leelavg: done.

leelavg:
Tested:
1. client-op is running standalone : proceed
2. client-op is installed alongside odf with
 a. storagecluster doesn't exist : wait
 b. storagecluster not in provider mode: wait
 c. storagecluster in provider mode: proceed

leelavg:
I have raised tracker issues for other u/s projects which tries to consume client-op as olm bundles, need to fix them if those projects required updated client-op bundles:
red-hat-storage/ocs-operator#2722
red-hat-storage/odf-operator#461

service/deployment-guard/main.go
```diff
@@ -0,0 +1,98 @@
...
+		)
+		if err := cl.List(ctx, storageClusters, client.InNamespace(namespace), client.Limit(1)); err != nil {
+			klog.Warning("Failed to list StorageCluster CR")
+			return false
+		}
+		if len(storageClusters.Items) < 1 {
+			klog.Info("StorageCluster CR does not exist")
+			return false
+		}
+		klog.Info("Checking for StorageCluster configuration be 'Provider' mode")
```
nb-ohad, line 90:
```suggestion
		klog.Info("Checking if StorageCluster indicates ODF is deployed in provider mode")
```

nb-ohad (changes requested):

service/deployment-guard/main.go
leelavg: done.

PR #199, opened 2024-08-01
hack: update csv timestamp only when there are changes in bundle

rewantsoni:
This PR includes the following changes
1. add a script not to update the timestamp in the csv if there are no bundle changes
2. add ci action to verify bundle changes

### Why?
With operator-sdk 1.26+ every time we generate a new bundle, the createdAt field in the csv is updated. 
This means we have to create a manual backport even when there are no bundle changes hence, resetting the createdAt field if there are no changes in the bundle.

This PR also introduces a ci action to check whether there are any uncommitted bundle changes.

### DIFF
--- a/hack/update-csv-timestamp.sh
+++ b/hack/update-csv-timestamp.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+git diff --quiet -I'^( )+createdAt: ' bundle
+if ((! $?)) ; then
+    git checkout --quiet bundle
+fi
--- a/.github/workflows/verify-generated.yaml
+++ b/.github/workflows/verify-generated.yaml
@@ -40,3 +40,9 @@ jobs:
 
       - name: Verify go deps
         run: make godeps-verify
+
+      - name: Verify bundle
+        run: |
+          make bundle
+          msg='Uncommitted bundle changes. Run `make bundle` and commit results.'
+          git diff --exit-code bundle || (echo -e '\e[31m'"$msg"; exit 1)
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,7 @@ bundle: manifests kustomize operator-sdk yq ## Generate bundle manifests and met
 	yq -i '.dependencies[0].value.version = "'${CSI_ADDONS_PACKAGE_VERSION}'"' config/metadata/dependencies.yaml
 	cp config/metadata/* bundle/metadata/
 	$(OPERATOR_SDK) bundle validate ./bundle
+	hack/update-csv-timestamp.sh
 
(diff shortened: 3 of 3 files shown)


hack/update-csv-timestamp.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
leelavg, line 1: I saw the issue where this was mentioned, in our case this will render w/ d/s bundle date not being updated.

Maybe we need to settle w/ manual backports if CSV is updated and separate that into an individual commit so that other commits part of same PR can be cherry-picked and bundle can be generated.

How about we check bundle generation diff w/ removing this field as done in https://github.com/red-hat-storage/odf-operator/commit/3e4f3eacff049a727348c6592a4b32d3698ae10c?

In addition to work done in #194, above does the same for bundle? wdyt?

rewantsoni: yes we can do that

nb-ohad:
@rewantsoni 
Technically I understand what you are doing.
But nowhere do you menton the reason why we are doing it. Can you please elaborate on that? 
Preferably add the reasoning to the description?

leelavg (approved):
I am not able to find the help of `git diff -I` option but the fix is working for me locally and so approving.

leelavg:
removed label due to a hold if there are any further questions.

rewantsoni:
@nb-ohad I have updated the PR description

nb-ohad:
> resetting the createdAt field if there are no changes in the bundle.
Ok, but if there are changes in the bundle, would we have the updated createdAt time or not?

rewantsoni:
> > resetting the createdAt field if there are no changes in the bundle.
> 
> Ok, but if there are changes in the bundle, would we have the updated createdAt time or not?

Yes, if the changes are in bundle the createdAt timestamp will get updated

leelavg:
> This means we have to create a manual backport even when there are no bundle changes hence, resetting the createdAt field if there are no changes in the bundle.
- I realized the issue was faced when there was a direct merge to release branch and this PR would help in that case, moreover it doesn't need to work in that case.

Overall this PR does address CI issues and some, will merge when Ohad acks https://github.com/red-hat-storage/ocs-client-operator/pull/199#issuecomment-2267443594

leelavg:
@rewantsoni possible to push this PR forward?

rewantsoni:
@nb-ohad Can you review this?

leelavg (approved):

rewantsoni:
synced with @nb-ohad offline and unholding the PR
/unhold

PR #200, opened 2024-08-01
controllers: trigger reconcile event to storage client if hash of desired config does not match

rchikatw:
- Save hash to storage client status
- add an annotation for the subscription channel to the storage client
- Add a new field to the status of the storage client
- Testing is pending will add results once it's done
- Depends on https://github.com/red-hat-storage/ocs-operator/pull/2676

### DIFF
--- a/pkg/templates/csi.go
+++ b/pkg/templates/csi.go
@@ -7,6 +7,7 @@ import (
 	secv1 "github.com/openshift/api/security/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/ptr"
 )
 
@@ -73,10 +74,27 @@ const CSIOperatorConfigName = "ceph-csi-operator-config"
 
 var CSIOperatorConfigSpec = csiopv1a1.OperatorConfigSpec{
 	DriverSpecDefaults: &csiopv1a1.DriverSpec{
+		Log: &csiopv1a1.LogSpec{
+			Rotation: &csiopv1a1.LogRotationSpec{
+				Periodicity: csiopv1a1.DailyPeriod,
+				MaxLogSize:  resource.MustParse("500M"),
+				MaxFiles:    7,
+				LogHostPath: "/var/lib/cephcsi",
+			},
+		},
 		AttachRequired:  ptr.To(true),
 		DeployCsiAddons: ptr.To(true),
 		FsGroupPolicy:   storagev1.FileFSGroupPolicy,
 		ControllerPlugin: &csiopv1a1.ControllerPluginSpec{
+			Privileged: ptr.To(true),
+			Resources: csiopv1a1.ControllerPluginResourcesSpec{
+				LogRotator: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("100m"),
+						corev1.ResourceMemory: resource.MustParse("32Mi"),
+					},
+				},
+			},
 			PodCommonSpec: csiopv1a1.PodCommonSpec{
 				PrioritylClassName: ptr.To("system-cluster-critical"),
 				ImagePullPolicy:    corev1.PullIfNotPresent,
@@ -84,6 +102,14 @@ var CSIOperatorConfigSpec = csiopv1a1.OperatorConfigSpec{
 			Replicas: ptr.To(int32(2)),
 		},
 		NodePlugin: &csiopv1a1.NodePluginSpec{
+			Resources: csiopv1a1.NodePluginResourcesSpec{
+				LogRotator: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("100m"),
+						corev1.ResourceMemory: resource.MustParse("32Mi"),
+					},
+				},
+			},
 			KubeletDirPath: "/var/lib/kubelet",
 			PodCommonSpec: csiopv1a1.PodCommonSpec{
 				PrioritylClassName: ptr.To("system-node-critical"),


service/status-report/main.go
```diff
@@ -109,12 +109,15 @@ func main() {
...
-			klog.Exitf("Failed to annotate storageclient %q: %v", storageClient.Name, err)
+	// statusResponse.DesiredClientOperatorChannel once ocs-o/r or merges will change this to read hash
+	if statusResponse.DesiredClientOperatorChannel != storageClient.Status.DesiredClientConfigHash {
+		storageClientCopy := &v1alpha1.StorageClient{}
+		storageClient.DeepCopyInto(storageClientCopy)
+		if utils.AddAnnotation(storageClient, utils.DesiredClientConfigAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
+			// patch is being used here as to not have any conflicts over storageclient cr changes as this annotation value doesn't depend on storageclient spec
+			if err := cl.Patch(ctx, storageClient, client.MergeFrom(storageClientCopy)); err != nil {
+				klog.Exitf("Failed to annotate storageclient %q: %v", storageClient.Name, err)
+			}
```
leelavg, line 120: the idea is you only set the hash and controller takes the action if hash is changed but not split action b/n reporter & controller.

internal/controller/storageclient_controller.go
```diff
@@ -210,7 +210,21 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshal ConfigMap data: %v", err)
+			}
+			if eResource.Name == "DesiredClientConfig" {
+				r.storageClient.Status.DesiredClientConfigHash = utils.GetMD5Hash(fmt.Sprintf("%v", data))
+				utils.AddAnnotation(r.storageClient, utils.DesiredSubscriptionChannelAnnotationKey, data["DesiredClientOperatorChannel"])
+				if err := r.update(r.storageClient); err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with DesiredClientOperatorChannel annotation: %v", err)
+				}
+			}
```
leelavg, line 225: thinking about it, we don't need a specialized structure for the hash. Provider has a state spread across multiple resources, you consolidate that state and send that as hearbeat response which triggers a reconcile on storageclient.

During the reconcile, it gets the info that it requires via usual resources, be it configmap, quota, new csi-op APIs etc. We are only trying to invoke the controller when it is required to reconcile and hash is the differentiating factor.

rchikatw: okay, I'm removing it in the updated commit.

internal/controller/storageclient_controller.go
```diff
@@ -235,7 +235,19 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			data := map[string]string{}
+			err := json.Unmarshal(eResource.Data, &data)
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshal ConfigMap data: %v", err)
+			}
+
+			utils.AddAnnotation(r.storageClient, utils.DesiredSubscriptionChannelAnnotationKey, data["ocs-operator"])
+			if err := r.update(r.storageClient); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with subscription channel annotation: %v", err)
+			}
```
nb-ohad, line 248: This change around the subscription notification mechanism is a different concern and should be moved to its own PR

service/status-report/main.go
```diff
@@ -117,12 +117,12 @@ func main() {
...
-	if utils.AddAnnotation(storageClient, utils.DesiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
-		// patch is being used here as to not have any conflicts over storageclient cr changes as this annotation value doesn't depend on storageclient spec
-		if err := cl.Patch(ctx, storageClient, client.MergeFrom(storageClientCopy)); err != nil {
-			klog.Exitf("Failed to annotate storageclient %q: %v", storageClient.Name, err)
+	// statusResponse.DesiredClientOperatorChannel will be replaced with hash field once ocs-o/r PR get merge.
+	if statusResponse.DesiredClientOperatorChannel != storageClient.Status.ProviderState {
+		storageClient.Status.ProviderState = statusResponse.DesiredClientOperatorChannel
+		statusErr := cl.Status().Update(ctx, storageClient)
+		if statusErr != nil {
+			klog.Exitf("Failed to update StorageClient status %v: %v", storageClient.Status.ConsumerID, err)
```
nb-ohad, line 125: This change around the subscription notification mechanism is a different concern and should be moved to its own PR

service/status-report/main.go
nb-ohad, line 1: This PR is missing the code to update the annotation on the client with the hash coming from the heartbeat

api/v1alpha1/storageclient_types.go
```diff
@@ -52,6 +52,9 @@ type StorageClientStatus struct {
 
 	// ConsumerID will hold the identity of this cluster inside the attached provider cluster
 	ConsumerID string `json:"id,omitempty"`
+
+	// ProviderState will hold the hash of provider state of client
+	ProviderState string `json:"providerState,omitempty"`
```
nb-ohad, line 57: `ProviderState` is the wrong none to use here, everything captured inside that hash is about consumer configuration not provider state. Also, a timestamp would really help in debugging


```
ConsumerConfig: 
    hash: <the  hash>
    timestamp: <the last time the hash changed>
```

@leelavg @rchikatw I don't really like the term `ConsumerConfig` so if you can come up with an alternative name it will be great!

nb-ohad (changes requested):

service/status-report/main.go
rchikatw: this is present at here: https://github.com/red-hat-storage/ocs-client-operator/pull/200/files#diff-67ad3ddf7a24ad71e9510d7547db6201545f32798543ef7d0752cc01b7fe6be7R122 I feel that setting an annotation is unnecessary. If the hash is different, I will update the status, and the reconciliation should trigger. I have yet to verify if the reconciliation will trigger or not.

rchikatw: Or, if you want the status setting to be part of the storage client controller, then I just have to set the annotation as part status reporter let me know @nb-ohad . The reason I have done it this way is because I don't want to calculate the hash on the client side.

service/status-report/main.go
```diff
@@ -117,12 +118,13 @@ func main() {
...
 	}
 
-	storageClientCopy := &v1alpha1.StorageClient{}
-	storageClient.DeepCopyInto(storageClientCopy)
-	if utils.AddAnnotation(storageClient, utils.DesiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
-		// patch is being used here as to not have any conflicts over storageclient cr changes as this annotation value doesn't depend on storageclient spec
-		if err := cl.Patch(ctx, storageClient, client.MergeFrom(storageClientCopy)); err != nil {
-			klog.Exitf("Failed to annotate storageclient %q: %v", storageClient.Name, err)
+	// statusResponse.DesiredClientOperatorChannel will be replaced with hash field once ocs-o/r PR get merge.
+	if statusResponse.DesiredClientOperatorChannel != storageClient.Status.ConsumerConfig.Hash {
```
leelavg, line 122: NO, what is the conditional trying to do comparing two different fields?

service/status-report/main.go
leelavg: > he reason I have done it this way is because I don't want to calculate the hash on the client side.
- I don't think you will be calculating it again irrespective of storing it in annotation/status.

rchikatw: @leelavg Yes, that is true. In that case, we don't even need to save the hash in the status. We can compare the hash from the annotation of the storage client CR, right? @nb-ohad "I don't see a reason for having hash in the storage client status section." wdyt?

nb-ohad: @rchikatw I think you are missing the point of the annotation+status mechanism. 
The status of a CR always reflects the actual status, never the desired status. This is why it is highly not recommended to change that status outside of a controller! This means that the heartbeat mechanism would not update the status of any CR.

The annotation is there to indicate to the controller the new desired state. An annotation can be the input or output of a controller (like any other metadata field). I would expect that the server will use something like `CreateOrUpdate`to make sure that the annotation is not being updated if the new value is exactly the same as the old value - you can get this for free with `CreateOrUpdate`, or you can write the code yourself. The bottom line is that we do not create a reconcile request unless something changes on the CR. 
The controller itself does not care about the annotation at all! when a reconcile happens (for any reason) the new config + new hash will be pulled from the provider

nb-ohad:
> I don't think you will be calculating it again irrespective of storing it in annotation/status.

@leelavg For the client-side code, the hash is opaque. This is a main pillar of the design! 
It is a magic number that changes whenever new state need to be "pushed" to the client

leelavg:
> > I don't think you will be calculating it again irrespective of storing it in annotation/status.
> 
> @leelavg For the client-side code, the hash is opaque. This is a main pillar of the design! 
> It is a magic number that changes whenever new state need to be "pushed" to the client

- yes, where did I mention it as otherwise 🤔?

rchikatw:
Something i did wrong and due to which my commit is removed and hence github assumes now this as nothing to merge and closed the PR. Will raise new PR I don't see a way to get my changes back.

leelavg:
> Something i did wrong and due to which my commit is removed and hence github assumes now this as nothing to merge and closed the PR. Will raise new PR I don't see a way to get my changes back.
- not a good idea closing this PR as good amount of discussion happened on this.
- you can get local commits w/ `git reflog` unless you nuked the dir, you can pull from that and update this PR.
- I suggest to grok git a bit as it's quite rare (unless a human error) to loose track of once commited code

PR #208, opened 2024-08-19
controllers: wait for 90 sec before checking StorageCluster CRD

leelavg:
when we are running alongside odf-operator the CRDs corresponding to an operator is installed first however it isn't guaranteed that CRDs of all dependent operators are installed first before deploying any CSV.

Due to above, StorageCluster CRD is sometimes created after we check for it's existence and 1 min is the delay observed when client-op is listed first and ocs-op listed last in the install plan for deploying CSVs.

Please note this should be temporary upto when we develop new operator which directly installs other operators based on runtime requirements outside of OLM.

This PR imposes a mandatory wait time for standalone client-op installs.

### DIFF
--- a/service/deployment-guard/main.go
+++ b/service/deployment-guard/main.go
@@ -36,6 +36,20 @@ func main() {
 	}
 	ctx := context.Background()
 
+	// NOTE: when we are running alongside odf-operator the CRDs corresponding
+	// to an operator is installed first however it isn't guaranteed that CRDs
+	// of all dependent operators are installed first before deploying any CSV.
+	//
+	// Due to above, StorageCluster CRD is sometimes created after we check for
+	// it's existence and 1 min is the delay observed when client-op is listed
+	// first and ocs-op listed last in the install plan for deploying CSVs.
+	//
+	// Please note this should be temporary upto when we develop new operator
+	// which directly installs other operators based on runtime requirements
+	// outside of OLM.
+	klog.Info("Waiting for 90 sec before checking to allow operator to run")
+	time.Sleep(90 * time.Second)
+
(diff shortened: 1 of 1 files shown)


leelavg:
/hold

for 2 reviews.

service/deployment-guard/main.go
```diff
@@ -36,6 +36,21 @@ func main() {
...
+	// Due to above, StorageCluster CRD is sometimes created after we check for
+	// it's existence and 2 minutes is chosen as it is double the time observed
+	// when client-op is listed first and ocs-op listed last in the install plan
+	// for deploying CSVs.
+	//
+	// Please note this should be temporary upto when we develop new operator
+	// which directly installs other operators based on runtime requirements
+	// outside of OLM.
+	klog.Info("Waiting for 2 minutes before checking to allow operator to run")
+	time.Sleep(2 * time.Minute)
```
Madhu-1, line 52: why not increase the Cap instead of putting a hard sleep of 2 minutes?

leelavg: that'll not work, there are two branches, the backoff is for what happens after we found the existence of CRD, the hard sleep is waiting to check for the CRD creation.

Madhu-1: IMO hard sleep might work in all cases but again it can cause issues as well if the timing is not met (in worst cases), can we check for CRD existing as well allowOperatorToRun function?

Madhu-1: got to know that we will not have CRD when running in consumer mode, resolving now

leelavg:
/hold

Will expand solution on not to wait everytime.

PR #211, opened 2024-08-19
controllers: read subscription as part ext resource

rchikatw:
- read subscription as part ext resource
- set the OCS-operator channel to storage client annotation

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -121,6 +121,7 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=cephconnections,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=noobaa.io,resources=noobaas,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=subscriptions,verbs=get;list;watch;update
 
 func (r *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var err error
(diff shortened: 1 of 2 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -235,6 +235,16 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 					return reconcile.Result{}, fmt.Errorf("failed to reconcile cephConnection: %v", err)
 				}
 			}
+		case "Subscription":
+			var subscriptionSpec *opv1a1.SubscriptionSpec
+			if err := json.Unmarshal(eResource.Data, &subscriptionSpec); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshall subscriptionSpec error: %v", err)
+			}
+
+			utils.AddAnnotation(r.storageClient, utils.DesiredSubscriptionChannelAnnotationKey, subscriptionSpec.Channel)
```
leelavg, line 244: this could be moved into a conditional, iirc `utils.AddAnnotation` return `true` if in-memory object is updated.

rchikatw: Done

leelavg:
If reporter isn't going to add annotation then that code should be removed, was that raised in a separate PR.

rchikatw:
There will be one more pr: which will do 2 things
1. Change in storage client controller to set the status 
2. Reporter will set the annotation using CreateOrUpdate

internal/controller/storageclient_controller.go
```diff
@@ -235,6 +235,17 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			var subscriptionSpec *opv1a1.SubscriptionSpec
+			if err := json.Unmarshal(eResource.Data, &subscriptionSpec); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshall subscriptionSpec error: %v", err)
+			}
+
+			if utils.AddAnnotation(r.storageClient, utils.DesiredSubscriptionChannelAnnotationKey, subscriptionSpec.Channel) {
+				if err := r.update(r.storageClient); err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with subscription channel annotation: %v", err)
+				}
+			}
```
nb-ohad, line 248: This entire change does not make sense. What we want ti achieve is the provider sending the desired subscription and the client code takes that and update the exiting subscription

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
rchikatw: @nb-ohad The existing code is simply adding the annotation on the storage client resource. So I have done the same here. I have checked with @leelavg regarding why we are not setting the channel in the subscription itself he explained to me with a scenario
-> Assume we have a client connected to 2 different providers and one of the providers is upgraded and the other provider is not in that case it's not good to upgrade the client and the status reporter is also not a good place to do all this logic hence it is used to add just annotation.  this logic is part of operatorconfig_controller [here](https://github.com/red-hat-storage/ocs-client-operator/blob/main/internal/controller/operatorconfigmap_controller.go#L218)

nb-ohad: But the subscription you are sending is the OCS subscription, not the client op subscription. I would expect the client-op subscription to be sent and blindly merged with the existing subscription.

rchikatw: Over huddle, its been decided to patch ocs-client-subs blindly and from the provider send the spec of ocs-client-operator.

leelavg: If there is a mismatch among other storageclients then the patch will not go through webhook and the reconciler will not make progress until the patch is successful.

If we don't want to consider more than one storageclient then the whole webhook mechanism doesn't need to be there in the first place.

rchikatw:
Closing this PR as We need to rethink after 4.18 changes.

PR #212, opened 2024-08-20
controllers: configure log rotater spec in csi-op operatorconfig

leelavg:
log rotation feature in csi-op provides a way to rotate csi logs based on custom configuration and this PR configures the default agreed upon values.

### DIFF
--- a/pkg/templates/csi.go
+++ b/pkg/templates/csi.go
@@ -7,6 +7,7 @@ import (
 	secv1 "github.com/openshift/api/security/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/ptr"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/templates/csi.go
```diff
@@ -73,17 +74,42 @@ const CSIOperatorConfigName = "ceph-csi-operator-config"
 
 var CSIOperatorConfigSpec = csiopv1a1.OperatorConfigSpec{
 	DriverSpecDefaults: &csiopv1a1.DriverSpec{
+		Log: &csiopv1a1.LogSpec{
+			Rotation: &csiopv1a1.LogRotationSpec{
+				Periodicity: csiopv1a1.DailyPeriod,
+				MaxLogSize:  resource.MustParse("500M"),
+				MaxFiles:    7,
```
parth-gr, line 81: do we define this as a constants values?

parth-gr (commented):
lgtm

Try running the changes on openshift cluster, and validate that things are working fine, plus see the pod yaml and validate that csi pod actually has the logrotate container added

pkg/templates/csi.go
leelavg: I have no preference of one over other however I have my reservations on when to use const, mostly for invariants across many releases, env var, cross-pkg knowledge etc and I don't deem above should be a const.

nb-ohad:
LGTM
@leelavg If you have no other reservation please merge

leelavg:
I'm still not able to test it, cluster availability, some other work etc. Will try to close it before EOD.

leelavg:
> Try running the changes on openshift cluster, and validate that things are working fine, plus see the pod yaml and validate that csi pod actually has the logrotate container added
- verified.

PR #217, opened 2024-08-29
controllers: trigger reconcile event to storage client if hash is not matched

rchikatw:
- Save hash to storage client status
- add an annotation for the subscription channel to the storage client
- Add a new field to the status of the storage client
- Testing is pending will add results once it's done
- I have messed PR (https://github.com/red-hat-storage/ocs-client-operator/pull/200) which has similar changes and due to which its closed automatically hence creating the current pr.


service/status-report/main.go
```diff
@@ -119,7 +119,7 @@ func main() {
 
 	storageClientCopy := &v1alpha1.StorageClient{}
 	storageClient.DeepCopyInto(storageClientCopy)
-	if utils.AddAnnotation(storageClient, utils.DesiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
+	if utils.AddAnnotation(storageClient, utils.DesiredConfigHashAnnotationKey, statusResponse.DesiredConfigHash) {
```
rchikatw, line 122: Not using `CreateOrUpdate` because utils.AddAnnotations works what we need in the current situation.

internal/controller/storageclient_controller.go
```diff
@@ -212,6 +212,11 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 		return res, err
 	}
 
+	if r.storageClient.Status.ConsumerConfig.Hash != storageClientResponse.DesiredConfigHash {
+		r.storageClient.Status.ConsumerConfig.Hash = storageClientResponse.DesiredConfigHash
+		r.storageClient.Status.ConsumerConfig.Timestamp = metav1.Now()
+	}
```
leelavg, line 218: shouldn't this preferably be at the end of reconcile? In current logic and updated hash isn't suggesting a successful reconcile, isn't it?

rchikatw: I can move it to end not a problem

service/status-report/main.go
```diff
@@ -119,7 +119,17 @@ func main() {
...
 	if utils.AddAnnotation(storageClient, utils.DesiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
+		updated = true
+	}
+
+	if statusResponse.DesiredConfigHash != storageClient.Status.ConsumerConfig.Hash &&
+		utils.AddAnnotation(storageClient, utils.DesiredConfigHashAnnotationKey, statusResponse.DesiredConfigHash) {
+		updated = true
+	}
+
+	if updated {
```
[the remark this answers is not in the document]
rchikatw: We have decided we won’t be touching subscriptions part ok. And if the hash changes we want to trigger the reconcile to storage client. I am patching annotation for config hash same way how it’s done for subscription.

internal/controller/storageclient_controller.go
rchikatw: Done.

leelavg:
LGTM but will wait for Ohad's approval.

nb-ohad:
/hold 

@rchikatw Something here seems to be in misalignment with the design. 
I can see 3 different values taking into consideration
1. The annotation
2. The value on the status
3. The value coming from the provider

One of them is unnecessary and redundant, I will vote out the status value.

rchikatw:
> /hold
> 
> @rchikatw Something here seems to be in misalignment with the design. I can see 3 different values taking into consideration
> 
> 1. The annotation
> 2. The value on the status
> 3. The value coming from the provider
> 
> One of them is unnecessary and redundant, I will vote out the status value.

If we decided to remove the status section from storage client then there is no point in keeping the root field at the storage config response [here](https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/proto/provider.proto#L87)

pkg/utils/k8sutils.go
```diff
@@ -41,6 +41,9 @@ const StatusReporterImageEnvVar = "STATUS_REPORTER_IMAGE"
 // Value corresponding to annotation key has subscription channel
 const DesiredSubscriptionChannelAnnotationKey = "ocs.openshift.io/subscription.channel"
 
+// Value corresponding to annotation key has desired client hash
+const DesiredConfigHashAnnotationKey = "ocs.openshift.io/desired.config.hash"
```
nb-ohad, line 45: 1. This is the name of an annotation on the CR so by default it represents the desired state. There is no reason to mention desired in the name of the annotation. 
2. The encoding of the data `hash` might change in the future, I don't think we need to mention that as part of the name as it is an impl detail.
3. The annotation name does not describe what kind of function this annotation represents

service/status-report/main.go
```diff
@@ -119,7 +119,8 @@ func main() {
 
 	storageClientCopy := &v1alpha1.StorageClient{}
 	storageClient.DeepCopyInto(storageClientCopy)
-	if utils.AddAnnotation(storageClient, utils.DesiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) {
+	if utils.AddAnnotation(storageClient, utils.DesiredSubscriptionChannelAnnotationKey, statusResponse.DesiredClientOperatorChannel) ||
+		utils.AddAnnotation(storageClient, utils.DesiredConfigHashAnnotationKey, statusResponse.DesiredConfigHash) {
 		// patch is being used here as to not have any conflicts over storageclient cr changes as this annotation value doesn't depend on storageclient spec
 		if err := cl.Patch(ctx, storageClient, client.MergeFrom(storageClientCopy)); err != nil {
```
nb-ohad, line 125: Are we sure that patch is the correct approach for this new annotation? What happens if a reconcile is in motion while you are trying to read and update the annotation. It might be that be the time you get to the patch command the CR already changed

nb-ohad (changes requested):

nb-ohad:
@rchikatw Why do the changes in this PR require changes in hundreds of vendor files?

service/status-report/main.go
rchikatw: I'm not sure which method I should use in this case. If I use "update" on CR, it sends the entire object to the API server, while "patch" sends only the changes that need to be made. There is a greater risk of conflict when using "update" and a lesser risk when using "patch," but in both cases, we can still encounter issues.
Even if I use PartialMetadata there will be same issue I feel

pkg/utils/k8sutils.go
rchikatw: how about changing it to `consumer.config.state`

leelavg:
> @rchikatw Why do the changes in this PR require changes in hundreds of vendor files?  
- should be solved by https://github.com/red-hat-storage/ocs-operator/pull/2798

rchikatw:
> > @rchikatw Why do the changes in this PR require changes in hundreds of vendor files?
> 
> * should be solved by [provider: split provider related pkgs into 3 go modules  ocs-operator#2798](https://github.com/red-hat-storage/ocs-operator/pull/2798)

Sure once this is merged I will rebase it.

service/status-report/main.go
nb-ohad: We want the conflict because we do not want to update based on stale data. Let's use update instead of paths for this new annotation. The old annotation should continue to use patch

pkg/utils/k8sutils.go
nb-ohad: What about `ocs.openshift.io/provider-side-state`? or something like that?

rchikatw: just `ocs.openshift.io/provider-state` is good enough i guess

nb-ohad: `provider-state` implies this represents the entire provider state which is wrong. 
`provider-side-state` implies this represent some state on the provider side

service/status-report/main.go
rchikatw: I will use `.Update` because it will consider the resource version while updating the resource.

leelavg:
@rchikatw new proto changes aren't being pulled?

rchikatw:
> @rchikatw new proto changes aren't being pulled?

It's pulled now. It took some time to push that change; meanwhile, you saw my PR.

leelavg (approved):

leelavg:
/unhold

From Ohad
```
The PR is so big I am unable to access the git without my tab crashing

@lgangava Can you please remove the hold on my behalf?
```

PR #227, opened 2024-09-06
controller: update storageclient connection fail log

rohan47:
updated storageclient connection fail log to show provider endpoint as well

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -367,7 +367,7 @@ func (r *StorageClientReconciler) newExternalClusterClient() (*providerClient.OC
 	ocsProviderClient, err := providerClient.NewProviderClient(
 		r.ctx, r.storageClient.Spec.StorageProviderEndpoint, time.Second*10)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create a new provider client: %v", err)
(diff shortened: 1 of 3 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -367,7 +367,7 @@ func (r *StorageClientReconciler) newExternalClusterClient() (*providerClient.OC
 	ocsProviderClient, err := providerClient.NewProviderClient(
 		r.ctx, r.storageClient.Spec.StorageProviderEndpoint, time.Second*10)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create a new provider client: %v", err)
+		return nil, fmt.Errorf("failed to create a new provider client while connecting to %v: %v", r.storageClient.Spec.StorageProviderEndpoint, err)
```
leelavg, line 370: we need this at all places where we are creating new client, like, in claim controller and in heartbeat as well.

rohan47: Updated

leelavg: Did you check other places as well? I mentioned what I remembered.

rohan47: I checked where `NewProviderClient` is created

leelavg (approved):

PR #232, opened 2024-09-16
use provider/api module as the dependency in place of whole ocs-operator

leelavg:
Based on https://github.com/red-hat-storage/ocs-operator/pull/2798 ocs-operator now offers services/provider/api as a standalone importable module without depending on ocs-operator and this helps in easier management of client-op deps by not always going after ocs-operator and it's own dependencies.


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/red-hat-storage/ocs-client-operator
 
-go 1.22.0
-
-toolchain go1.22.3
+go 1.22.5
```
Madhu-1, line 3: Can you please check we have this version in downstream if you are planning for backport to 4.17

leelavg: I'm not thinking of backport, nevertheless d/s does support this version.

Madhu-1 (commented):
/lgtm
/approve

PR #241, opened 2024-09-24
controllers: Fix secret creation in storageclient controller

ezio-auditore:
Introduced in https://github.com/red-hat-storage/ocs-client-operator/pull/176 the storage client controller is trying to create a secret with non existent storage client namespace name. This patch fixes the above by providing operator namespace instead.

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -247,7 +247,7 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 			}
 			secret := &corev1.Secret{}
 			secret.Name = eResource.Name
-			secret.Namespace = r.storageClient.Namespace
+			secret.Namespace = r.OperatorNamespace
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -247,7 +247,7 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 			}
 			secret := &corev1.Secret{}
 			secret.Name = eResource.Name
-			secret.Namespace = r.storageClient.Namespace
+			secret.Namespace = r.OperatorNamespace
```
leelavg, line 250: w/ a simple search I could see this in more than 1 place, pls check and fix https://github.com/search?q=repo%3Ared-hat-storage%2Focs-client-operator%20r.storageClient.Namespace&type=code

ezio-auditore: Updated accordingly

Madhu-1:
/lgtm


/assign @leelavg 

for approval

PR #249, opened 2024-10-01
controllers: add option to disable key rotation

black-dragon74:
This PR allows an user to annotate the storageclaim with `keyrotation.csiaddons.openshift.io/enable=false` to disable key rotation for the RBD storageclasses.

### DIFF
--- a/internal/controller/storageclaim_controller.go
+++ b/internal/controller/storageclaim_controller.go
@@ -50,9 +50,10 @@ import (
 )
 
 const (
-	storageClaimFinalizer  = "storageclaim.ocs.openshift.io"
-	storageClaimAnnotation = "ocs.openshift.io/storageclaim"
-	keyRotationAnnotation  = "keyrotation.csiaddons.openshift.io/schedule"
+	storageClaimFinalizer       = "storageclaim.ocs.openshift.io"
+	storageClaimAnnotation      = "ocs.openshift.io/storageclaim"
+	keyRotationAnnotation       = "keyrotation.csiaddons.openshift.io/schedule"
+	keyRotationEnableAnnotation = "keyrotation.csiaddons.openshift.io/enable"
 
 	pvClusterIDIndexName  = "index:persistentVolumeClusterID"
 	vscClusterIDIndexName = "index:volumeSnapshotContentCSIDriver"
@@ -397,6 +398,10 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					storageClass = r.getCephFSStorageClass(data)
 				} else if resource.Name == "ceph-rbd" {
 					storageClass = r.getCephRBDStorageClass(data)
+
+					if enable, ok := r.storageClaim.GetAnnotations()[keyRotationEnableAnnotation]; ok && enable != "true" {
+						utils.AddAnnotation(storageClass, keyRotationEnableAnnotation, "false")
+					}
 				}
 				utils.AddAnnotation(storageClass, storageClaimAnnotation, r.storageClaim.Name)
 				err = r.createOrReplaceStorageClass(storageClass)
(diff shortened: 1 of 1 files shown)


internal/controller/storageclaim_controller.go
```diff
@@ -397,6 +398,10 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					storageClass = r.getCephFSStorageClass(data)
 				} else if resource.Name == "ceph-rbd" {
 					storageClass = r.getCephRBDStorageClass(data)
+
+					if enable, ok := r.storageClaim.GetAnnotations()[keyRotationEnableAnnotation]; ok && enable != "true" {
```
leelavg, line 402: does this ann. value supports anything other than `true/false`? why to check for a negation, like why not `ann[key] == "false" ? set-false`

if the value is "true"/"false" then it can be directly used, isn't it?
`utils.AddAnnotation(storageClass, keyRotationEnableAnnotation, r.storageClaim.GetAnnotations()[keyRotationEnableAnnotation])`

black-dragon74: No it does not support any other value than true/false, values other than `true` are considered false. I have no good reason on why it cannot be used directly (as it will be parsed in csi-addons anyways)

The reason I did not do it such way is to avoid setting it unconditionally because in `utils.AddAnnotation` we set the `value` even if it is empty. Want me to change it?

Also, while we are at it, should I update the code to update the SC by patching it instead of deleting and recreating in case the annotations differ? Would be a little efficient IG?

leelavg: > Want me to change it?
- No, annotation w/ an empty value is also a valid identifier, let's not change it

> Also, while we are at it, should I update the code to update the SC by patching it instead of deleting and recreating in case the annotations differ? Would be a little efficient IG?
- these values converge easily in a couple of reconciles and a patch here makes less difference. 
- Nevertheless will revisit when VolumeAttributes (Madhu mentioned it once) feature which helps in storageclass having almost static data.

black-dragon74: > * No, annotation w/ an empty value is also a valid identifier, let's not change it
> 
> >

I meant changing the current pr aka my change-set to add annotation unconditionally :)

I am still leaning towards adding the annotation only when its value is not truthy.

black-dragon74: > * these values converge easily in a couple of reconciles and a patch here makes less difference.

Would you please explain it to me? Acc to my understanding we would be deleting and recreating the SC in case annotations differ. That is two different actions while patching the annotations on an existing SC will not disrupt any operations and is one shot.

leelavg: > in case annotations differ
- yes, I mean these doesn't change often, isn't it.

nb-ohad:
@black-dragon74 
To the best of my understanding, Key rotation controls are storage admin controls. StoargeClaim is not a storage admin-owned CR. It is a CR intended to be used by the application owner and I believe the key rotation control should not be managed by the application owner. 

In addition, we are phasing out the StorageClaim CR as a concept, so we should not change/add any new nob on that CR

iPraveenParihar:
> @black-dragon74 To the best of my understanding, Key rotation controls are storage admin controls. StoargeClaim is not a storage admin-owned CR. It is a CR intended to be used by the application owner and I believe the key rotation control should not be managed by the application owner.
> 
> In addition, we are phasing out the StorageClaim CR as a concept, so we should not change/add any new nob on that CR

@nb-ohad, Initially, for the ReclaimSpace disable option we planned to implement on the StorageRequest CR, but IIRC from our last meeting, we agreed to use annotations on the StorageClaim CR for now because we don't have mechanism to pass annotations from the StorageRequest CR to the StorageClass created in the client cluster. 
And same approach was decided to be followed for key rotation disable option.

cc @Madhu-1

internal/controller/storageclaim_controller.go
black-dragon74: Correct.

PR #254, opened 2024-10-09
controllers: send storage utilization ratio in heartbeat

rchikatw:
-  send storage utilization ratio in a heartbeat
- This will be calculated from a CRQ resource 
- This field will be used to show the storage utilization of the client in the provider mode dashboard
- go.mod file changes was added by taking this PR into account: https://github.com/red-hat-storage/ocs-operator/pull/2697


rchikatw:
@leelavg @nb-ohad please review.

service/status-report/main.go
```diff
@@ -215,6 +222,25 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
 
 }
 
+func setStorageQuotaUtilizationRatio(ctx context.Context, cl client.Client, status interfaces.StorageClientStatus) {
+	clusterResourceQuota := &quotav1.ClusterResourceQuota{}
+	clusterResourceQuota.Name = status.GetClientName()
+
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(clusterResourceQuota), clusterResourceQuota); err != nil {
+		// No need to check for NotFound because unlimited quota client will not have CRQ resource
+		if !errors.IsNotFound(err) {
```
leelavg, line 231:
```suggestion
	if err := cl.Get(ctx, client.ObjectKeyFromObject(clusterResourceQuota), clusterResourceQuota); client.IgnoreNotFound(err) != nil {
```

service/status-report/main.go
```diff
@@ -215,6 +222,25 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
+			klog.Warningf("Failed to get clusterResourceQuota %q: %v", clusterResourceQuota.Name, err)
+		}
+	}
+	if clusterResourceQuota.Status.Total.Hard != nil {
+		var total = clusterResourceQuota.Status.Total.Hard["requests.storage"]
+		var used = clusterResourceQuota.Status.Total.Used["requests.storage"]
+		var ratio = used.AsApproximateFloat64() / total.AsApproximateFloat64()
```
leelavg, line 238: what was the issue w/ using the walrus op `:=`?

service/status-report/main.go
```diff
@@ -215,6 +222,25 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
+	clusterResourceQuota.Name = status.GetClientName()
+
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(clusterResourceQuota), clusterResourceQuota); err != nil {
+		// No need to check for NotFound because unlimited quota client will not have CRQ resource
+		if !errors.IsNotFound(err) {
+			klog.Warningf("Failed to get clusterResourceQuota %q: %v", clusterResourceQuota.Name, err)
+		}
+	}
+	if clusterResourceQuota.Status.Total.Hard != nil {
+		var total = clusterResourceQuota.Status.Total.Hard["requests.storage"]
```
leelavg, line 236: total will never be zero, isn't it?

rchikatw: it can be zero incase of unlimited quota.

service/status-report/main.go
rchikatw: no issue, while typing i missed i guess.

service/status-report/main.go
leelavg: I mean in this block, if it's zero it might panic w/ dividebyzero isn't it?

vendor/github.com/red-hat-storage/ocs-operator/services/provider/api/v4/interfaces/interfaces.go
```diff
@@ -9,12 +9,14 @@ type StorageClientStatus interface {
 	GetClusterID() string
 	GetClusterName() string
 	GetClientName() string
+	GetStorageQuotaUtilizationRatio() float32
```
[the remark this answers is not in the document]
rchikatw: Okay, I will work on it in a separate pull request.

service/status-report/main.go
rchikatw: no issue by typing end up writing this. will change.

service/status-report/main.go
rchikatw: ohh yeah, I got your point. will handle that condition

service/status-report/main.go
leelavg: Pls add the comment, change I suggested is only to make a point that it can be condensed.

leelavg:
LGTM. will wait for other review.

service/status-report/main.go
```diff
@@ -215,6 +221,29 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
+	}
+
+	if clusterResourceQuota.Status.Total.Hard != nil {
+		total := clusterResourceQuota.Status.Total.Hard["requests.storage"]
+		used := clusterResourceQuota.Status.Total.Used["requests.storage"]
+		ratio := used.AsApproximateFloat64() / total.AsApproximateFloat64()
+		if ratio > 1 {
+			status.SetStorageQuotaUtilizationRatio(1)
+		} else {
+			status.SetStorageQuotaUtilizationRatio(float32(ratio))
```
rohan47, line 240: are we going to show only ratio? and not capacity utilized

nb-ohad: @rohan47 Because this is not about ODF capacity utilization, the quota reports the number of bytes out of the quota for total cluster storage utilization. I made the call to use ratio because we are talking about a number that describes how close we are to the quota, I agree that the world utilization here is problematic, but we couldn't find better terminology. If you have a suggestion for a more fitting terminology, please reply

service/status-report/main.go
```diff
@@ -215,6 +221,29 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
+func setStorageQuotaUtilizationRatio(ctx context.Context, cl client.Client, status interfaces.StorageClientStatus) {
+	clusterResourceQuota := &quotav1.ClusterResourceQuota{}
+	clusterResourceQuota.Name = status.GetClientName()
+
+	// No need to check for NotFound because unlimited quota client will not have CRQ resource
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(clusterResourceQuota), clusterResourceQuota); client.IgnoreNotFound(err) != nil {
+		klog.Warningf("Failed to get clusterResourceQuota %q: %v", clusterResourceQuota.Name, err)
+	}
+
+	if clusterResourceQuota.Status.Total.Hard != nil {
```
nb-ohad, line 233: Are we sure none of parts of `.Status.Total.Hard` is a pointer? If any is you will need a `nil` pointer check

service/status-report/main.go
```diff
@@ -215,6 +221,29 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
+		total := clusterResourceQuota.Status.Total.Hard["requests.storage"]
+		used := clusterResourceQuota.Status.Total.Used["requests.storage"]
+		ratio := used.AsApproximateFloat64() / total.AsApproximateFloat64()
+		if ratio > 1 {
+			status.SetStorageQuotaUtilizationRatio(1)
+		} else {
+			status.SetStorageQuotaUtilizationRatio(float32(ratio))
+		}
```
nb-ohad, line 241: 1. What if total is 0, you cannot divide by 0
2. Why an `if ... else ...`? it makes more sense using `math.Min` to get the min between the ratio and 1

nb-ohad (changes requested):

service/status-report/main.go
rchikatw: Yes, I have verified with multiple scenario.

service/status-report/main.go
rchikatw: 1. Total will never be zero here because we are creating a resource, and we will always have a value greater than 0. We can't generate a token with a value of 0 because that validation is added as part of the token generation API. I have now included the validation for that edge case(If the client admin created a CRQ resource with a quota on a different resource.).
2. math.Min is not available for float32 I have to do lot of to and from conversion like this
`status.SetStorageQuotaUtilizationRatio(float32(math.Min(float64(ratio), float64(1.0))))`

rchikatw: Done.

service/status-report/main.go
nb-ohad: There is no need to look at scenarios, just jump the type definition of `Status` and see if `Status` or `Total` are defined as pointer/interfaces or not

service/status-report/main.go
```diff
@@ -215,6 +222,26 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
...
+
+	// No need to check for NotFound because unlimited quota client will not have CRQ resource
+	if err := cl.Get(ctx, client.ObjectKeyFromObject(clusterResourceQuota), clusterResourceQuota); client.IgnoreNotFound(err) != nil {
+		klog.Warningf("Failed to get clusterResourceQuota %q: %v", clusterResourceQuota.Name, err)
+	}
+
+	if clusterResourceQuota.Status.Total.Hard != nil {
+		total := clusterResourceQuota.Status.Total.Hard["requests.storage"]
+		if total.AsApproximateFloat64() > 0 {
+			used := clusterResourceQuota.Status.Total.Used["requests.storage"]
```
nb-ohad, line 237: Hard and Used are of type `map[string]corev1.ResourceQuantity` this means that if the key `requests.storage` is not defined in either of the maps you will get a default `corev1.ResourceQuantity` which points to a value of 0. But that does not mean the utilization is 0, it just means that no controller has set this value. 

```suggestion
		total, totalExists := clusterResourceQuota.Status.Total.Hard["requests.storage"]
		used, usedExists := clusterResourceQuota.Status.Total.Used["requests.storage"]
		if totalExists && usedExists && total.AsApproximateFloat64() > 0 {
```

service/status-report/main.go
rchikatw: Done.

PR #264, opened 2024-11-04
controllers: new controller for maintenance mode

rewantsoni:
Depends on https://github.com/red-hat-storage/ocs-operator/pull/2870

A controller which will watch the StorageClient and MaintenanceMode CR by ramen to send a RequestMaintenaneMode RPC call to provider and set the status from response received from GetStorageConfig

### DIFF
--- a/api/v1alpha1/storageclient_types.go
+++ b/api/v1alpha1/storageclient_types.go
@@ -50,6 +50,8 @@ type StorageClientSpec struct {
 type StorageClientStatus struct {
 	Phase storageClientPhase `json:"phase,omitempty"`
 
+	InMaintenanceMode bool `json:"inMaintenanceMode"`
+
 	// ConsumerID will hold the identity of this cluster inside the attached provider cluster
 	ConsumerID string `json:"id,omitempty"`
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -154,7 +154,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	_, err = getAvailableCRDNames(context.Background(), apiClient)
+	availCrds, err := getAvailableCRDNames(context.Background(), apiClient)
 	if err != nil {
 		setupLog.Error(err, "Unable get a list of available CRD names")
 		os.Exit(1)
@@ -204,9 +204,20 @@ func main() {
 		Scheme:            mgr.GetScheme(),
 		OperatorNamespace: utils.GetOperatorNamespace(),
 		ConsolePort:       int32(consolePort),
+		AvailableCrds:     availCrds,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "OperatorConfigMapReconciler")
 		os.Exit(1)
+	}
+
+	if availCrds[controller.MaintenanceModeCRDName] {
+		if err = (&controller.MaintenanceModeReconciler{
+			Client: mgr.GetClient(),
+			Scheme: mgr.GetScheme(),
+		}).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "MaintenanceMode")
+			os.Exit(1)
+		}
 	}
 
 	setupLog.Info("starting manager")
--- a/internal/controller/maintenancemode_controller.go
+++ b/internal/controller/maintenancemode_controller.go
@@ -1,0 +1,192 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-logr/logr"
+	ramenv1alpha1 "github.com/ramendr/ramen/api/v1alpha1"
+	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+	providerclient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"slices"
+)
+
+const (
+	MaintenanceModeCRDName = "maintenancemodes.ramendr.openshift.io"
+)
+
+// MaintenanceModeReconciler reconciles a ClusterVersion object
+type MaintenanceModeReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+
+	log logr.Logger
+	ctx context.Context
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	generationChangePredicate := predicate.GenerationChangedPredicate{}
+	maintenanceModeChangedPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*v1alpha1.StorageClient)
+			newObj := e.ObjectNew.(*v1alpha1.StorageClient)
+			return oldObj.Status.InMaintenanceMode != newObj.Status.InMaintenanceMode
+		},
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("MaintenanceMode").
+		Watches(
+			&ramenv1alpha1.MaintenanceMode{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(generationChangePredicate),
+		).
+		Watches(
+			&v1alpha1.StorageClaim{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(generationChangePredicate),
+		).
+		Watches(
+			&v1alpha1.StorageClient{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				generationChangePredicate,
+				maintenanceModeChangedPredicate,
+			),
+		).
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients;storageclaims,verbs=get;list;watch
+
+func (r *MaintenanceModeReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	r.ctx = ctx
+	r.log = log.FromContext(ctx)
+	r.log.Info("Starting reconcile")
+
+	nameToStorageClient := map[string]*v1alpha1.StorageClient{}
+
+	maintenanceModes := &ramenv1alpha1.MaintenanceModeList{}
+	if err := r.list(maintenanceModes); err != nil {
+		r.log.Error(err, "failed to list the MaintenanceMode CRs")
+		return reconcile.Result{}, err
+	}
+
+	for i := range maintenanceModes.Items {
+		mm := &maintenanceModes.Items[i]
+		sc := &v1alpha1.StorageClaim{}
+		// MMode's TargetID is replicationID, which in our case is storageClaim name
+		sc.Name = mm.Spec.TargetID
+		if err := r.get(sc); err != nil {
+			return ctrl.Result{}, err
+		}
+		clientName := sc.Spec.StorageClient
+		if clientName == "" {
+			return ctrl.Result{}, fmt.Errorf("StorageClaim %s does not have a StorageClient defined", sc.Name)
+		}
+		if nameToStorageClient[clientName] == nil {
+			storageClient := &v1alpha1.StorageClient{}
+			storageClient.Name = clientName
+			if err := r.get(storageClient); err != nil {
+				return ctrl.Result{}, err
+			}
+			nameToStorageClient[clientName] = storageClient
+		}
+		if nameToStorageClient[clientName].Status.InMaintenanceMode {
+			if err := r.updateStatusCompletedForMM(mm); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to update status for MaintenanceMode %s: %w", mm.Name, err)
+			}
+		}
+	}
+
+	storageClients := &v1alpha1.StorageClientList{}
+	if err := r.list(storageClients); err != nil {
+		r.log.Error(err, "failed to list the Storage Clients")
+		return reconcile.Result{}, err
+	}
+
+	for i := range storageClients.Items {
+		storageClient := &storageClients.Items[i]
+		_, needsMaintenanceMode := nameToStorageClient[storageClient.Name]
+		if needsMaintenanceMode != storageClient.Status.InMaintenanceMode {
+			if err := r.toggleMaintenanceModeForClient(storageClient, needsMaintenanceMode); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *MaintenanceModeReconciler) toggleMaintenanceModeForClient(storageClient *v1alpha1.StorageClient, enable bool) error {
+	providerClient, err := providerclient.NewProviderClient(
+		r.ctx,
+		storageClient.Spec.StorageProviderEndpoint,
+		utils.OcsClientTimeout,
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"failed to create provider client with endpoint %v: %v",
+			storageClient.Spec.StorageProviderEndpoint,
+			err,
+		)
+	}
+	// Close client-side connections.
+	defer providerClient.Close()
+
+	_, err = providerClient.RequestMaintenanceMode(r.ctx, storageClient.Status.ConsumerID, enable)
+	if err != nil {
+		return fmt.Errorf("failed to Request maintenance mode: %v", err)
+	}
+	return nil
+}
+
+func (r *MaintenanceModeReconciler) updateStatusCompletedForMM(maintenanceMode *ramenv1alpha1.MaintenanceMode) error {
+	// Ramen reads the State and Conditions in order to determine that the MaintenanceMode is Completed
+
+	condition := metav1.Condition{
+		Type:               string(ramenv1alpha1.MModeConditionFailoverActivated),
+		Status:             metav1.ConditionTrue,
+		Reason:             string(ramenv1alpha1.MModeStateCompleted),
+		ObservedGeneration: maintenanceMode.Generation,
+	}
+
+	updateRequired := false
+	updateRequired = updateRequired || (maintenanceMode.Status.State != ramenv1alpha1.MModeStateCompleted)
+	updateRequired = updateRequired || slices.Contains(maintenanceMode.Status.Conditions, condition)
+
+	if updateRequired {
+		maintenanceMode.Status.State = ramenv1alpha1.MModeStateCompleted
+		maintenanceMode.Status.ObservedGeneration = maintenanceMode.Generation
+		meta.SetStatusCondition(&maintenanceMode.Status.Conditions, condition)
+
+		if err := r.Client.Status().Update(r.ctx, maintenanceMode); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *MaintenanceModeReconciler) list(obj client.ObjectList, opts ...client.ListOption) error {
+	return r.List(r.ctx, obj, opts...)
+}
+
+func (r *MaintenanceModeReconciler) get(obj client.Object, opts ...client.GetOption) error {
+	return r.Get(r.ctx, client.ObjectKeyFromObject(obj), obj, opts...)
+}
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -80,6 +80,7 @@ type OperatorConfigMapReconciler struct {
 	OperatorNamespace string
 	ConsolePort       int32
 	Scheme            *runtime.Scheme
+	AvailableCrds     map[string]bool
 
 	log                 logr.Logger
 	ctx                 context.Context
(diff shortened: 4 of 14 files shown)


rewantsoni:
review only the last commit

api/v1alpha1/storageclient_types.go
```diff
@@ -46,10 +46,16 @@ type StorageClientSpec struct {
...
 
+type ProviderAttributes struct {
+	InMaintenanceMode bool `json:"inMaintenanceMode"`
+}
+
 // StorageClientStatus defines the observed state of StorageClient
 type StorageClientStatus struct {
 	Phase storageClientPhase `json:"phase,omitempty"`
 
+	Provider ProviderAttributes `json:"providerAttributes,omitempty"`
```
nb-ohad, line 57: 1. I don't see why we need a sub struct here. 
2. I am not sure we want to use `provider` as terminology in our APIs

cmd/main.go
```diff
@@ -208,6 +209,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	if availCrds[controller.MaintenanceModeCRDName] {
```
nb-ohad, line 212: I'm not sure why we need to involve available CRSs here. We don't have conditional watches at this point just a single decision on should we run the controller or not

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	if len(storageClassList.Items) != 1 {
+		return ctrl.Result{}, fmt.Errorf("failed to find storageClass for maintenance mode")
+	}
+
+	storageClass := &storageClassList.Items[0]
+
+	// find the storageClient
+	// TODO: when storageClass is owned by storageClient, move to find the Client by ownerRef
+	storageClient := &v1alpha1.StorageClient{}
+	storageClient.Name = storageClass.GetLabels()[storageClientLabel]
```
nb-ohad, line 92: Please use the owner ref, i am not sure about if this label even exists anymore and we have no gurentees we will keep it in the future

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	storageClassList := &storagev1.StorageClassList{}
+	err := r.list(storageClassList, client.MatchingLabels{ramenReplicationIDLabel: maintenanceMode.Spec.TargetID})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if len(storageClassList.Items) != 1 {
+		return ctrl.Result{}, fmt.Errorf("failed to find storageClass for maintenance mode")
+	}
+
+	storageClass := &storageClassList.Items[0]
```
nb-ohad, line 87: Why the first storage class? how do you know that the first one belongs to the/a proper client?

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if reconcileErr != nil {
+		err = reconcileErr
+	} else if statusErr != nil {
+		err = statusErr
+	}
+	return result, err
+}
+
+func (r *MaintenanceModeReconciler) reconcileStates(maintenanceMode *ramenv1alpha1.MaintenanceMode, storageClient *v1alpha1.StorageClient) (ctrl.Result, error) {
```
nb-ohad, line 113: Our terminology is usually reconcilePhases. Can you please align here?

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if !maintenanceMode.GetDeletionTimestamp().IsZero() {
```
nb-ohad, line 115: For readability please handle deletion after normal reconciliation

nb-ohad (changes requested):

internal/controller/maintenancemode_controller.go
rewantsoni: We are finding out the storageClass via the replicationID, which is uniqueID added to the storageClass required for ramen. There will be only one such storageClass, I have a check just above to fail if we find multiple storageClasses with the same replicationID

cmd/main.go
rewantsoni: If we don't have the CRD for maintenanceMode created, the operator pod won't start as we have a new controller For MaintenanceMode.
If the CRD is created at a later point in time, we need to restart the process for ocs-client-op so that we start the maintenanceMode controller

api/v1alpha1/storageclient_types.go
rewantsoni: Currently, we are sending one piece of information from the provider, in the future we might want multiple information. So, I think it would be good to have it under a substruct.

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,202 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	// ensure finalizer
+	if controllerutil.AddFinalizer(maintenanceMode, maintenanceModeFinalizer) {
+		r.log.Info("finalizer missing on the Maintenance Mode resource, adding...")
+		if err := r.Client.Update(r.ctx, maintenanceMode); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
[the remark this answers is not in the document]
rewantsoni: To Stop the maintenanceMode, we need to send the `ReqMaintenanceMode(false)` to the provider and make sure the MaintenanceMode is stopped and the cephRBDMirror is back again

internal/controller/maintenancemode_controller.go
rewantsoni: There is no onwerRef for storageClient on the StorageClasses

cmd/main.go
rewantsoni: Yes, right we have a availCrds loaded at this position and we are reusing it

internal/controller/maintenancemode_controller.go
rewantsoni: Yes, I have added a TODO for it, and will open an issue for it when we move to direct ownership of storageClass by the clients

internal/controller/maintenancemode_controller.go
nb-ohad: I am not sure replicationID is unique to a storage class. Who selects it and sets it up?

internal/controller/maintenancemode_controller.go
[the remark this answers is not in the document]
nb-ohad: That does not answer my question. Who is the one issuing the delete operation? if it is this controller then there is no point in having a finalizer

internal/controller/maintenancemode_controller.go
rewantsoni: Both replicationID and storageID are unique and they are set from the provider server.
replicationID is the hash of storageClass name (so that it is same across cluster, this is a requirement from ramen) and storageID is also unique for each storageclass

cmd/main.go
nb-ohad: Ok, in that case, you might have a different issue. what happens if the crd is not avialable at operator startup but becomes available later on? your controller is not running.

Do we have a mechanism in place to restart the operator in such cases to allow the new controller to come up?

internal/controller/maintenancemode_controller.go
nb-ohad: ok, resolving

cmd/main.go
rewantsoni: Yes, we implemented it with https://github.com/red-hat-storage/ocs-client-operator/pull/253

internal/controller/maintenancemode_controller.go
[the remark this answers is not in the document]
nb-ohad: Ok, resolving

cmd/main.go
nb-ohad: ok, resolving

internal/controller/storageclient_controller.go
```diff
@@ -207,6 +207,8 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 		return reconcile.Result{}, fmt.Errorf("failed to get StorageConfig: %v", err)
 	}
 
+	r.storageClient.Status.InMaintenanceMode = storageClientResponse.SystemAttributes.SystemInMaintenanceMode
```
nb-ohad, line 210: What if `SystemAttribute` is null?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -157,7 +158,15 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&csiopv1a1.OperatorConfig{}, builder.WithPredicates(generationChangePredicate)).
 		Owns(&csiopv1a1.Driver{}, builder.WithPredicates(generationChangePredicate)).
 		Watches(&configv1.ClusterVersion{}, enqueueConfigMapRequest, clusterVersionPredicates).
-		Watches(&extv1.CustomResourceDefinition{}, enqueueConfigMapRequest, builder.OnlyMetadata).
+		Watches(
+			&extv1.CustomResourceDefinition{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				utils.NamePredicate(MaintenanceModeCRDName),
+				utils.CrdCreateAndDeletePredicate(&c.log, MaintenanceModeCRDName, c.AvailableCrds[MaintenanceModeCRDName]),
```
nb-ohad, line 166: We have a generalization (in a util) that allows us to select the events that we would like to select. 
I see no need for this specific predicted or the complexity that comes with it

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*v1alpha1.StorageClient)
+			newObj := e.ObjectNew.(*v1alpha1.StorageClient)
+			return !reflect.DeepEqual(oldObj.Status.InMaintenanceMode, newObj.Status.InMaintenanceMode)
+		},
+	}
+	return ctrl.NewControllerManagedBy(mgr).Named("Maintenance Mode").
```
nb-ohad, line 53: please use proper indentation, and prefer a name that include no whitespaces
```suggestion
	return ctrl.NewControllerManagedBy(mgr)
		.Named("MaintenanceMode").
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+//+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients;storageclaims,verbs=get;list;watch
+
+func (r *MaintenanceModeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.ctx = ctx
+	r.log = log.FromContext(ctx, "MaintenanceMode", req)
+	r.log.Info("Reconciling MaintenanceMode")
+
+	mapClientNameToObj := map[string]*v1alpha1.StorageClient{}
```
nb-ohad, line 81: var names should not be structured as an action (unless it is a pointer to a function). Do not use a verb as the starting work in a var name
```suggestion
	nameToStorageClient := map[string]*v1alpha1.StorageClient{}
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+
+	for i := range maintenanceModes.Items {
+		mm := &maintenanceModes.Items[i]
+		sc := &v1alpha1.StorageClaim{}
+		// MMode's TargetID is replicationID, which in our case is storageClaim name
+		sc.Name = mm.Spec.TargetID
+		err := r.get(sc)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
nb-ohad, line 97: Use nested scopes for errors where possible
```suggestion
		if err := r.get(sc); err != nil {
			return ctrl.Result{}, err
		}
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+		clientName := sc.Spec.StorageClient
+		if mapClientNameToObj[clientName] == nil {
```
nb-ohad, line 99: Validate a non empty client name before reading/writing to the map

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+			storageClient := &v1alpha1.StorageClient{}
+			storageClient.Name = clientName
+			err := r.get(storageClient)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
```
nb-ohad, line 105:
```suggestion
			if err := r.get(storageClient); err != nil {
				return ctrl.Result{}, err
			}
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+			mapClientNameToObj[clientName] = storageClient
+		}
+		if mapClientNameToObj[clientName].Status.InMaintenanceMode {
+			err := r.updateStatusCompletedForMM(mm)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
```
nb-ohad, line 112:
```suggestion
			if err := r.updateStatusCompletedForMM(mm); err != nil {
				return ctrl.Result{}, err
			}
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+			if err := r.sendRequest(storageClient, false); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *MaintenanceModeReconciler) sendRequest(storageClient *v1alpha1.StorageClient, enable bool) error {
```
nb-ohad, line 138: This is a very generic name for a util. Please change to something more specific. Maybe `toogleMaintanceModeForClient`

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+		storageClient := &storageClients.Items[i]
+		if mapClientNameToObj[storageClient.Name] != nil && !storageClient.Status.InMaintenanceMode {
+			if err := r.sendRequest(storageClient, true); err != nil {
+				return ctrl.Result{}, err
+			}
+		} else if mapClientNameToObj[storageClient.Name] == nil && storageClient.Status.InMaintenanceMode {
+			if err := r.sendRequest(storageClient, false); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
nb-ohad, line 132: You don't need to separate cases here. Just check if the desired state (`needsMaintenanceMode`) is different from the actual state (`.Status.InMaintenanceMode`)
```suggestion
		_ , needsMaintenanceMode  = mapClientNameToObl[storageClient.Name]
		if needsMaintenanceMode != storageClient.Status.InMaintenanceMode {
			if err := r.sendRequest(storageClient, clientRequestMM); err != nil {
				return ctrl.Result{}, err
			}
		}
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,188 @@
...
+			Reason:             string(ramenv1alpha1.MModeStateCompleted),
+			Status:             metav1.ConditionTrue,
+		},
+	)
+
+	statusErr := r.Client.Status().Update(r.ctx, maintenanceMode)
+	if statusErr != nil {
+		r.log.Error(statusErr, "Failed to update MaintenanceMode status", "MaintenanceMode", maintenanceMode.Name)
+		return statusErr
+	}
```
nb-ohad, line 178: 1. Simplify
2. Error logging should happen at the upper level
```suggestion
	if err := r.Client.Status().Update(r.ctx, maintenanceMode); err != nil {
		return err
	}
```

nb-ohad (changes requested):

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	generationChangePredicate := predicate.GenerationChangedPredicate{}
+	storageClientStatusPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*v1alpha1.StorageClient)
+			newObj := e.ObjectNew.(*v1alpha1.StorageClient)
+			return !reflect.DeepEqual(oldObj.Status.InMaintenanceMode, newObj.Status.InMaintenanceMode)
```
nb-ohad, line 50: There is no reason to use DeepEqual if the type is a value type and not a composite type.

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	Scheme *runtime.Scheme
+
+	log logr.Logger
+	ctx context.Context
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	generationChangePredicate := predicate.GenerationChangedPredicate{}
+	storageClientStatusPredicate := predicate.Funcs{
```
nb-ohad, line 43: You are not really checking the entire status, just the InMaintenanceMode flag, so the name does not fit.
Let's call this: `maintenanceModeChangedPredicate`

nb-ohad, line 43: Is the default for all other events false or true?

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+		).
+		Watches(
+			&v1alpha1.StorageClaim{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(generationChangePredicate),
+		).
+		Watches(
+			&v1alpha1.StorageClient{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(generationChangePredicate, storageClientStatusPredicate),
```
nb-ohad, line 68: Easier on the eye
```suggestion
			builder.WithPredicates(
				generationChangePredicate, 
				storageClientStatusPredicate,
			),
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients;storageclaims,verbs=get;list;watch
+
+func (r *MaintenanceModeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.ctx = ctx
+	r.log = log.FromContext(ctx, "MaintenanceMode", req)
```
nb-ohad, line 79: Why do you need to add `MaintenanceMode` where the controller name is already set to `MaintenanceMode`?

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	r.log.Info("Reconciling MaintenanceMode")
```
nb-ohad, line 80: No need to repeat the phrase `MaintenanceMode` multiple times on the same log line.
```suggestion
	r.log.Info("Starting reconcile")
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+
+	nameToStorageClient := map[string]*v1alpha1.StorageClient{}
+
+	maintenanceModes := &ramenv1alpha1.MaintenanceModeList{}
+	if err := r.list(maintenanceModes); err != nil {
+		r.log.Error(err, "failed to list the Maintenance Modes")
```
nb-ohad, line 86:
```suggestion
		r.log.Error(err, "failed to list the MaintenanceMode CRs")
```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+		sc := &v1alpha1.StorageClaim{}
+		// MMode's TargetID is replicationID, which in our case is storageClaim name
+		sc.Name = mm.Spec.TargetID
+		if err := r.get(sc); err != nil {
+			return ctrl.Result{}, err
+		}
+		if sc.Spec.StorageClient == "" {
+			return ctrl.Result{}, fmt.Errorf("StorageClaim %s does not have a StorageClient defined", mm.Name)
+		}
+		clientName := sc.Spec.StorageClient
```
nb-ohad, line 101:
```suggestion
		clientName := sc.Spec.StorageClient
		if clientName == "" {
			return ctrl.Result{}, fmt.Errorf("StorageClaim %s does not have a StorageClient defined", mm.Name)
		}

```

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *MaintenanceModeReconciler) toggleMaintenanceModeForClient(storageClient *v1alpha1.StorageClient, enable bool) error {
+	providerClient, err := providerclient.NewProviderClient(
+		r.ctx,
+		storageClient.Spec.StorageProviderEndpoint,
+		10*time.Second,
```
nb-ohad, line 140: why 10? how did we come up with this number?

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,184 @@
...
+	meta.SetStatusCondition(&maintenanceMode.Status.Conditions,
+		metav1.Condition{
+			Type:               string(ramenv1alpha1.MModeConditionFailoverActivated),
+			ObservedGeneration: maintenanceMode.Generation,
+			Reason:             string(ramenv1alpha1.MModeStateCompleted),
+			Status:             metav1.ConditionTrue,
+		},
+	)
+
+	if err := r.Client.Status().Update(r.ctx, maintenanceMode); err != nil {
```
nb-ohad, line 172: Is there really a reason to update if nothing changed?

nb-ohad (changes requested):

internal/controller/maintenancemode_controller.go
rewantsoni: The default value is true for all other events, I think we need to address it to send false for all other event types

internal/controller/maintenancemode_controller.go
rewantsoni: This was used since start in provider-client communication

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,199 @@
...
+	ctx context.Context
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	generationChangePredicate := predicate.GenerationChangedPredicate{}
+	maintenanceModeChangedPredicate := predicate.Funcs{
+		CreateFunc: func(_ event.CreateEvent) bool {
+			return false
+		},
```
nb-ohad, line 45: You cannot ignore create events, as you cannot assume that the client staus will start with an MM set to false

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,199 @@
...
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*v1alpha1.StorageClient)
+			newObj := e.ObjectNew.(*v1alpha1.StorageClient)
+			return oldObj.Status.InMaintenanceMode != newObj.Status.InMaintenanceMode
+		},
+		DeleteFunc: func(_ event.DeleteEvent) bool {
+			return false
+		},
```
nb-ohad, line 56: You cannot ignore delete events, if the client get deleted then the MM needs to be updated as well

internal/controller/maintenancemode_controller.go
nb-ohad: I am ok with the number, but don't we need a const to make sure all places use the same value?

internal/controller/maintenancemode_controller.go
```diff
@@ -0,0 +1,206 @@
...
+			ObservedGeneration: maintenanceMode.Generation,
+			Reason:             string(ramenv1alpha1.MModeStateCompleted),
+			Status:             metav1.ConditionTrue,
+		},
+	)
+
+	if err := r.Client.Status().Update(r.ctx, maintenanceMode); err != nil {
+		return err
+	}
+	return nil
```
nb-ohad, line 197:
```suggestion
	condition := metav1.Condition{
		Type:               string(ramenv1alpha1.MModeConditionFailoverActivated),
		Status:             metav1.ConditionTrue,
		Reason:             string(ramenv1alpha1.MModeStateCompleted),
		ObservedGeneration: maintenanceMode.Generation,
	}

	updateRequired := false
	updateRequired &= (maintenanceMode.Status.State != ramenv1alpha1.MModeStateCompleted)
	updateRequired &= slices.contains(maintenanceMode.Status.Conditions, condition)
	
	if updateRequired {
		maintenanceMode.Status.State = ramenv1alpha1.MModeStateCompleted
		maintenanceMode.Status.ObservedGeneration = maintenanceMode.Generation
		meta.SetStatusCondition(&maintenanceMode.Status.Conditions, condition)
	
		if err := r.Client.Status().Update(r.ctx, maintenanceMode); err != nil {
			return err
		}
	}
	return nil
```

nb-ohad (changes requested):

PR #265, opened 2024-11-07
controllers: create clientprofilemapping required for DR

rewantsoni:
Create the clientProfileMapping sent from the provider server required for DR

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -235,6 +235,21 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 			}); err != nil {
 				return reconcile.Result{}, fmt.Errorf("failed to reconcile cephConnection: %v", err)
 			}
+		case "ClientProfileMapping":
+			clientProfileMapping := &csiopv1a1.ClientProfileMapping{}
+			clientProfileMapping.Name = eResource.Name
+			clientProfileMapping.Namespace = r.OperatorNamespace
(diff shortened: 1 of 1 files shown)


internal/controller/storageclaim_controller.go
```diff
@@ -435,6 +435,21 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
...
 				}
+			case "ClientProfileMapping":
+				clientProfileMapping := &csiopv1a1.ClientProfileMapping{}
+				clientProfileMapping.Name = r.storageClaimHash
+				clientProfileMapping.Namespace = r.OperatorNamespace
+				if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, clientProfileMapping, func() error {
+					if err := r.own(clientProfileMapping); err != nil {
+						return fmt.Errorf("failed to own clientProfile resource: %v", err)
+					}
+					if err := json.Unmarshal(resource.Data, &clientProfileMapping.Spec); err != nil {
```
rchikatw, line 446: it would be good if we unmarshal first and if it goes successful we should own the resource WDYT?

rewantsoni: Followed what was there in other CreateOrUpdate

nb-ohad: It does not matter, the own operation only changes in memory settings. And as @rewantsoni mentioned above we usually start with the own operation.

rchikatw: Ack. you can resolve this conv.

PR #268, opened 2024-11-12
api: adding role and set correct name to get CRQ

rchikatw:
- Adding a cluster role to get the cluster resource quota from a status reporter
- the name of CRQ is not always same as storage client so changing the logic as per the format

### DIFF
--- a/config/rbac/status-reporter-clusterrole.yaml
+++ b/config/rbac/status-reporter-clusterrole.yaml
@@ -29,3 +29,9 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - quota.openshift.io
+    resources:
+      - clusterresourcequotas
+    verbs:
+      - get
--- a/pkg/utils/k8sutils.go
+++ b/pkg/utils/k8sutils.go
@@ -166,3 +166,7 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj client.Object) 
 	}
 	return nil
 }
+
+func GetClusterResourceQuotaName(name string) string {
+	return fmt.Sprintf("storage-client-%s-resourceqouta", GetMD5Hash(name))
+}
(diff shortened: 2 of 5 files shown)


service/status-report/main.go
```diff
@@ -207,7 +207,7 @@ func setClusterInformation(ctx context.Context, cl client.Client, status interfa
 
 func setStorageQuotaUtilizationRatio(ctx context.Context, cl client.Client, status interfaces.StorageClientStatus) {
 	clusterResourceQuota := &quotav1.ClusterResourceQuota{}
-	clusterResourceQuota.Name = status.GetClientName()
+	clusterResourceQuota.Name = fmt.Sprintf("storage-client-%s-resourceqouta", utils.GetMD5Hash(status.GetClientName()))
```
Madhu-1, line 210: please use helper function to get the Name so that we will have single source of truth to generate names, if not we are going to get into the problems where we are generating names in different places with  strings

config/rbac/status-reporter-clusterrole.yaml
```diff
@@ -29,3 +29,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - quota.openshift.io
+    resources:
+      - clusterresourcequotas
+    verbs:
+      - get
+      - list
```
Madhu-1, line 38: why do we need list access?

pkg/utils/k8sutils.go
```diff
@@ -166,3 +166,7 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj client.Object)
 	}
 	return nil
 }
+
+func GetClusterResourceQuotaName(name string) string {
+	return fmt.Sprintf("storage-client-%s-resourceqouta", GetMD5Hash(name))
```
Madhu-1, line 171: typo in `qouta` its `quota` 

btw why did we need resoucequota in the name as its already ClusterResourceQuota.

rchikatw: I cannot change the name of the resource right now because it has inherited from version 4.17. If I change it, I will need to conduct tests for the upgrade and some other scenarios. I will create a new task for this and work on it. Thank you for eyeing on that.

PR #266, opened 2024-11-10
Labels `VolumeSnapshotClass`

raaizik:
## Changes

- Labels `VolumeSnapshotClass` with RDR required parameters
[RHSTOR-5795](https://issues.redhat.com/browse/RHSTOR-5795)

PR #277, opened 2024-11-22
controllers: add missing owner on sc and vsc

rewantsoni:
Adding the ownerRef that was missed as a part of https://github.com/red-hat-storage/ocs-client-operator/pull/266


internal/controller/storageclaim_controller.go
```diff
@@ -365,6 +365,9 @@ func (r *StorageClaimReconciler) reconcilePhases() (reconcile.Result, error) {
 					storageClass = r.getCephRBDStorageClass()
 				}
 				err = utils.CreateOrReplace(r.ctx, r.Client, storageClass, func() error {
+					if err := r.own(storageClass); err != nil {
```
leelavg, line 368: before the regression the expectation was createorreplace code creates the ownerereference and is there any other place that it isn't intended?

leelavg (approved):

PR #286, opened 2024-12-02
controllers: deploy omap sidecar only when required

rewantsoni:
Instead of always deploying the omap sidecar, move to deploying it only when required.


internal/controller/operatorconfigmap_controller.go
```diff
@@ -151,6 +152,17 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	mirrorEnabledChangedPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*v1alpha1.StorageClient)
+			newObj := e.ObjectNew.(*v1alpha1.StorageClient)
+			return oldObj.Status.MirrorEnabled != newObj.Status.MirrorEnabled
+		},
+	}
```
nb-ohad, line 164: We usually do not predicate on a single field in the spec but on the entire spec using a generation change predicate (as we consider this an overoptimization), especially when considering CRs that are almost singletons.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -532,9 +558,20 @@ func (c *OperatorConfigMapReconciler) getNoobaaSubManagementConfig() bool {
...
+	if err := c.list(storageClients); err != nil {
+		return false, err
+	}
+
+	for idx := range storageClients.Items {
+		if storageClients.Items[idx].Status.MirrorEnabled {
+			return true, nil
+		}
+
+	}
```
nb-ohad, line 573: If we are to generate omap data in cases where a single client requests it, how would that affect the other clients that don't need it?

internal/controller/storageclient_controller.go
```diff
@@ -207,6 +207,7 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
 
 	if storageClientResponse.SystemAttributes != nil {
 		r.storageClient.Status.InMaintenanceMode = storageClientResponse.SystemAttributes.SystemInMaintenanceMode
+		r.storageClient.Status.MirrorEnabled = storageClientResponse.SystemAttributes.MirrorEnabled
```
nb-ohad, line 210: Because we are talking about the status section (where we reflect the actual state and not the desired state) I am not sure it is correct to reflect that the client is mirror-enabled before validating that the omap generator is running. In addition shouldn't we reflect similar things for other clients? shouldn't this be more granular?

nb-ohad (changes requested):

internal/controller/operatorconfigmap_controller.go
rewantsoni: Here, we are storing the MirroringEnabled flag in the status of the storageClient CR, I want the operatorConfigmap controller to trigger a reconciliation only when the field changes. I could watch for the entire status but I think it's better to be specific here

internal/controller/operatorconfigmap_controller.go
rewantsoni: If any client is requesting for the omap info, we are going to enable it, as this is a operator level operation we cannot restrict it to per client

internal/controller/storageclient_controller.go
rewantsoni: Here, I want to convey that the client is enabled for mirroring on the provider side. Hence storing the actual state in the Status, the desired state would be in the storage-client-mapping in the Provider side

internal/controller/operatorconfigmap_controller.go
nb-ohad: We have an issue here, status should not be treated as desired state and should be considered as something that can be deleted at any point in time.

internal/controller/operatorconfigmap_controller.go
nb-ohad: I understand, but I was asking about the impact/the side effects. Not the procedure

internal/controller/storageclient_controller.go
nb-ohad: We have an issue here, status should not be treated as desired state and should be considered as something that can be deleted at any point in time.

internal/controller/operatorconfigmap_controller.go
rewantsoni: The impact would be that we enable it for all other clients as well. I don't think we have any side effects if it is always enabled

PR #287, opened 2024-12-03
controllers: remove storageclaim from manager and move functionality

leelavg:
first cut of removal of storageclaim controller and moving the work over to the storageclient controller. Existing resources that are owned by storageclaims will become unmanaged but protected against deletion due to finalizer for which current expectation is to be removed manually

no provider changes done atm and the expectation is, provider would send info that it used to send for storageclaim now to storageclient and any opportunity to change the message struct will be made and corresondingly a new PR will be raised.


leelavg:
/hold

Against any accidental approvals.

leelavg:
closing in favour of #295

leelavg:
closing in favor of #302

PR #288, opened 2024-12-03
controllers: fix predicate to trigger reconcile for mmode controller

rewantsoni:
add an predicate.Or trigger reconcile of mmode controller when the storageClient's status is updated

### DIFF
--- a/internal/controller/maintenancemode_controller.go
+++ b/internal/controller/maintenancemode_controller.go
@@ -63,10 +63,7 @@ func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(
 			&v1alpha1.StorageClient{},
 			&handler.EnqueueRequestForObject{},
-			builder.WithPredicates(
-				generationChangePredicate,
-				maintenanceModeChangedPredicate,
-			),
+			builder.WithPredicates(maintenanceModeChangedPredicate),
 		).
 		Complete(r)
 }


internal/controller/maintenancemode_controller.go
```diff
@@ -64,8 +64,10 @@ func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&v1alpha1.StorageClient{},
 			&handler.EnqueueRequestForObject{},
 			builder.WithPredicates(
-				generationChangePredicate,
-				maintenanceModeChangedPredicate,
+				predicate.Or(
+					generationChangePredicate,
+					maintenanceModeChangedPredicate,
+				),
```
leelavg, line 70: not sure about this, we usually want to skip generation changes for objects to which we are only interested in spec changes and since you already have a predicate for MMChanged the other predicate isn't required.

rewantsoni: ack makes sense, I will update it

rewantsoni:
Testing:
```
without restarted the operator, mmode status was updated
[rewantsoni scripts] k get po | grep client
ocs-client-operator-console-7f75dbbb6d-dg95n                      1/1     Running     0               2d4h
ocs-client-operator-controller-manager-7b7b5df754-hhtrx           2/2     Running     0               4m51s
storageclient-737342087af10580-status-reporter-28888415-sntq8     0/1     Completed   0               30s
```
Screenshot 2024-12-04 at 3 06 48 PM

leelavg (approved):

PR #292, opened 2024-12-09
controllers: webhook for restricting storageclaims to auto created ones

leelavg:
we only support a single storageclaim of a specific type created corresponding to a storageclient and this webhook ensures that support.

this helps in avoiding surprises during storageclaim controller removal.


leelavg:
/hold

for testing.

rewantsoni (commented):
Since we are reconciling both subscription and storageClaim webhooks, can we rename `reconcileSubscriptionValidatingWebhook`?

leelavg:
not required as we didn't document support for creation of new claims and w/ or w/o this PR an already created claim has to be worked around.

PR #302, opened 2025-02-10
remove storageclaim api and corresponding controller along with references

leelavg:
api and controller along with the references of the same in other
controllers is removed.


leelavg:
Part of RHSTOR-6913

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +401,155 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
-		}
+		case "StorageClass":
+			storageClass := &storagev1.StorageClass{}
+			if err := json.Unmarshal(eResource.Data, &storageClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshal storage configuration response: %v", err)
+			}
+			if err := r.own(storageClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to own Storage Class resource: %v", err)
+			}
+			utils.AddLabels(storageClass, eResource.Labels)
```
rewantsoni, line 412: Had a silly question, CreateOrReplace will fetch the resource, apply the mutate function, and then compare the fetched resource with mutated right? 
If we don't add labels under CreateOrReplace won't that mean that labels won't persist?
Similarly won't we have the same problem for owning the StorageClass

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +401,155 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			// check for CRD availability
+			if r.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
+				volumeGroupSnapshotClass := &groupsnapapi.VolumeGroupSnapshotClass{}
+				volumeGroupSnapshotClass.Name = eResource.Name
+				err = utils.CreateOrReplace(r.ctx, r.Client, volumeGroupSnapshotClass, func() error {
+					if existing := metav1.GetControllerOfNoCopy(volumeGroupSnapshotClass); existing != nil &&
+						existing.Kind != r.storageClient.Kind {
+						existing.BlockOwnerDeletion = nil
+						existing.Controller = nil
+					}
```
rewantsoni, line 454: Won't we need this for the default StorageClass as well?
For VolumeSnapshotClass the names will change and we won't need this right?

config/rbac/role.yaml
```diff
@@ -175,14 +175,6 @@ rules:
...
   - list
   - watch
-- apiGroups:
-  - groupsnapshot.storage.k8s.io
-  resources:
-  - volumegroupsnapshotcontents
-  verbs:
-  - get
-  - list
-  - watch
```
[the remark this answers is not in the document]
leelavg: I'll recheck, probably the marker wasn't migrated properly.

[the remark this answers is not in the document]
leelavg: done, marker was missed.

internal/controller/storageclient_controller.go
leelavg: done.

internal/controller/storageclient_controller.go
leelavg: done, another reason why I didn't do is, initial assumption was storageclass will be reconciled at ocs-op as well and now that isn't correct made the changes rather than waiting for another release.

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +402,165 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
-		}
+		case "StorageClass":
+			storageClass := &storagev1.StorageClass{}
+			if err := json.Unmarshal(eResource.Data, &storageClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshal storage configuration response: %v", err)
+			}
+			// TODO: there will be a clash if storageclass is being reconcile by another entity
+			// ex: in internal mode storageclass was created by storagecluster controller
+			// and if we are mutating anything other than params there'll be unending reconcile loop
+			if err := utils.CreateOrReplace(r.ctx, r.Client, storageClass, func() error {
```
rewantsoni, line 413: - There is no name for StorageClass, createOrReplace will fail
- in the CreateOrReplace the provisioner and the reclaimPolicy is missing

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +402,165 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+				utils.AddLabels(storageClass, eResource.Labels)
+				scParams := storageClass.Parameters
+				scParams["csi.storage.k8s.io/provisioner-secret-namespace"] = r.OperatorNamespace
+				scParams["csi.storage.k8s.io/node-stage-secret-namespace"] = r.OperatorNamespace
+				scParams["csi.storage.k8s.io/controller-expand-secret-namespace"] = r.OperatorNamespace
 
-		if err := r.reconcileSharedfileStorageClaim(); err != nil {
-			return reconcile.Result{}, err
-		}
+				storageClass.Parameters = scParams
```
rewantsoni, line 428: We don't need a new variable here

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +402,165 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+				return nil
+			}); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to create or update StorageClass: %s", err)
+			}
+		case "VolumeSnapshotClass":
+			snapshotClass := &snapapi.VolumeSnapshotClass{}
+			if err := json.Unmarshal(eResource.Data, snapshotClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to unmarshal storage configuration response: %v", err)
+			}
+			if err := utils.CreateOrReplace(r.ctx, r.Client, snapshotClass, func() error {
```
rewantsoni, line 438: - snapshotClass name is empty
- fields driver and deletionPolicy is missing

internal/controller/storageclient_controller.go
leelavg: fixed.

internal/controller/storageclient_controller.go
leelavg: no, that data should be coming from provider and unmarshalled before mutation.

rewantsoni: But we are unmarshalling eResource.Data that contains the parameters isn't it? is there a provider side PR too to have these changes?

leelavg: updated to be consistent, always respect separate labels & annotations from proto and everything else from unmarshalled data, since storage & snapshot classes doesn't have a spec we'll update whatever required (namespace) and consume the remaining of whatever provider sends.

internal/controller/storageclient_controller.go
leelavg: updated.

internal/controller/maintenancemode_controller.go
```diff
@@ -55,11 +55,6 @@ func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&handler.EnqueueRequestForObject{},
 			builder.WithPredicates(generationChangePredicate),
 		).
-		Watches(
-			&v1alpha1.StorageClaim{},
-			&handler.EnqueueRequestForObject{},
-			builder.WithPredicates(generationChangePredicate),
-		).
```
nb-ohad, line 62: You are removing the watch but you don't address the code in the controller that deals with StorageClaim resources

internal/controller/operatorconfigmap_controller.go
```diff
@@ -393,13 +393,13 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
...
+	clientsList := &v1alpha1.StorageClientList{}
+	if err := c.list(clientsList, client.Limit(1)); err != nil {
+		c.log.Error(err, "unable to verify StorageClients presence prior to removal of CSI resources")
 		return err
-	} else if len(claimsList.Items) != 0 {
-		err = fmt.Errorf("failed to clean up resources: storage claims are present on the cluster")
-		c.log.Error(err, "Waiting for all storageClaims to be deleted.")
+	} else if len(clientsList.Items) != 0 {
+		err = fmt.Errorf("failed to clean up resources: storage clients are present on the cluster")
+		c.log.Error(err, "Waiting for all storageClients to be deleted.")
```
nb-ohad, line 402: Why is that needed? The reason we where looking into the existence of claim was that claims themselves where protected from deletion until all PV/PVCs are gone. Are we now giving similar guarantees with the clients? And if so where is the change?

nb-ohad (changes requested):

internal/controller/operatorconfigmap_controller.go
leelavg: yes, all the code was mode to storageclient controller, maybe you need to expand on the hidden files in github. for ex https://github.com/red-hat-storage/ocs-client-operator/blob/e701fb8129fd86e36538b94348d3cc7309b6a1e0/internal/controller/storageclient_controller.go#L829

internal/controller/storageclient_controller.go
```diff
@@ -55,54 +64,130 @@ const (
...
+					return []string{parts[2]}
+				}
+			}
+			return nil
+		}); err != nil {
+			return fmt.Errorf("unable to set up FieldIndexer for VGSC csi driver name: %v", err)
+		}
 	}
 
+	enqueueStorageClientRequest := handler.EnqueueRequestsFromMapFunc(
```
Madhu-1, line 145: There was a change done to reconcile the storageclaims when the annotation is changed, is this logic moved to storageclient in this PR or its going to be a separate PR?

internal/controller/storageclient_controller.go
```diff
@@ -118,21 +203,38 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=noobaa.io,resources=noobaas,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofilemappings,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list;watch;create;update;patch;delete
```
Madhu-1, line 206: i assume its a copy paste code, but we dont need update or patch access to class isnt it?

internal/controller/storageclient_controller.go
```diff
@@ -118,21 +203,38 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+//+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotclasses,verbs=get;list;watch;create;delete
+//+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotclasses,verbs=create;delete;get;list;watch
```
Madhu-1, line 208: can you please order the verb like in other places `get;list;...`

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +403,125 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			})
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to create or update VolumeReplicationClass: %s", err)
+			}
+		case "ClientProfile":
+			clientProfile := &csiopv1a1.ClientProfile{}
+			clientProfile.Name = eResource.Name
+			clientProfile.Namespace = r.OperatorNamespace
+			if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, clientProfile, func() error {
+				removeStorageClaimAsOwner(clientProfile)
```
Madhu-1, line 511: This is not required as claim never owned clientProfile?

internal/controller/storageclient_controller.go
```diff
@@ -638,36 +802,49 @@ func (r *StorageClientReconciler) reconcileClientStatusReporterJob() (reconcile.
 	return reconcile.Result{}, nil
 }
 
-func (r *StorageClientReconciler) list(obj client.ObjectList, listOptions ...client.ListOption) error {
-	return r.Client.List(r.ctx, obj, listOptions...)
+func (r *StorageClientReconciler) hasPersistentVolumes() (bool, error) {
+	pvList := &corev1.PersistentVolumeList{}
+	if err := r.list(pvList, client.MatchingFields{pvClusterIDIndexName: r.storageClientHash}, client.Limit(1)); err != nil {
```
Madhu-1, line 807: Will this also works for the existing cephfs/rbd PV's (upgraded)? same is applicabel for other resouces like (VGSC,VGC etc)

internal/controller/storageclient_controller.go
leelavg: reconcile on annotation change is originally on storageclient and it continue to exists.

internal/controller/storageclient_controller.go
leelavg: good catch, this is a bug in two ways
1. w/ convergence the clusterID will not be a hash but an UUID which is a bit difficult to get now, needs some thinking
2. your original assessment, ie, this skips pre-convergence PVs/VSCs/VGSCs

nb-ohad: Why can't we go by provisioner? is the concern that multiple clients will be using the same provisioner name. I am not sure that will be the case

leelavg: rechecked, there is no reference to provisioner in PV

internal/controller/storageclient_controller.go
leelavg: cephconnection wasn't owned by claim but clientprofile is.

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go
leelavg: updated.

Madhu-1: >Why can't we go by provisioner? is the concern that multiple clients will be using the same provisioner name. I am not sure that will be the case

AFAIK it will be the same provisioner, Think about the case of RDR/Metro DR the provisioner names need to be the same for the PV's and also in ODF we never had concept of multiple provisioners (even for internal+external mode). If this is something we are planning to change we need a bigger discussion

leelavg:
/hold

Will squash after final review

Madhu-1 (commented):
LGTM

internal/controller/storageclient_controller.go
Madhu-1: This is addressed now, not using the provisioner for identification

leelavg:
/unhold

squashed and rebased.

internal/controller/storageclient_controller.go
```diff
@@ -55,54 +64,143 @@ const (
...
-
+	Scheme            *runtime.Scheme
 	OperatorNamespace string
+	AvailableCrds     map[string]bool
+
+	log               klog.Logger
+	ctx               context.Context
+	recorder          *utils.EventReporter
+	storageClient     *v1alpha1.StorageClient
+	storageClientHash string
```
nb-ohad, line 92: You should not store request specific information on a reconciler object that is shared across all reconciles

internal/controller/storageclient_controller.go
```diff
@@ -55,54 +64,143 @@ const (
...
+				// second entry in the volumeID is clusterID which is unique across the cluster
+				return []string{parts[2]}
+			}
+		}
+		return nil
+	}); err != nil {
+		return fmt.Errorf("unable to set up FieldIndexer for VSC csi driver name: %v", err)
+	}
+
+	if r.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
```
nb-ohad, line 128: Why a map? is there a reason to introduce a map and not just read/list from the cached client?

internal/controller/storageclient_controller.go
```diff
@@ -55,54 +64,143 @@ const (
...
+			storageClients := &v1alpha1.StorageClientList{}
+			if err := r.Client.List(ctx, storageClients); err != nil {
+				return []reconcile.Request{}
+			}
+
+			request := []reconcile.Request{}
+			for idx := range storageClients.Items {
+				request = append(request, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&storageClients.Items[idx])})
+			}
+			return request
```
nb-ohad, line 169: Why are we queueing an event for all storage clients?

internal/controller/storageclient_controller.go
```diff
@@ -55,54 +64,143 @@ const (
...
+					false,
+					true,
+					false,
+				),
+			),
+			builder.OnlyMetadata,
+		)
+	if r.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
+		bldr = bldr.Owns(&groupsnapapi.VolumeGroupSnapshotClass{})
+	}
```
nb-ohad, line 203: We know how to add events dynamically. Why are we moving code that contain a legacy pattern?

internal/controller/storageclient_controller.go
```diff
@@ -118,21 +216,38 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofiles,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplicationclasses,verbs=get;list;watch;create;delete
+//+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotcontents,verbs=get;list;watch
 
 func (r *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	var err error
 	r.ctx = ctx
-	r.Log = log.FromContext(ctx, "StorageClient", req)
-	r.Log.Info("Reconciling StorageClient")
+	r.log = log.FromContext(ctx, "StorageClient", req)
```
nb-ohad, line 231: You should not be saving a request specific logger on the reconciler

internal/controller/storageclient_controller.go
```diff
@@ -118,21 +216,38 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	r.log.Info("Reconciling StorageClient")
+
+	crd := &metav1.PartialObjectMetadata{}
+	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+	crd.Name = VolumeGroupSnapshotClassCrdName
+	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
+		r.log.Error(err, "Failed to get CRD", "CRD", crd.Name)
+		return reconcile.Result{}, err
+	}
+	utils.AssertEqual(r.AvailableCrds[crd.Name], crd.UID != "", utils.ExitCodeThatShouldRestartTheProcess)
```
nb-ohad, line 241: My understanding is that assert will kill the process, we don't want that

internal/controller/storageclient_controller.go
```diff
@@ -118,21 +216,38 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+//+kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list;watch;create;delete
+//+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotclasses,verbs=get;list;watch;create;delete
+//+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotclasses,verbs=get;list;watch;create;delete;
+//+kubebuilder:rbac:groups=core,resources=persistentvolumes,verbs=get;list;watch
+//+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotcontents,verbs=get;list;watch
+//+kubebuilder:rbac:groups=csi.ceph.io,resources=clientprofiles,verbs=get;list;update;create;watch;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplicationclasses,verbs=get;list;watch;create;delete
+//+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotcontents,verbs=get;list;watch
 
 func (r *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
nb-ohad, line 228: You have a lot of returns in this function where you are not guaranteeing update to the client's status.

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +416,125 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
 
-		if err := r.reconcileSharedfileStorageClaim(); err != nil {
-			return reconcile.Result{}, err
-		}
+				if storageClass.Parameters == nil {
+					storageClass.Parameters = map[string]string{}
+				}
+				storageClass.Parameters["csi.storage.k8s.io/provisioner-secret-namespace"] = r.OperatorNamespace
+				storageClass.Parameters["csi.storage.k8s.io/node-stage-secret-namespace"] = r.OperatorNamespace
+				storageClass.Parameters["csi.storage.k8s.io/controller-expand-secret-namespace"] = r.OperatorNamespace
```
nb-ohad, line 438: Why are we adding these parameters here and not on the provider?

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +416,125 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+					return fmt.Errorf("failed to own VolumeSnapshotClass resource: %v", err)
+				}
 
-	if utils.AddAnnotation(r.storageClient, utils.DesiredConfigHashAnnotationKey, storageClientResponse.DesiredConfigHash) {
-		if err := r.update(r.storageClient); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to update StorageClient with desired config hash annotation: %v", err)
+				if snapshotClass.Parameters == nil {
+					snapshotClass.Parameters = map[string]string{}
+				}
+				snapshotClass.Parameters["csi.storage.k8s.io/snapshotter-secret-namespace"] = r.OperatorNamespace
```
nb-ohad, line 462: Why are we adding these parameters here and not on the provider?

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +416,125 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+					volumeGroupSnapshotClass.Name = eResource.Name
+					removeStorageClaimAsOwner(volumeGroupSnapshotClass)
+					if err := r.own(volumeGroupSnapshotClass); err != nil {
+						return fmt.Errorf("failed to own VolumeGroupSnapshotClass resource: %v", err)
+					}
+
+					if volumeGroupSnapshotClass.Parameters == nil {
+						volumeGroupSnapshotClass.Parameters = map[string]string{}
+					}
+					volumeGroupSnapshotClass.Parameters["csi.storage.k8s.io/group-snapshotter-secret-namespace"] = r.OperatorNamespace
```
nb-ohad, line 489: Why are we adding these parameters here and not on the provider?

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +416,125 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+			vrc.Name = eResource.Name
+			err := utils.CreateOrReplace(r.ctx, r.Client, vrc, func() error {
+				removeStorageClaimAsOwner(vrc)
+				if err := r.own(vrc); err != nil {
+					return fmt.Errorf("failed to own VolumeReplicationClass resource: %v", err)
+				}
+				if err := json.Unmarshal(eResource.Data, &vrc.Spec); err != nil {
+					return fmt.Errorf("failed to unmarshall VolumeReplicationClass spec: %v", err)
+				}
+				vrc.Spec.Parameters["replication.storage.openshift.io/replication-secret-namespace"] = r.OperatorNamespace
```
nb-ohad, line 510: Why are we adding these parameters here and not on the provider?

internal/controller/storageclient_controller.go
```diff
@@ -309,26 +416,125 @@ func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
...
+				removeStorageClaimAsOwner(clientProfile)
+				if err := r.own(clientProfile); err != nil {
+					return fmt.Errorf("failed to own clientProfile resource: %v", err)
+				}
+				if err := json.Unmarshal(eResource.Data, &clientProfile.Spec); err != nil {
+					return fmt.Errorf("failed to unmarshall clientProfile spec: %v", err)
+				}
+				clientProfile.Spec.CephConnectionRef = corev1.LocalObjectReference{
+					Name: r.storageClient.Name,
+				}
```
nb-ohad, line 533: Why can't we send this from the provider? we should not handle specific logic for different types of resources on the client side

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
leelavg: `VolumeGroupSnapshotClassCrd` are not available in some OCP versions. @Madhu-1 do you know from which version these APIs are available by default?

internal/controller/storageclient_controller.go
leelavg: this is enqueued only when the vgsc crd becomes available and all clients need to fetch for vgsc.

internal/controller/storageclient_controller.go
leelavg: sry, I don't follow, we only have one extra return for the crd check and the remaining are usual one's.

internal/controller/storageclient_controller.go
leelavg: #323 is the client side fix and provider side will be done at a later point.

internal/controller/storageclient_controller.go
leelavg: #323 is the client side fix and provider side will be done at a later point.

internal/controller/storageclient_controller.go
leelavg: #323 is the client side fix and provider side will be done at a later point.

internal/controller/storageclient_controller.go
leelavg: #323 is the client side fix and provider side will be done at a later point.

internal/controller/storageclient_controller.go
leelavg: yes, will do it from provider end.

internal/controller/storageclient_controller.go
leelavg: pinged for ack to make changes offline.

internal/controller/storageclient_controller.go
leelavg: pinged for ack to make changes offline.

internal/controller/storageclient_controller.go
leelavg: pinged for ack to make changes offline.

internal/controller/storageclient_controller.go
leelavg: pinged for ack to make changes offline.

internal/controller/storageclient_controller.go
leelavg: I checked w/ Rakshith, the API is not available by default in 4.19.

Madhu-1: sorry i missed the ping, Openshift is planning to have volumegroupsnapshot GA in next OCP 4.18.z release which means it might come in OCP 4.19

internal/controller/storageclient_controller.go
leelavg: will be worked in a different PR.

internal/controller/storageclient_controller.go
leelavg: will be worked in a different PR.

internal/controller/storageclient_controller.go
leelavg: will be worked in a different PR.

PR #304, opened 2025-02-10
controllers: update maintenance mode controller

rewantsoni:
update the maintenance mode controller to put the storage clients in maintenance mode when a request is received

### DIFF
--- a/internal/controller/maintenancemode_controller.go
+++ b/internal/controller/maintenancemode_controller.go
@@ -77,39 +77,15 @@ func (r *MaintenanceModeReconciler) Reconcile(ctx context.Context, _ ctrl.Reques
 	r.log = log.FromContext(ctx)
 	r.log.Info("Starting reconcile")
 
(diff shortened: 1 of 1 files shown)


rewantsoni:
Tested it and the maintenanceMode is reconciled when it's created as per the expectaion

internal/controller/maintenancemode_controller.go
```diff
@@ -77,50 +77,46 @@ func (r *MaintenanceModeReconciler) Reconcile(ctx context.Context, _ ctrl.Reques
 	r.log = log.FromContext(ctx)
 	r.log.Info("Starting reconcile")
 
-	nameToStorageClient := map[string]*v1alpha1.StorageClient{}
+	//TODO: This controller assumes that there is only one client and that the maintenance mode is created only for that
```
leelavg, line 80: will this be implemented in next version? if not, create a corresponding issue and tag to this PR for tracking.

PR #311, opened 2025-02-26
remove kube-rbac-proxy container from deployment

leelavg:
we are not exporting any metrics exclusively and kube-rbac-proxy prominently used to authenticate api access to metrics endpoint available on operator (manager) which isn't mandatory for us and this commit removes the same.

note that controller-runtime itself exports some metrics and this commit doesn't bind any port for controller-runtime which disables inbuilt metrics being available but have all the support infra intact for future expansion.

ref: https://book.kubebuilder.io/reference/metrics-reference and https://book.kubebuilder.io/reference/metrics


bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
```diff
@@ -482,28 +482,7 @@ spec:
...
-                    cpu: 500m
-                    memory: 128Mi
-                  requests:
-                    cpu: 5m
-                    memory: 64Mi
-              - args:
-                - --health-probe-bind-address=:8081
-                - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
-                - --console-port=9001
```
Madhu-1, line 506: is this change related to kube-proxy and its expected?

cmd/main.go
```diff
@@ -89,12 +93,28 @@ func init() {
...
 	var probeAddr string
+	var secureMetrics bool
 	var consolePort int
 	var webhookPort int
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	var tlsOpts []func(*tls.Config)
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
+		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
```
Madhu-1, line 107: This looks like duplicate flag declaration

leelavg: yes, will remove.

bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
leelavg: only the metrics-bind-address, can remove the remaining as well as they are using defaults defined in code.

leelavg: all other options are defaulted and console-port is being set explicitly even though it's the same value

cmd/main.go
leelavg: done.

leelavg:
@Madhu-1 possible to try `make bundle-build` w/ this PR? It's failing only in CI, thanks.

leelavg:
disabling golang cache based on https://github.com/actions/setup-go/issues/403#issuecomment-2299976549

Madhu-1 (approved):

PR #314, opened 2025-03-04
add nfs service account to scc

Madhu-1:
Add missing nfs service account to scc


leelavg (approved):

Madhu-1:
@leelavg any idea why lgtm and approved labels are not added?

PR #318, opened 2025-03-13
ci: refactor and fix gitignore & make targets

leelavg:
refactor of CI jobs:
divide the checks b/n code vs non-code which helps in easily turning off one or the other during fixing any CI issues and also consolidates jobs by similarity.

gitignore fix:
pr #311 brought new vendor dependency and that is being git ignored unintentionally which created failures while running in CI, previous PR was merged as not to get blocked by explicitly doing a `go mod tidy` and `go mod vendor` which masked the original issue.

as not to hit this again, added a new condition for verifying deps from mandatory parts for build if they are being git ignored.

golang cache:
as we were using `vendor/` for the project there is no need to cache mod dependencies again and this is removed except for a single job where it makes sense, infact storing & loading cache is increasing the overall time w/ no added benefit.

concurrency of jobs:
added a concurrency check to the action as to cancel running jobs if PR is pushed in quick successions as not to do work on older changes.

makefile:
do not update dependencies implicitiy before testing as that could mask potentially issues which effectively makes updating dependencies a separate step.


leelavg:
/unhold

no more changed expected, done for this round of CI fixes.

hack/godeps-verify.sh
```diff
@@ -6,5 +6,8 @@ if [[ -n "$(git status --porcelain "${paths[@]}")" ]]; then
 	git diff -u "${paths[@]}"
 	echo "Inconsistency found in dependency files. Run 'make godeps-update' and commit results."
 	exit 1
+elif [[ -n "$(git status --ignored --porcelain "${paths[@]}")" ]]; then
+	echo "Your [.gitignore] file is excluding required files for build from [${paths[@]}]"
```
Madhu-1, line 10:
```suggestion
echo ".gitignore file is excluding required files for build from [${paths[@]}]"
```

.github/workflows/verify-non-code.yaml
```diff
@@ -0,0 +1,51 @@
+---
+name: non-coding-checks
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+on:
+  push:
+    branches: ['*']
```
Madhu-1, line 8: we dont need to run this workflow on the push , can we remove it?

.github/workflows/verify-code.yaml
```diff
@@ -1,63 +1,61 @@
 ---
-name: image-builder
-
+name: coding-checks
```
Madhu-1, line 2:
```suggestion
name: code-and-container-pipeline
```

can we use somethink like above or `operator-build-pipeline`? etc

.github/workflows/verify-non-code.yaml
```diff
@@ -0,0 +1,51 @@
+---
+name: non-coding-checks
```
Madhu-1, line 2: `formatting-and-linting` or `code-integrity-checks` or `quality-check` or something like that?

hack/godeps-verify.sh
leelavg: done.

.github/workflows/verify-non-code.yaml
leelavg: done.

.github/workflows/verify-code.yaml
leelavg: done.

.github/workflows/verify-non-code.yaml
leelavg: used `lint-and-spellcheck`

Madhu-1 (approved):

leelavg:
I've pinned the latest actions in https://github.com/red-hat-storage/ocs-client-operator/actions and will be disabling the rest after the merge.

PR #322, opened 2025-03-21
controllers: remove ack step from storageclient onboarding

leelavg:
storageconsumer is pre-created on the provider side and the token contains to which consumer current client should connect to and once connected (~onboarded) storageconfig is exchanged without any other intermediate step.

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -20,15 +20,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	quotav1 "github.com/openshift/api/quota/v1"
-	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
-	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 	"os"
 	"strings"
 
+	quotav1 "github.com/openshift/api/quota/v1"
+	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+
 	csiopv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
-	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -406,68 +405,50 @@ func (r *StorageClientReconciler) newExternalClusterClient() (*providerClient.OC
...
 
 	// TODO Need to find a way to get rid of ClusterVersion here as it is OCP
 	// specific one.
 	clusterVersion := &configv1.ClusterVersion{}
 	clusterVersion.Name = "version"
 	if err := r.get(clusterVersion); err != nil {
 		r.Log.Error(err, "failed to get the clusterVersion version of the OCP cluster")
-		return reconcile.Result{}, fmt.Errorf("failed to get the clusterVersion version of the OCP cluster: %v", err)
+		return fmt.Errorf("failed to get the clusterVersion version of the OCP cluster: %v", err)
 	}
```
rewantsoni, line 417: We don't need to fetch the clusterVersion anymore, it was used to decide the name of the storageConsumer

rewantsoni (changes requested):

internal/controller/storageclient_controller.go

rewantsoni:
LGTM

internal/controller/storageclient_controller.go
```diff
@@ -406,68 +404,41 @@ func (r *StorageClientReconciler) newExternalClusterClient() (*providerClient.OC
...
 	r.storageClient.Status.ConsumerID = response.StorageConsumerUUID
-	r.storageClient.Status.Phase = v1alpha1.StorageClientOnboarding
-
-	r.Log.Info("onboarding started")
-	return reconcile.Result{Requeue: true}, nil
-}
-
-func (r *StorageClientReconciler) acknowledgeOnboarding(externalClusterClient *providerClient.OCSProviderClient) (reconcile.Result, error) {
-
-	_, err := externalClusterClient.AcknowledgeOnboarding(r.ctx, r.storageClient.Status.ConsumerID)
```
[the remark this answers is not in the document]
leelavg: > AFAIK there is no flow change on the ocs-client-operator it will remain same
- pre4.19, enable is set to true when ack rpc is received as we are doing that during onboard we don't need the ack step.

PR #325, opened 2025-03-27
controllers: reconcile a single request in isolation with other requests

leelavg:
removal of event reporting is also done due to:
1. as of now it only supports a single type event for failed scenarios
2. we want a reporting which can raise multiple types of events


internal/controller/storageclient_controller.go
```diff
@@ -101,111 +98,126 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	r.Log = log.FromContext(ctx, "StorageClient", req)
-	r.Log.Info("Reconciling StorageClient")
-
-	r.storageClient = &v1alpha1.StorageClient{}
-	r.storageClient.Name = req.Name
-	if err = r.get(r.storageClient); err != nil {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Starting reconcile iteration for StorageClient", "req", req)
+
+	reconcileHander := storageClientReconcile{}
```
nb-ohad, line 114: I think you meant to call this reconcileHandler and not reconcileHander

internal/controller/storageclient_controller.go
```diff
@@ -101,111 +98,126 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	reconcileHander.ctx = ctx
+	reconcileHander.log = log
+	reconcileHander.storageClient.Name = req.Name
+
+	err := reconcileHander.reconcile()
+	if err != nil {
+		log.Error(err, "StorageClient reconciliation failed")
+	} else {
+		log.Info("StorageClient reconciliation completed successfully")
+	}
```
nb-ohad, line 125: if we are to delegate this to an individual request handler then it make more sense to have the entirety of the process handler internally

 ```suggestion
        handler := storageClientReconcile{ StorageClientReconciler = r }
        return handler.reconcile(ctx, request)
```

File: internal/controller/storageclient_controller.go [hunk 3]
```diff
@@ -101,111 +98,126 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	return ctrl.Result{}, err
+}
+
+func (r *storageClientReconcile) reconcile() error {
```
nb-ohad, line 130: We should return a result object as well.

internal/controller/storageclient_controller.go
```diff
@@ -101,111 +98,126 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		err = statusErr
+		return statusErr
 	}
-	return result, err
+
+	return nil
 }
 
-func (r *StorageClientReconciler) reconcilePhases() (ctrl.Result, error) {
+func (r *storageClientReconcile) reconcilePhases() error {
```
nb-ohad, line 158: We should return a result object as well.

internal/controller/storageclient_controller.go
```diff
@@ -386,82 +392,31 @@ func (r *StorageClientReconciler) onboardConsumer(externalClusterClient *provide
...
-		r.recorder.ReportIfNotPresent(r.storageClient, eventType, eventReason, msg)
+func (r *storageClientReconcile) offboardConsumer(externalClusterClient *providerClient.OCSProviderClient) error {
+	if _, err := externalClusterClient.OffboardConsumer(r.ctx, r.storageClient.Status.ConsumerID); err != nil {
+		return fmt.Errorf("failed to offboard consumer: %v", err)
 	}
+	return nil
 }
 
-func (r *StorageClientReconciler) reconcileClientStatusReporterJob() (reconcile.Result, error) {
+func (r *storageClientReconcile) reconcileClientStatusReporterJob() error {
```
nb-ohad, line 419: We should return a result object as well.

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
```diff
@@ -65,19 +61,21 @@ const (
...
-	Scheme        *runtime.Scheme
-	recorder      *utils.EventReporter
-	storageClient *v1alpha1.StorageClient
-
+	Scheme            *runtime.Scheme
 	OperatorNamespace string
 }
 
+type storageClientReconcile struct {
+	StorageClientReconciler
```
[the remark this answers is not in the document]
leelavg: yes, I fixed it in next PR but will make a change here.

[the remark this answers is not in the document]

internal/controller/storageclient_controller.go
leelavg: removed.

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

PR #326, opened 2025-03-27
controllers: add dynamic watches for volumesnapshotclass

leelavg:
add a watch on the crd by listening to it's creation dynamically. this effectively removes container restarts for establishing new watches.

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -21,28 +21,38 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
+	"sync"
 
 	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/templates"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 
 	csiopv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
 	"github.com/go-logr/logr"
+	groupsnapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	quotav1 "github.com/openshift/api/quota/v1"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -54,20 +65,32 @@ const (
...
 // StorageClientReconciler reconciles a StorageClient object
 type StorageClientReconciler struct {
 	client.Client
 	Scheme            *runtime.Scheme
 	OperatorNamespace string
+
+	cache            cache.Cache
+	controller       controller.Controller
+	mu               sync.RWMutex
+	crdsBeingWatched map[string]bool
```
nb-ohad, line 89: Why use this and not a sync map to simplify the code and maintenance?

internal/controller/storageclient_controller.go
```diff
@@ -54,20 +65,32 @@ const (
...
+	vgscClusterIDIndexName = "index:volumeGroupSnapshotContentCSIDriver"
+
 	csvPrefix = "ocs-client-operator"
 
 	VolumeGroupSnapshotClassCrdName = "volumegroupsnapshotclasses.groupsnapshot.storage.k8s.io"
 )
 
+var (
+	crdsToWatch = []string{VolumeGroupSnapshotClassCrdName}
+	csiDrivers  = []string{templates.RBDDriverName, templates.CephFsDriverName}
```
nb-ohad, line 77:
```suggestion
csiDrivers = []string{
    templates.RBDDriverName, 
    templates.CephFsDriverName,
 }

File: internal/controller/storageclient_controller.go [hunk 3]
```diff
@@ -77,15 +100,47 @@ type storageClientReconcile struct {
...
 		Owns(&csiopv1a1.CephConnection{}, builder.WithPredicates(generationChangePredicate)).
-		Owns(&csiopv1a1.ClientProfileMapping{}, builder.WithPredicates(generationChangePredicate))
-	return bldr.Complete(r)
+		Owns(&csiopv1a1.ClientProfileMapping{}, builder.WithPredicates(generationChangePredicate)).
+		Watches(
+			&extv1.CustomResourceDefinition{},
+			enqueueStorageClients,
+			builder.WithPredicates(
+				predicate.NewPredicateFuncs(func(obj client.Object) bool {
+					return slices.Contains(crdsToWatch, obj.GetName())
```
nb-ohad, line 131: Cant we just use the keys of the map for this?

internal/controller/storageclient_controller.go
```diff
@@ -156,8 +210,77 @@ func (r *storageClientReconcile) reconcile() error {
...
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.crdsBeingWatched[crdName]
+}
+
+func (r *storageClientReconcile) markCrdBeingWatched(crdName string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.crdsBeingWatched[crdName] = true
+}
```
nb-ohad, line 223: You are just implementing SyncMap set and get operation here. Just another reason to use SyncMap

internal/controller/storageclient_controller.go
```diff
@@ -156,8 +210,77 @@ func (r *storageClientReconcile) reconcile() error {
...
+
+func (r *storageClientReconcile) reconcileDynamicWatches() error {
+	for idx := range crdsToWatch {
```
nb-ohad, line 226: There is no point in running in a loop if each iteration need specific logic based on the name.
Just have normal non iterative code to set up the watch. Similar to what we have in SetupWithManager

nb-ohad:
Please add a description to the PR

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

PR #327, opened 2025-03-27
controllers: reconcile resources that will be part of GetStorageConfig

leelavg:
these were the resources which were previously being reconciled by storageclaim controller and now moved to storageclient controller as they will now be sent as response for GetStorageConfig.

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -30,14 +30,18 @@ import (
 	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 
 	csiopv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
+	replicationv1a1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
 	"github.com/go-logr/logr"
 	groupsnapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
+	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	quotav1 "github.com/openshift/api/quota/v1"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -428,6 +479,100 @@ func (r *storageClientReconcile) reconcilePhases() error {
...
+				storageClass.Name = eResource.Name
+				removeStorageClaimAsOwner(storageClass)
+				if err := r.own(storageClass); err != nil {
+					return fmt.Errorf("failed to own Storage Class resource: %v", err)
+				}
+				utils.AddLabels(storageClass, eResource.Labels)
+				utils.AddAnnotations(storageClass, eResource.Annotations)
+				return nil
+			}); err != nil {
+				return fmt.Errorf("failed to create or update StorageClass: %s", err)
```
nb-ohad, line 499: The majority of the steps in these new cases (and most of the old cases) are identical and can work on a client.Object interface. Can we please use a util function for all of these steps? (everything after the Unmarshal?

internal/controller/storageclient_controller.go
```diff
@@ -609,6 +779,64 @@ func (r *storageClientReconcile) reconcileClientStatusReporterJob() error {
 	return nil
 }
 
+func (r *storageClientReconcile) hasPersistentVolumes(clientProfileNames []string) (bool, error) {
```
nb-ohad, line 782: Cant we implement a single func that handles all of the `has` cases using `PartialMetadataObjectList`?

nb-ohad (changes requested):

leelavg:
rebased.

internal/controller/storageclient_controller.go
leelavg: not possible as we have index on the volumehand keyed by the clientprofile and so we can't do w/ partial obj.

leelavg:
rebased.

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go
leelavg: deferred as agreed but need to optimize this as there could be a lot of PVs in the cluster and all we need is to only know the presence rather than ballooning the cache.

leelavg:
after initial approval only linting issue raised by CI is fixed and no change introduced.

PR #333, opened 2025-04-01
bundle: remove optional mount for nginx-configmap

leelavg:
when an optional mount with subPath is used we are consistently observing failure at kubelet.

Error: container create failed: mount `/var/lib/kubelet/pods/<uid>/ volume-subpaths/ ocs-client-operator-console-nginx-conf/ ocs-client-operator-console/1` to `etc/nginx/nginx.conf`: Not a directory

we no longer need an `optional` mount for this configMap which only delays the starting of the container until CM is available but we no longer hit above issue.


leelavg:
Tested w/ reproducing the issue.

leelavg:
/lgtm

just kicking in the bot.

PR #334, opened 2025-04-02
controllers: fix ensureClusterScopedResources

rewantsoni:
the mutate function should have the unmarshall, setting of owner, adding the labels and annotations if we do it outside the mutate function the values won't be updated

Tested the changes with @leelavg and it works

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -782,25 +782,24 @@ func removeStorageClaimAsOwner(obj client.Object) {
 func (r *storageClientReconcile) EnsureClusterScopedResource(obj client.Object, extRes *providerpb.ExternalResource, useReplace bool) error {
 	obj.SetName(extRes.Name)
 
-	if err := json.Unmarshal(extRes.Data, obj); err != nil {
-		return fmt.Errorf("failed to unmarshal %s configuration response: %v", obj.GetName(), err)
+	mutateFunc := func() error {
+		if err := json.Unmarshal(extRes.Data, obj); err != nil {
+			return fmt.Errorf("failed to unmarshal %s configuration response: %v", obj.GetName(), err)
+		}
+		removeStorageClaimAsOwner(obj)
+		if err := r.own(obj); err != nil {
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -782,23 +782,31 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
 		err = r.createOrUpdate(obj, func() error {
+			if err := json.Unmarshal(extRes.Data, obj); err != nil {
+				return fmt.Errorf("failed to unmarshal %s configuration response: %v", obj.GetName(), err)
+			}
+			removeStorageClaimAsOwner(obj)
+			if err := r.own(obj); err != nil {
+				return fmt.Errorf("failed to own %s resource: %v", obj.GetName(), err)
+			}
+			utils.AddLabels(obj, extRes.Labels)
+			utils.AddAnnotations(obj, extRes.Annotations)
```
nb-ohad, line 809: This change creates duplicate code that need to be maintained in multiple places. As an alternative I suggest the creation of the mutation function at the same place where the original code was then passing it to `createOrUpdate` or `createOrReplace`

```suggestion
	mutateFunc := func() error { 
		if err := json.Unmarshal(extRes.Data, obj); err != nil {
			return fmt.Errorf("failed to unmarshal %s configuration response: %v", obj.GetName(), err)
		}
		removeStorageClaimAsOwner(obj)
		if err := r.own(obj); err != nil {
			return fmt.Errorf("failed to own %s resource: %v", obj.GetName(), err)
		}
		utils.AddLabels(obj, extRes.Labels)
		utils.AddAnnotations(obj, extRes.Annotations)
	}
	
	var err error
	if useReplace {
		err = utils.CreateOrReplace(r.ctx, r.Client, obj, mutateFunc)
	} else {
		err = r.createOrUpdate(obj, mutateFunc)
	}

```

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
rewantsoni: done

PR #335, opened 2025-04-05
sync between csi deletion by rook and csi creation by csi-operator

leelavg:
this is a tough find worthy of a explainer and here it goes

kubelet plugin registration mechanism:
watch a particular folder on node and for every sock creation send an rpc to get it registered and on sock deletion de-register the plugin

node driver registrar:
a sidecar in nodeplugin and on start create a sock when kubelet asks for the name gets the name from csi plugin and send to kubelet, when container is deleted this sidecar removes the sock (hostPath)

issue:
rook managed csi is running and already registered the csidrivers and on upgrade node registrar from csi-op managed csi tries to create the sock and doesn't error out as it already exists

however, due to the order of operations rook deletes it's csi and the sock in the host also gets deleted but node registrar doesn't create it again (as per design) and kubelet when trying to mount a volume on that node fails to get a handle on the csidriver as it was de-registered.

usual recommendation is restart of nodeplugin but in our case for many clusters getting upgraded it's not a good UX

fix:
wait for rook managed csi to go away before we create csi-op (driver) cr's for bringning the new csi pods

https://issues.redhat.com/browse/DFBUGS-2068

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -41,6 +41,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/version"
 	k8sYAML "k8s.io/apimachinery/pkg/util/yaml"
@@ -183,6 +184,7 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
 //+kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch
 //+kubebuilder:rbac:groups="apps",resources=deployments/finalizers,verbs=update
+//+kubebuilder:rbac:groups="apps",resources=daemonsets,verbs=get;list;watch
 //+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups="",resources=configmaps/finalizers,verbs=update
 //+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;list;watch;create;patch;update;delete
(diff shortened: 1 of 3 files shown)


leelavg:
tested for regression and also against reproducing it in a synthetic manner.

there are multiple search hits w/ `driver name not found in the list of registered CSI drivers` I looked at some and didn't see any RC mentioned but the usual restart of nodeplugin which we can't perform manually due to multiple clusters being upgraded and csi ownership change w/ current release.

https://github.com/user-attachments/assets/fd9619d2-af15-44a3-82ab-3710f7e66bd0

after above, creating a new pod fails w/ (which is similar to issue reported)
```
> k describe po -ndefault ubi9-bk-0 | grep -A10 Events
Events:
  Type     Reason                  Age                From                     Message
  ----     ------                  ----               ----                     -------
  Warning  FailedScheduling        98s                default-scheduler        0/6 nodes are available: pod has unbound immediate PersistentVolumeClaims. preemption: 0/6 nodes are available: 6 Preemption is not helpful for scheduling.
  Normal   Scheduled               96s                default-scheduler        Successfully assigned default/ubi9-bk-0 to ip-10-0-9-16.ap-southeast-1.coms frompute.internal
  Normal   SuccessfulAttachVolume  96s                attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-0ae02c35-ff4e-415e-939d-f2b95b1a48bf"
  Warning  FailedMapVolume         62s (x7 over 93s)  kubelet                  MapVolume.SetUpDevice failed for volume "pvc-0ae02c35-ff4e-415e-939d-f2b95b1a48bf" : kubernetes.io/csi: blockMapper.SetUpDevice failed to get CSI client: driver name openshift-storage.rbd.csi.ceph.com not found in the list of registered CSI drivers
```

since I wasn't able to find the RC & above error being slightly different (also my knowledge on CSI is diminishing 😅) I'd ask @Madhu-1 / @nixpanic to confirm the RC once, thanks.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -314,7 +316,57 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			name: "csi-rbdplugin-provisioner",
+		},
+		{
+			gvk:  appsv1.SchemeGroupVersion.WithKind("Deployment"),
+			name: "csi-cephfsplugin-provisioner",
+		},
+		{
+			gvk:  appsv1.SchemeGroupVersion.WithKind("Deployment"),
+			name: "csi-nfsplugin-provisioner",
+		},
```
Madhu-1, line 335: What is the reason to check the provisioners?

leelavg: We don't want half from rook and another half from csi-op

Madhu-1: Rook also creates the services for csi driver, IMO we are adding extra code that is not required for this problem.

Madhu-1: Lets followup to remove the deployments

nb-ohad:
As discussed offline, this problem is better suited to be solved by waiting on the actual socket to get free then trying to understand the same by looking at some hardcoded k8s resources.

The current approach is having an init controller  on the node plugin pod to wait until the socket is freed regardless of who is holding it. A new PR will be introduced on the [Ceph CSI Repo](https://github.com/ceph/ceph-csi-operator) to achieve this. 
At that point we will revert this change

@Madhu-1 Please provide your opinion on the above

Madhu-1:
> As discussed offline, this problem is better suited to be solved by waiting on the actual socket to get free then trying to understand the same by looking at some hardcoded k8s resources.
> 
> The current approach is having an init controller on the node plugin pod to wait until the socket is freed regardless of who is holding it. A new PR will be introduced on the [Ceph CSI Repo](https://github.com/ceph/ceph-csi-operator) to achieve this. At that point we will revert this change
> 
> @Madhu-1 Please provide your opinion on the above

we can discuss more about it, IMHO This is not something that worth to be fixed in ceph csi operator.

PR #336, opened 2025-04-07
controllers: set csidriver.spec.seLinuxMount to true

leelavg:
when client-op was created this field probably didn't exist (1.25) and when migration to csi-op is carried out only the existing fields from code was transferred.

this is a hint for Linux kernel that each volume is independently accessible and seems to be related to ReadWriteOncePod feature as well.

seLinuxMount is set to true in the yamls deployed by all ceph-csi (repo) supported drivers.

### DIFF
--- a/pkg/templates/csi.go
+++ b/pkg/templates/csi.go
@@ -114,6 +114,7 @@ var CSIOperatorConfigSpec = csiopv1a1.OperatorConfigSpec{
 			Replicas: ptr.To(int32(2)),
 		},
 		NodePlugin: &csiopv1a1.NodePluginSpec{
+			EnableSeLinuxHostMount: ptr.To(true),
 			Resources: csiopv1a1.NodePluginResourcesSpec{
 				LogRotator: &corev1.ResourceRequirements{
 					Requests: corev1.ResourceList{


leelavg:
/hold for clarification
based on the docs, I see this option fails on existing mounts on upgrade but all internal & external mode volumes have this option set to true and probably not setting this now also wrong.

if we set it now, mounted provider mode volumes would fail and need confirmation/clarification on what the correct way is from @Madhu-1 

ref: https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1710-selinux-relabeling/README.md#volume-mounting

Madhu-1:
> /hold for clarification based on the docs, I see this option fails on existing mounts on upgrade but all internal & external mode volumes have this option set to true and probably not setting this now also wrong.
> 
> if we set it now, mounted provider mode volumes would fail and need confirmation/clarification on what the correct way is from @Madhu-1
> 
> ref: https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1710-selinux-relabeling/README.md#volume-mounting

More details about it can be found here, https://github.com/rook/rook/pull/8547/files, Is it enable by default in ODF? can you please check, if its not we can drop it

nb-ohad:
As discussed offline, this problem is better suited to be solved by separating SeLinux enablement and mount configuration as part of the ceph-csi-operator's driver API. 

For now, we are going to treat this fix as a temporary workaround in order to meet some internal deadlines. Then revet this in favor of a proper fix on ceph-csi-operator when available.

Madhu-1:
Sounds good, let's get this temp fix until will fix it in csi operator

PR #338, opened 2025-04-07
controllers: use GetDesiredClientState instead of GetStorageConfig

rewantsoni:
from 4.19 we will use the new rpc call GetDesiredClientState to fetch the client info instead of GetStorageConfig

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -37,7 +37,6 @@ import (
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	quotav1 "github.com/openshift/api/quota/v1"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
-	providerpb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
 	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -779,27 +661,33 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
-func (r *storageClientReconcile) EnsureClusterScopedResource(obj client.Object, extRes *providerpb.ExternalResource, useReplace bool) error {
-	obj.SetName(extRes.Name)
+func (r *storageClientReconcile) ensureResource(obj client.Object, rawObject []byte, useReplace bool, clusterScoped bool) error {
```
nb-ohad, line 664:
```suggestion
func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject []byte, useReplace bool, clusterScoped bool) error {
```

internal/controller/storageclient_controller.go
```diff
@@ -779,27 +661,33 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	objectMeta := metav1.ObjectMeta{}
+	if err := json.Unmarshal(rawObject, &objectMeta); err != nil {
+		return err
+	}
+
+	obj.SetName(objectMeta.Name)
+	if !clusterScoped {
+		obj.SetNamespace(r.OperatorNamespace)
+	}
```
nb-ohad, line 673: This should not matter, even if you set a namespace on a cluster scooped resource the update/create should still work. There is no point in managing this manually
```suggestion
	obj.SetNamespace(r.OperatorNamespace)
```

internal/controller/storageclient_controller.go
```diff
@@ -779,27 +661,33 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
 
-func (r *storageClientReconcile) EnsureClusterScopedResource(obj client.Object, extRes *providerpb.ExternalResource, useReplace bool) error {
-	obj.SetName(extRes.Name)
+func (r *storageClientReconcile) ensureResource(obj client.Object, rawObject []byte, useReplace bool, clusterScoped bool) error {
+	objectMeta := metav1.ObjectMeta{}
+	if err := json.Unmarshal(rawObject, &objectMeta); err != nil {
+		return err
+	}
+
+	obj.SetName(objectMeta.Name)
```
nb-ohad, line 670: What is the point in setting the name if the unmarsheling will override it?

internal/controller/storageclient_controller.go
```diff
@@ -779,27 +661,33 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	if !clusterScoped {
+		obj.SetNamespace(r.OperatorNamespace)
```
nb-ohad, line 672: To be moved into the createOrUpdate / createOrReplace mutate function

internal/controller/storageclient_controller.go
```diff
@@ -779,27 +661,33 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	}
 
 	mutateFunc := func() error {
-		if err := json.Unmarshal(extRes.Data, obj); err != nil {
+		if err := json.Unmarshal(rawObject, obj); err != nil {
 			return fmt.Errorf("failed to unmarshal %s configuration response: %v", obj.GetName(), err)
 		}
```
nb-ohad, line 678: Is the intent to completely ignore any local metadata and just override everything?
Also what will happen to resources with a Status sub struct that is not a pointer?

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
rewantsoni: We need to name in order to createOrUpdate/createOrReplcace the object

internal/controller/storageclient_controller.go
```diff
@@ -357,141 +356,52 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	for _, kubeResource := range storageClientResponse.KubeResources {
 		var err error
+		typeMeta := metav1.TypeMeta{}
+		if err := json.Unmarshal(kubeResource, &typeMeta); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to unmarshal the type for the Object: %w", err)
+		}
+
 		// Create the received resources, if necessary.
-		switch eResource.Kind {
+		switch typeMeta.Kind {
```
nb-ohad, line 374: Kind is not good enough, what if the same kind belong to 2 different groups?

internal/controller/storageclient_controller.go
```diff
@@ -779,27 +661,31 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	if err := json.Unmarshal(rawObject, &objectMeta); err != nil {
+		return err
+	}
+
+	obj.SetName(objectMeta.Name)
+	obj.SetNamespace(objectMeta.Namespace)
 
 	mutateFunc := func() error {
-		if err := json.Unmarshal(extRes.Data, obj); err != nil {
+		if err := json.Unmarshal(rawObject, obj); err != nil {
```
nb-ohad, line 674: I believe we agreed to a comment here explaining the merge nature of unmarshal which is necessary for the correctness of this code

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
```diff
@@ -193,6 +205,46 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return err
+	}
+
+	if CsiCephIoV1alpha1ClientProfile, err = apiutil.GVKForObject(&csiopv1a1.ClientProfile{}, r.Scheme); err != nil {
+		return err
+	}
+
+	if CsiCephIoV1alpha1ClientProfileMapping, err = apiutil.GVKForObject(&csiopv1a1.ClientProfileMapping{}, r.Scheme); err != nil {
+		return err
+	}
```
nb-ohad, line 246: This create more complexity then solves any. 
Looking into the implementation of GVKForObject it seems that we are to run a lot of code that have multiple failure point just to acquire at runtime knowledge we have at build time. 

My suggestion is to revert this change and go back to having each case based on building a GVK on the flay using the GroupVersion and adding the kind based on the type

nb-ohad (changes requested):

PR #339, opened 2025-04-08
controllers: read subscription channel from GetDesiredClientState rpc

leelavg:
move updating the channel annotation from heartbeat job to controller as the channel is already part of state hash and we reconcile when the hash changes.

Depends on #338


leelavg:
rebased.

PR #341, opened 2025-04-09
Updated golang.org/x/oauth2 version from v0.24.0 to v0.27.0

ezio-auditore:
Fixes: https://issues.redhat.com/browse/DFBUGS-1993
CVE-2025-22868


leelavg:
The diff in the libs is only in comments, is that the only change?

ezio-auditore:
> The diff in the libs is only in comments, is that the only change

Yes

leelavg (approved):
we don't use the vuln jws directly but bumping the version to be safer.

PR #344, opened 2025-04-14
bundle: add priorityClassName to csv deployments

iamniting:
Signed-off-by: Nitin Goyal <nigoyal@redhat.com>

Bug: https://issues.redhat.com/browse/DFBUGS-394

### DIFF
--- a/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
@@ -491,6 +491,7 @@ spec:
                 volumeMounts:
                 - mountPath: /etc/tls/private
                   name: webhook-cert-secret
+              priorityClassName: system-cluster-critical
               securityContext:
                 runAsNonRoot: true
               serviceAccountName: ocs-client-operator-controller-manager
@@ -553,6 +554,7 @@ spec:
                   name: ocs-client-operator-console-nginx-log
                 - mountPath: /var/lib/nginx/tmp
                   name: ocs-client-operator-console-nginx-tmp
+              priorityClassName: system-cluster-critical
               securityContext:
                 runAsNonRoot: true
               serviceAccountName: ocs-client-operator-controller-manager
--- a/config/console/console_init.yaml
+++ b/config/console/console_init.yaml
@@ -66,3 +66,4 @@ spec:
         key: node.ocs.openshift.io/storage
         operator: Equal
         value: "true"
+      priorityClassName: system-cluster-critical
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -86,3 +86,4 @@ spec:
         key: node.ocs.openshift.io/storage
         operator: Equal
         value: "true"
+      priorityClassName: system-cluster-critical


leelavg:
Can you confirm that this priority is less than more important components like ceph daemons?

iamniting:
> Can you confirm that this priority is less than more important components like ceph daemons?

In those components we are using `openshift-user-critical`

```
$ oc get priorityclasses.scheduling.k8s.io 
NAME                      VALUE        GLOBAL-DEFAULT   AGE     PREEMPTIONPOLICY
openshift-user-critical   1000000000   false            4h55m   PreemptLowerPriority
system-cluster-critical   2000000000   false            4h56m   PreemptLowerPriority
system-node-critical      2000001000   false            4h56m   PreemptLowerPriority
```

leelavg:
> In those components we are using openshift-user-critical
- I don't think that's good, unless in a rare scenarios I would never want operator to replace a ceph daemon like OSD.


/hold

Will go through the BZ once, pls excuse.

@Madhu-1 fyi...

leelavg:
1. correct: https://github.com/red-hat-storage/ocs-operator/blob/3f085cdaabfed24137d186e785134f83c6b9ed1e/controllers/storagecluster/cephcluster.go#L537-L541 (mon, mgr, osd)
2. wrong: https://github.com/red-hat-storage/ocs-operator/blob/3f085cdaabfed24137d186e785134f83c6b9ed1e/controllers/storagecluster/cephnfs.go#L38-L40 (nfs)
3. wrong: https://github.com/red-hat-storage/ocs-operator/blob/3f085cdaabfed24137d186e785134f83c6b9ed1e/controllers/storagecluster/cephfilesystem.go#L44-L45 (mds)

could you also correct (2) & (3) to be `system-node-critical`?

iamniting:
> 1. correct: https://github.com/red-hat-storage/ocs-operator/blob/3f085cdaabfed24137d186e785134f83c6b9ed1e/controllers/storagecluster/cephcluster.go#L537-L541 (mon, mgr, osd)
> 2. wrong: https://github.com/red-hat-storage/ocs-operator/blob/3f085cdaabfed24137d186e785134f83c6b9ed1e/controllers/storagecluster/cephnfs.go#L38-L40 (nfs)
> 3. wrong: https://github.com/red-hat-storage/ocs-operator/blob/3f085cdaabfed24137d186e785134f83c6b9ed1e/controllers/storagecluster/cephfilesystem.go#L44-L45 (mds)
> 
> could you also correct (2) & (3) to be `system-node-critical`?

Could we talk about this in the JIRA so we can decide based on the discussion?

leelavg:
I already added a comment to Jira.

Madhu-1:
> > In those components we are using openshift-user-critical
> 
> * I don't think that's good, unless in a rare scenarios I would never want operator to replace a ceph daemon like OSD.
> 
> /hold
> 
> Will go through the BZ once, pls excuse.
> 
> @Madhu-1 fyi...

@iamniting what is the reason to choose a priorityclass for these pods? AFAIK this are not critical pods like CSI or the ceph core components. we should not consider all the pods has highest priority when they are not highest priority, if We really need to have a priority we can choose some medium priority for operator pods?

iamniting:
> > > In those components we are using openshift-user-critical
> > 
> > 
> > 
> > * I don't think that's good, unless in a rare scenarios I would never want operator to replace a ceph daemon like OSD.
> > 
> > /hold
> > Will go through the BZ once, pls excuse.
> > @Madhu-1 fyi...
> 
> @iamniting what is the reason to choose a priorityclass for these pods? AFAIK this are not critical pods like CSI or the ceph core components. we should not consider all the pods has highest priority when they are not highest priority, if We really need to have a priority we can choose some medium priority for operator pods?

Can you pls take a look at JIRA, It will give you the good idea.

PR #346, opened 2025-04-15
controllers: disable csi ctrlPlugin hostNetwork for internal client

rohan47:
Disable CSI controller plugin host network when the client operator is running on the local storage cluster and the storage cluster is not on host network

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -64,11 +64,12 @@ const (
 	operatorConfigMapName = "ocs-client-operator-config"
 	// ClusterVersionName is the name of the ClusterVersion object in the
 	// openshift cluster.
-	clusterVersionName     = "version"
-	manageNoobaaSubKey     = "manageNoobaaSubscription"
-	subscriptionLabelKey   = "managed-by"
-	subscriptionLabelValue = "webhook.subscription.ocs.openshift.io"
-	generateRbdOMapInfoKey = "generateRbdOMapInfo"
+	clusterVersionName                 = "version"
+	manageNoobaaSubKey                 = "manageNoobaaSubscription"
+	subscriptionLabelKey               = "managed-by"
+	subscriptionLabelValue             = "webhook.subscription.ocs.openshift.io"
+	generateRbdOMapInfoKey             = "generateRbdOMapInfo"
+	useHostNetworkForCsiControllersKey = "useHostNetworkForCsiControllers"
 
 	operatorConfigMapFinalizer = "ocs-client-operator.ocs.openshift.io/storageused"
 	subPackageIndexName        = "index:subscriptionPackage"
@@ -410,7 +411,8 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		if c.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
 			csiOperatorConfig.Spec.DriverSpecDefaults.SnapshotPolicy = csiopv1a1.VolumeGroupSnapshotPolicy
 		}
-
+		csiCtrlPluginHostNetwork, _ := strconv.ParseBool(c.operatorConfigMap.Data[useHostNetworkForCsiControllersKey])
+		csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.HostNetwork = ptr.To(csiCtrlPluginHostNetwork)
 		return nil
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile csi operator config: %v", err)
--- a/pkg/templates/csi.go
+++ b/pkg/templates/csi.go
@@ -111,7 +111,8 @@ var CSIOperatorConfigSpec = csiopv1a1.OperatorConfigSpec{
 					},
 				},
 			},
-			Replicas: ptr.To(int32(2)),
+			Replicas:    ptr.To(int32(2)),
+			HostNetwork: ptr.To(true),
 		},
 		NodePlugin: &csiopv1a1.NodePluginSpec{
 			EnableSeLinuxHostMount: ptr.To(true),


pkg/templates/csi.go
```diff
@@ -111,7 +111,8 @@ var CSIOperatorConfigSpec = csiopv1a1.OperatorConfigSpec{
 					},
 				},
 			},
-			Replicas: ptr.To(int32(2)),
+			Replicas:    ptr.To(int32(2)),
+			HostNetwork: ptr.To(true),
```
Madhu-1, line 115: Why this is set to true here if its controlled in the operatorConfig CR?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -410,6 +411,9 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		if c.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
 			csiOperatorConfig.Spec.DriverSpecDefaults.SnapshotPolicy = csiopv1a1.VolumeGroupSnapshotPolicy
 		}
+		if val, ok := c.operatorConfigMap.Data[csiCtrlPluginHostNetworkKey]; ok && val == strconv.FormatBool(false) {
+			csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.HostNetwork = ptr.To(false)
+		}
```
Madhu-1, line 416:
```suggestion
if val, ok := c.operatorConfigMap.Data[csiCtrlPluginHostNetworkKey]; ok && val == strconv.FormatBool(true) {
			csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.HostNetwork = ptr.To(true)
}
```

Lets the controller plugin run with pod network by default and set it to true only if required?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -64,11 +64,12 @@ const (
...
-	manageNoobaaSubKey     = "manageNoobaaSubscription"
-	subscriptionLabelKey   = "managed-by"
-	subscriptionLabelValue = "webhook.subscription.ocs.openshift.io"
-	generateRbdOMapInfoKey = "generateRbdOMapInfo"
+	clusterVersionName          = "version"
+	manageNoobaaSubKey          = "manageNoobaaSubscription"
+	subscriptionLabelKey        = "managed-by"
+	subscriptionLabelValue      = "webhook.subscription.ocs.openshift.io"
+	generateRbdOMapInfoKey      = "generateRbdOMapInfo"
+	csiCtrlPluginHostNetworkKey = "csi-ctrlplugin-host-network"
```
nb-ohad, line 72: Please see the following comment on the ocs-operator PR: https://github.com/red-hat-storage/ocs-operator/pull/3167#discussion_r2048383998

internal/controller/operatorconfigmap_controller.go
```diff
@@ -410,6 +411,9 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		if c.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
 			csiOperatorConfig.Spec.DriverSpecDefaults.SnapshotPolicy = csiopv1a1.VolumeGroupSnapshotPolicy
 		}
+		if val, ok := c.operatorConfigMap.Data[csiCtrlPluginHostNetworkKey]; ok && val == strconv.FormatBool(false) {
```
nb-ohad, line 414: - There is no point in checking the existence of the key if we only care about a true value as string 
- Please use a case insensitive compare operation, preferably convert the key into a bool and not the other way around: 
```suggestion
		useHostNetworkForCsiControllers := strconv.ParseBool(c.operatorConfigMap.Data[csiCtrlPluginHostNetworkKey])
		if !useHostNetworkForCsiControllers {
```

nb-ohad (changes requested):

pkg/templates/csi.go
nb-ohad: @Madhu-1 
We only have automation control over the config map for the internal client (via ocs-operator). Not for any other client. This means we must assume `true` as a default value then have ocs-opeartor turn this off for "internal mode" like deployments.

In any case, this is a temporary workaround and we do have a better long term solution coming soon

Madhu-1: what will happen for the mode where we have no internal but only external ceph cluster?

nb-ohad: @Madhu-1 What are you expecting it to be in that case? with or without setting the controller to use host network?
Please remember that the only reason we are setting the controller to use host network is when we apply data separation technics without Multus. I don't think we support it with external mode only deployments

internal/controller/operatorconfigmap_controller.go
```diff
@@ -64,11 +64,12 @@ const (
...
-	manageNoobaaSubKey     = "manageNoobaaSubscription"
-	subscriptionLabelKey   = "managed-by"
-	subscriptionLabelValue = "webhook.subscription.ocs.openshift.io"
-	generateRbdOMapInfoKey = "generateRbdOMapInfo"
+	clusterVersionName          = "version"
+	manageNoobaaSubKey          = "manageNoobaaSubscription"
+	subscriptionLabelKey        = "managed-by"
+	subscriptionLabelValue      = "webhook.subscription.ocs.openshift.io"
+	generateRbdOMapInfoKey      = "generateRbdOMapInfo"
+	csiCtrlPluginHostNetworkKey = "useHostNetworkForCsiControllers"
```
nb-ohad, line 72: Please have the var name match the value
```suggestion
	useHostNetworkForCsiControllersKey = "useHostNetworkForCsiControllers"
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -410,7 +411,11 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		if c.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
 			csiOperatorConfig.Spec.DriverSpecDefaults.SnapshotPolicy = csiopv1a1.VolumeGroupSnapshotPolicy
 		}
-
+		csiCtrlPluginHostNetwork, err := strconv.ParseBool(c.operatorConfigMap.Data[csiCtrlPluginHostNetworkKey])
+		if err != nil {
+			return fmt.Errorf("failed to parse %s: %v", csiCtrlPluginHostNetworkKey, err)
+		}
```
nb-ohad, line 417: There is no point in handling parsing errors here. We can just treat them as false
```suggestion
		csiCtrlPluginHostNetwork, _ := strconv.ParseBool(c.operatorConfigMap.Data[csiCtrlPluginHostNetworkKey])

```

nb-ohad (changes requested):

pkg/templates/csi.go
Madhu-1: For external mode, we should still be using pod networking only (this is the current behavior today)

internal/controller/operatorconfigmap_controller.go
rohan47: We want to run the controller plugin with Host Network always and only use pod network in case of internal mode scenario where Storage Cluster is on pod network

PR #349, opened 2025-04-24
restrict manager to watch selective  namespaces

ShravaniVangur:
Currently, the manager watches for resources in all namespaces. This ends up causing a increase in memory usage which hinders the regular deployment and activity of the operator, leading to the operator getting killed due to insufficient memory. 
The current proposed fix ensures that the manager caches resources from namespaces where the operator is deployed or that are specified in  the `WATCH_NAMESPACE` env variable.

Fixes: [DFBUGS-2056](https://issues.redhat.com/browse/DFBUGS-2056)  

### Testing performed: 
-------------------------------------------------------------------------------------------------------------------------
After deploying the `ocs-client-operator`, various resources such as `ConfigMaps` and `Secrets` were created in a namespace different from the operator's own namespace.

* Behaviour without the fix: It was observed that the memory usage of the `ocs-client-operator` continuously increased over time, eventually leading to the pod being terminated with `exit code 137 (OOMKilled)` due to exceeding its `256Mi` memory limit.
```
Every 5.0s: oc adm top pod | grep ocs-client-o... Shravanis-MacBook-Pro.  
ocs-client-operator-controller-manager-548d95968b-wpz9c                 1m                            236Mi
```

*  Behaviour with the fix: After applying the fix, the memory usage of the `ocs-client-operator` remained stable, and the pod continued to run without issues.
```
Every 5.0s: oc adm top pod | grep ocs-client-o... Shravanis-MacBook-Pro.  
ocs-client-operator-controller-manager-548d95968b-wpz9c                 2m                            30Mi
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	apiv1alpha1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/internal/controller"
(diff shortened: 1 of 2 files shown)


hack/make-bundle-vars.mk
```diff
@@ -107,6 +107,7 @@ endif
 OPERATOR_NAMEPREFIX ?= ocs-client-operator-
 OPERATOR_NAMESPACE ?= $(OPERATOR_NAMEPREFIX)system
 OPERATOR_CATALOGSOURCE ?= oco-catalogsource
+WATCH_NAMESPACE ?= ""
```
leelavg, line 110: Can introduce this when used in a bundle.

cmd/main.go
```diff
@@ -186,6 +187,19 @@ func main() {
 		})
 	}
 
+	defaultNamespaces := map[string]cache.Config{}
+	operatorNamespace := utils.GetOperatorNamespace()
+	defaultNamespaces[operatorNamespace] = cache.Config{}
+
+	watchNamespace := utils.GetWatchNamespace()
+	if watchNamespace == "" {
+		setupLog.Info("$WATCH_NAMESPACE has not been set. Manager will only watch for resources in the operator deployed namespace.")
```
leelavg, line 196:
```suggestion
		setupLog.Info("No value for env WATCH_NAMESPACE is set. Manager will only watch for resources in the operator deployed namespace.")
```

leelavg (approved):
/hold

awaiting the comment wrt testing.

PR #356, opened 2025-04-29
controllers: not send client status when client marked for deletion

rohan47:
When a storage client is marked for deletion, do not report the status


service/status-report/main.go
```diff
@@ -117,9 +118,12 @@ func main() {
...
 	setStorageQuotaUtilizationRatio(ctx, cl, status)
-	statusResponse, err := providerClient.ReportStatus(ctx, storageClient.Status.ConsumerID, status)
-	if err != nil {
-		klog.Exitf("Failed to report status of storageClient %v: %v", storageClient.Status.ConsumerID, err)
+	statusResponse := &providerpb.ReportStatusResponse{}
+	if storageClient.GetDeletionTimestamp().IsZero() {
+		statusResponse, err = providerClient.ReportStatus(ctx, storageClient.Status.ConsumerID, status)
+		if err != nil {
+			klog.Exitf("Failed to report status of storageClient %v: %v", storageClient.Status.ConsumerID, err)
+		}
```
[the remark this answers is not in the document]
rohan47: Does it make sense to break out of code after getting storageclient if the storageclient is marked for deletion

nb-ohad, line 126: Why does this change make sense? I would think that if we are stuck on deletion we should still report heartbeat info. The reporting will stop the min that the job will be gone which should happen as part of the clean up we should do when the client CR is getting deleted

rohan47: Agreed, 
changing it to returning error on Offboarding failure

rewantsoni:
@rohan47 There can be a scenario like this:
```
t0: offboarding rpc is called, and the client section is set to nil
t1: the status reporter sends the status and updates the status of the consumer. 
t2: storageClient is deleted
```

If such a scenario occurs, offboarding is not successful. The consumer status will still have client information on it

rohan47:
> @rohan47 There can be a scenario like this:
> 
> ```
> t0: offboarding rpc is called, and the client section is set to nil
> t1: the status reporter sends the status and updates the status of the consumer. 
> t2: storageClient is deleted
> ```
> 
> If such a scenario occurs, offboarding is not successful. The consumer status will still have client information on it

But as @nb-ohad  mentioned we need to update the status till the deletion is complete, I cant think of any solution we can implement on the client side to takle the above scenario.
But can we put some marking (annotation for example) on the storageConsumer to signify that the consumer is offboarded and based on that marking we remove the storageClient status

internal/controller/storageclient_controller.go
```diff
@@ -447,8 +447,17 @@ func (r *storageClientReconcile) deletionPhase(externalClusterClient *providerCl
 		}
 	}
 
+	job := &batchv1.CronJob{}
```
leelavg, line 450: here var `job` is incorrect, either `cronjob` or `statusReporter` would be good.

internal/controller/storageclient_controller.go
```diff
@@ -447,8 +447,17 @@ func (r *storageClientReconcile) deletionPhase(externalClusterClient *providerCl
...
+	job.Name = fmt.Sprintf("storageclient-%s-status-reporter", utils.GetMD5Hash(r.storageClient.Name)[:16])
+	job.Namespace = r.OperatorNamespace
+	r.log.Info("deleting status reporter job", "StatusReporter", job.Name)
+	if err := r.Client.Delete(r.ctx, job); err != nil {
```
leelavg, line 454: better ignore not found error if there is already an event scheduled you'll always fail here.

internal/controller/storageclient_controller.go
rohan47: Changed

leelavg:
/cherrypick release-4.19
via DFBUGS-2298

adding here to have a link created in Jira.

internal/controller/storageclient_controller.go
```diff
@@ -447,8 +447,17 @@ func (r *storageClientReconcile) deletionPhase(externalClusterClient *providerCl
 		}
 	}
 
+	statusReporterJob := &batchv1.CronJob{}
+	statusReporterJob.Name = fmt.Sprintf("storageclient-%s-status-reporter", utils.GetMD5Hash(r.storageClient.Name)[:16])
+	statusReporterJob.Namespace = r.OperatorNamespace
+	r.log.Info("deleting status reporter job", "StatusReporter", statusReporterJob.Name)
+	if err := r.Client.Delete(r.ctx, statusReporterJob); err != nil && !kerrors.IsNotFound(err) {
```
leelavg, line 454:
```suggestion
	if err := r.Client.Delete(r.ctx, statusReporterJob); client.IgnoreNotFound(err) != nil {
```

rohan47: Changed

service/status-report/main.go
```diff
@@ -104,6 +104,10 @@ func main() {
 		klog.Exitf("Failed to get storageClient %q/%q: %v", storageClient.Namespace, storageClient.Name, err)
 	}
 
+	if !storageClient.GetDeletionTimestamp().IsZero() {
+		klog.Exitf("StorageClient %q/%q is being deleted", storageClient.Namespace, storageClient.Name)
```
nb-ohad, line 108: Storage client is cluster scoped and does not have a namespace
```suggestion
		klog.Exitf("Skipping report, StorageClient %q is being deleted", storageClient.Name)
```

nb-ohad (changes requested):

service/status-report/main.go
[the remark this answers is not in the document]
rohan47: Yes, klog.Exitf() calls OsExit(1) after logging

internal/controller/storageclient_controller.go
```diff
@@ -502,7 +502,7 @@ func (r *storageClientReconcile) onboardConsumer(externalClusterClient *provider
 // offboardConsumer makes an API call to the external storage provider cluster for offboarding
 func (r *storageClientReconcile) offboardConsumer(externalClusterClient *providerClient.OCSProviderClient) (reconcile.Result, error) {
 	if _, err := externalClusterClient.OffboardConsumer(r.ctx, r.storageClient.Status.ConsumerID); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to offboard consumer: %v", err)
+		return reconcile.Result{Requeue: true}, fmt.Errorf("failed to offboard consumer: %v", err)
```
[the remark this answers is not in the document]
rohan47: The caller function is not returning on error, so returning non empty reconcile.Result
```
if res, err := r.offboardConsumer(externalClusterClient); err != nil {
		r.log.Error(err, "Offboarding in progress.")
	} else if !res.IsZero() {
		// result is not empty
		return res, nil
	}
```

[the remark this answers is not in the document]
nb-ohad: So you are fixing the wrong thing. Why does the caller function does not handle the error? what is the reason? After you answer these questions you could make the proper changes

service/status-report/main.go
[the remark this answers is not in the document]
nb-ohad: Existing with a none 0 value would be considered a Exist in error state. That is the exact reason I am asking the question

internal/controller/storageclient_controller.go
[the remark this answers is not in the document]
rohan47: What I can think of  is WRT old flow, If the reconcile is requeued and the storageConsumer is already deleted, the code will get stuck unless reconcile.Result is not nil.
This is not required anymore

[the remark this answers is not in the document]
rohan47: In the old flow where offboarding was deleting storageConsumer, the error was not being handled because for example the offboarding is triggered and storageConsumer is deleted and requeue because of error in removing finalizer, while offboarding is called again in requeue, it will error out if the error was getting returned

internal/controller/storageclient_controller.go
```diff
@@ -447,12 +447,10 @@ func (r *storageClientReconcile) deletionPhase(externalClusterClient *providerCl
 		}
 	}
 
-	if res, err := r.offboardConsumer(externalClusterClient); err != nil {
-		r.log.Error(err, "Offboarding in progress.")
-	} else if !res.IsZero() {
-		// result is not empty
-		return res, nil
+	if _, err := r.offboardConsumer(externalClusterClient); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to offboard consumer for storageclient %v: %v", r.storageClient.Name, err)
```
nb-ohad, line 451: The function signature implies that a non-error requeue result might be returned from the `r.offboardConsumer` function. In that case, you cannot just ignore it and continue to remove the finalizer. 

There are 2 options here:
If the function can and might return a non-error requeue result then you need to hand it (similar to the original code). If it can't, the the Result object should be removed from the function signature.

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
rohan47: removed the result object from r.offboardConsumer()

PR #358, opened 2025-04-30
controllers: allow ctrlplugin to be upgraded on 2 node clusters

leelavg:
the default deployment strategy of csi ctrlplugin requires 3 nodes during an upgrade, however the plugin already runs w/ learderelection providing fault tolerance and this commit relaxes the node requirements to 2.

### DIFF
--- a/pkg/templates/csi.go
+++ b/pkg/templates/csi.go
@@ -4,10 +4,13 @@ import (
 	"fmt"
 
 	csiopv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
+
 	secv1 "github.com/openshift/api/security/v1"
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


leelavg:
already tested earlier.

leelavg:
/cherrypick release-4.19
via DFBUGS-1876

leelavg:
/hold

needs a discussion whether this should be made default or allow overrides for specific setups.

Madhu-1:
> /hold
> 
> needs a discussion whether this should be made default or allow overrides for specific setups.

@leelavg we dont need to provide override and we can keep this as default behavior (today this is the default behavior in Rook)

leelavg:
@nb-ohad didn't approve the change, mentioned we need the redundancy.

leelavg:
dropped, will only be a doc change.

PR #360, opened 2025-05-06
Remove master node tolerations added to CSI node plugin

malayparida2000:
The master node tolerations were added so that the node plugin pods could be scheduled on HCP nodes. But with convergence now CSI is deployed by client operator always. This means node plugin pods are getting schedules on master nodes even in normal odf clusters. We will implement another way to specify any kind of tolerations for the node plugin pods.


rewantsoni:
LGTM

leelavg:
we should send this all the way back to 4.17, isn't it?

malayparida2000:
Till 4.18 Rook was deploying the CSI pods right? if yes then the master node toleratiosn would not be present on Internal mode clusters.

leelavg:
agreed, my question was for provider mode, rethinking again, adding to the docs would suffice even for older provider clusters.

PR #362, opened 2025-05-06
controllers: don't replace the obj if it can be updated

rewantsoni:
Use createOrUpdate first, and if it fails with Forbidden then use createOrReplace

Fixes: https://issues.redhat.com/browse/DFBUGS-2289

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -373,28 +373,28 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 		// Create the received resources, if necessary.
 		switch typeMeta.GroupVersionKind() {
 		case quotav1.SchemeGroupVersion.WithKind("ClusterResourceQuota"):
-			err = r.reconcileResource(&quotav1.ClusterResourceQuota{}, kubeResource, false)
+			err = r.reconcileResource(&quotav1.ClusterResourceQuota{}, kubeResource)
 		case csiopv1a1.GroupVersion.WithKind("CephConnection"):
-			err = r.reconcileResource(&csiopv1a1.CephConnection{}, kubeResource, false)
+			err = r.reconcileResource(&csiopv1a1.CephConnection{}, kubeResource)
 		case csiopv1a1.GroupVersion.WithKind("ClientProfileMapping"):
-			err = r.reconcileResource(&csiopv1a1.ClientProfileMapping{}, kubeResource, false)
+			err = r.reconcileResource(&csiopv1a1.ClientProfileMapping{}, kubeResource)
 		case corev1.SchemeGroupVersion.WithKind("Secret"):
-			err = r.reconcileResource(&corev1.Secret{}, kubeResource, false)
+			err = r.reconcileResource(&corev1.Secret{}, kubeResource)
 		case nbv1.SchemeGroupVersion.WithKind("Noobaa"):
-			err = r.reconcileResource(&nbv1.NooBaa{}, kubeResource, false)
+			err = r.reconcileResource(&nbv1.NooBaa{}, kubeResource)
 		case storagev1.SchemeGroupVersion.WithKind("StorageClass"):
-			err = r.reconcileResource(&storagev1.StorageClass{}, kubeResource, true)
+			err = r.reconcileResource(&storagev1.StorageClass{}, kubeResource)
 		case snapapi.SchemeGroupVersion.WithKind("VolumeSnapshotClass"):
-			err = r.reconcileResource(&snapapi.VolumeSnapshotClass{}, kubeResource, true)
+			err = r.reconcileResource(&snapapi.VolumeSnapshotClass{}, kubeResource)
 		case groupsnapapi.SchemeGroupVersion.WithKind("VolumeGroupSnapshotClass"):
 			if val, _ := r.crdsBeingWatched.Load(VolumeGroupSnapshotClassCrdName); !val.(bool) {
 				continue
 			}
-			err = r.reconcileResource(&groupsnapapi.VolumeGroupSnapshotClass{}, kubeResource, true)
+			err = r.reconcileResource(&groupsnapapi.VolumeGroupSnapshotClass{}, kubeResource)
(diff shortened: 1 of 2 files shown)


leelavg:
had an offline discussion, the intention seems to be automatically choosing replace over update if a certain kind of error occurs, yet to check if this error can be taken as source of truth for replacement.

other point is, w/ this change we are pushing the build time convenience/flexibility of choosing what resources are replaceable to runtime dynamic behavior based on a heuristic.

internal/controller/storageclient_controller.go
```diff
@@ -694,10 +694,9 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject
 	}
 
 	var err error
-	if useReplace {
+	_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, obj, mutateFunc)
+	if utils.IsForbiddenError(err) {
 		err = utils.CreateOrReplace(r.ctx, r.Client, obj, mutateFunc)
```
nb-ohad, line 699: Does not make a lot of sense to run createOrReplace. We know create will fail. Is there a reason we cannot just run the replace code?

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
```diff
@@ -692,16 +692,33 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject
...
-	if err != nil {
+		if err := r.Client.Delete(r.ctx, obj); client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
+		}
+
+		// k8s doesn't allow us to create objects when resourceVersion is set, as we are DeepCopying the
+		// object, the resource version also gets copied, hence we need to set it to empty before creating it
+		obj.SetResourceVersion("")
+		if err := r.Client.Create(r.ctx, obj); err != nil {
+			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
```
nb-ohad, line 717: You should not be doing delete and create inside CreateOrUpdate mutate func. CreateOrUpdate will do its own Update operation when the mutate func finishes

pkg/utils/k8sutils.go
```diff
@@ -141,54 +137,19 @@ func GetMD5Hash(text string) string {
...
+		if statusErr.ErrStatus.Details.Causes[i].Type == metav1.CauseTypeForbidden {
+			return true
+		}
 	}
-	return nil
-}
-
-func GetClusterResourceQuotaName(name string) string {
-	return fmt.Sprintf("storage-client-%s-resourceqouta", GetMD5Hash(name))
+	return false
```
nb-ohad, line 154: - For a function returning a boolean, please prefer only one return of true value and only one return of false value.
- Please use a non panicing down cast 
```suggestion
	statusErr, ok := err.(*errors.StatusError)
	if ok {
		for i := range statusErr.ErrStatus.Details.Causes {
			if statusErr.ErrStatus.Details.Causes[i].Type == metav1.CauseTypeForbidden {
				return true
			}
		}
	}
	return false
```

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
rewantsoni: The Create and delete is outside the CreateOrUpdate mutate function

internal/controller/storageclient_controller.go
```diff
@@ -692,16 +692,33 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject
...
-		err = utils.CreateOrReplace(r.ctx, r.Client, obj, mutateFunc)
-	} else {
-		_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, obj, mutateFunc)
-	}
+	_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, obj, mutateFunc)
+	if utils.IsForbiddenError(err) {
+		key := client.ObjectKeyFromObject(obj)
+		if err := r.Client.Get(r.ctx, key, obj); client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
+		}
```
nb-ohad, line 701: There is no reason to try to read it again. It was already read and mutated in the create of update function

internal/controller/storageclient_controller.go
```diff
@@ -692,16 +692,33 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject
...
+		if err := mutateFunc(); err != nil {
+			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
+		}
```
nb-ohad, line 704: Why are we running the mutate function again? what would have changed?

internal/controller/storageclient_controller.go
```diff
@@ -692,16 +692,22 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject
...
+		if err := r.Client.Delete(r.ctx, obj); client.IgnoreNotFound(err) != nil {
+			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
+		}
 
-	if err != nil {
+		// k8s doesn't allow us to create objects when resourceVersion is set, as we are DeepCopying the
+		// object, the resource version also gets copied, hence we need to set it to empty before creating it
+		obj.SetResourceVersion("")
+		if err := r.Client.Create(r.ctx, obj); err != nil {
+			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
```
nb-ohad, line 706:
```suggestion
			return fmt.Errorf("failed to replace %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetNamespacedName(), err)
```

nb-ohad (changes requested):

PR #364, opened 2025-05-09
controllers: remove resources not sent by provider

leelavg:
storageclient owns all the resources sent by provider and when some resource is dropped from the list we need to clean those stale resources

in every reconcile we will remove the diff from the provider sent & client owned resources.

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -46,6 +46,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -84,6 +85,8 @@ var (
 		templates.CephFsDriverName,
 	}
 )
+
+type gvkToObjectKeys = map[schema.GroupVersionKind][]types.NamespacedName
 
 // StorageClientReconciler reconciles a StorageClient object
 type StorageClientReconciler struct {
@@ -368,6 +371,7 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 
 	r.storageClient.Status.InMaintenanceMode = storageClientResponse.MaintenanceMode
 
+	resourceMapping := gvkToObjectKeys{}
 	for _, kubeResource := range storageClientResponse.KubeResources {
 		var err error
 		typeMeta := &metav1.TypeMeta{}
@@ -378,28 +382,28 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 		// Create the received resources, if necessary.
 		switch typeMeta.GroupVersionKind() {
 		case quotav1.SchemeGroupVersion.WithKind("ClusterResourceQuota"):
-			err = r.reconcileResource(&quotav1.ClusterResourceQuota{}, kubeResource, false)
+			err = r.reconcileResource(&quotav1.ClusterResourceQuota{}, kubeResource, false, resourceMapping)
 		case csiopv1a1.GroupVersion.WithKind("CephConnection"):
-			err = r.reconcileResource(&csiopv1a1.CephConnection{}, kubeResource, false)
+			err = r.reconcileResource(&csiopv1a1.CephConnection{}, kubeResource, false, resourceMapping)
 		case csiopv1a1.GroupVersion.WithKind("ClientProfileMapping"):
-			err = r.reconcileResource(&csiopv1a1.ClientProfileMapping{}, kubeResource, false)
+			err = r.reconcileResource(&csiopv1a1.ClientProfileMapping{}, kubeResource, false, resourceMapping)
 		case corev1.SchemeGroupVersion.WithKind("Secret"):
-			err = r.reconcileResource(&corev1.Secret{}, kubeResource, false)
+			err = r.reconcileResource(&corev1.Secret{}, kubeResource, false, resourceMapping)
 		case nbv1.SchemeGroupVersion.WithKind("Noobaa"):
-			err = r.reconcileResource(&nbv1.NooBaa{}, kubeResource, false)
+			err = r.reconcileResource(&nbv1.NooBaa{}, kubeResource, false, resourceMapping)
 		case storagev1.SchemeGroupVersion.WithKind("StorageClass"):
-			err = r.reconcileResource(&storagev1.StorageClass{}, kubeResource, true)
+			err = r.reconcileResource(&storagev1.StorageClass{}, kubeResource, true, resourceMapping)
 		case snapapi.SchemeGroupVersion.WithKind("VolumeSnapshotClass"):
-			err = r.reconcileResource(&snapapi.VolumeSnapshotClass{}, kubeResource, true)
+			err = r.reconcileResource(&snapapi.VolumeSnapshotClass{}, kubeResource, true, resourceMapping)
 		case groupsnapapi.SchemeGroupVersion.WithKind("VolumeGroupSnapshotClass"):
 			if val, _ := r.crdsBeingWatched.Load(VolumeGroupSnapshotClassCrdName); !val.(bool) {
 				continue
 			}
-			err = r.reconcileResource(&groupsnapapi.VolumeGroupSnapshotClass{}, kubeResource, true)
+			err = r.reconcileResource(&groupsnapapi.VolumeGroupSnapshotClass{}, kubeResource, true, resourceMapping)
 		case replicationv1a1.GroupVersion.WithKind("VolumeReplicationClass"):
-			err = r.reconcileResource(&replicationv1a1.VolumeReplicationClass{}, kubeResource, true)
+			err = r.reconcileResource(&replicationv1a1.VolumeReplicationClass{}, kubeResource, true, resourceMapping)
 		case csiopv1a1.GroupVersion.WithKind("ClientProfile"):
-			err = r.reconcileResource(&csiopv1a1.ClientProfile{}, kubeResource, false)
+			err = r.reconcileResource(&csiopv1a1.ClientProfile{}, kubeResource, false, resourceMapping)
 		}
 		if err != nil {
 			return reconcile.Result{}, err
@@ -419,7 +423,34 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 		}
 	}
 
+	if err := r.deleteStaleResources(resourceMapping); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to delete stale resources: %v", err)
+	}
+
 	return reconcile.Result{}, nil
+}
+
+func (r *storageClientReconcile) deleteStaleResources(resourceMapping gvkToObjectKeys) error {
+	for gvk, providerSentKeys := range resourceMapping {
+		existingObjList := &metav1.PartialObjectMetadataList{}
+		existingObjList.SetGroupVersionKind(gvk)
+		if err := r.list(existingObjList); err != nil {
+			return err
+		}
+		retainKeys := make(map[types.NamespacedName]bool, len(providerSentKeys))
+		for idx := range providerSentKeys {
+			retainKeys[providerSentKeys[idx]] = true
+		}
+		for idx := range existingObjList.Items {
+			obj := &existingObjList.Items[idx]
+			if metav1.IsControlledBy(obj, &r.storageClient) && !retainKeys[client.ObjectKeyFromObject(obj)] {
+				if err := r.Delete(r.ctx, obj); client.IgnoreNotFound(err) != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
 }
 
 func (r *storageClientReconcile) deletionPhase(externalClusterClient *providerClient.OCSProviderClient) (ctrl.Result, error) {
@@ -653,7 +684,7 @@ func (r *storageClientReconcile) update(obj client.Object, opts ...client.Update
 }
 
 func (r *storageClientReconcile) own(dependent metav1.Object) error {
-	return controllerutil.SetOwnerReference(&r.storageClient, dependent, r.Scheme)
+	return controllerutil.SetControllerReference(&r.storageClient, dependent, r.Scheme)
 }
 
 func removeStorageClaimAsOwner(obj client.Object) {
@@ -665,7 +696,7 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
-func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject []byte, useReplace bool) error {
+func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject []byte, useReplace bool, resourceMapping gvkToObjectKeys) error {
 	objectMeta := &metav1.PartialObjectMetadata{}
 	if err := json.Unmarshal(rawObject, objectMeta); err != nil {
 		return err
@@ -700,6 +731,9 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject 
 	if err != nil {
 		return fmt.Errorf("failed to create or update %v %v: %v", obj.GetObjectKind().GroupVersionKind(), obj.GetName(), err)
 	}
+
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	resourceMapping[gvk] = append(resourceMapping[gvk], client.ObjectKeyFromObject(obj))
 	return nil
 }
 


leelavg:
/hold
for testing

leelavg:
/cherrypick release-4.19
via DFBUGS-2291

internal/controller/storageclient_controller.go
```diff
@@ -368,6 +371,7 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 
 	r.storageClient.Status.InMaintenanceMode = storageClientResponse.MaintenanceMode
 
+	resourceMapping := gvkToObjectKeys{}
 	for _, kubeResource := range storageClientResponse.KubeResources {
```
nb-ohad, line 375: What is the benefit of sending the resourceMapping into each case? Why can't you just append to the list in a central place above or below the switch case?

internal/controller/storageclient_controller.go
```diff
@@ -419,9 +423,36 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+		}
+		retainKeys := make(map[types.NamespacedName]bool, len(providerSentKeys))
+		for idx := range providerSentKeys {
+			retainKeys[providerSentKeys[idx]] = true
+		}
+		for idx := range existingObjList.Items {
+			obj := &existingObjList.Items[idx]
+			if metav1.IsControlledBy(obj, &r.storageClient) && !retainKeys[client.ObjectKeyFromObject(obj)] {
+				if err := r.Delete(r.ctx, obj); client.IgnoreNotFound(err) != nil {
+					return err
```
nb-ohad, line 448: There is not reason to stop deleting resources just because one resource failed. Doing might create a situation where we are stack on a single resource that cannot be deleted for one reason or another

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
nb-ohad, line 448: This logic is dangerous, you are deleting anything that you can find of a type. Please limit what you can find based on an OwnerRef (or a label if not possible)

leelavg:
/unhold
testing
```
> cat /tmp/temp-secret.yaml | tee /dev/tty | k create -f -
apiVersion: v1
data:
[...]
kind: Secret
metadata:
  creationTimestamp: "2025-04-30T12:26:35Z"
  name: cephfs-provisioner-temp
  namespace: openshift-storage
  ownerReferences:
  - apiVersion: ocs.openshift.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: StorageClient
    name: ocs-storagecluster
    uid: 3af778cb-4cca-415d-9767-9b3ef856b5f3
  resourceVersion: "40599122"
  uid: f00114cf-d522-4b69-8b45-591b6f3d1fe0
type: Opaque
secret/cephfs-provisioner-temp created

> k get secret cephfs-provisioner-temp
NAME                      TYPE     DATA   AGE
cephfs-provisioner-temp   Opaque   5      18s

> k scale deploy/ocs-client-operator-controller-manager --replicas=1
deployment.apps/ocs-client-operator-controller-manager scaled

--- a resource which isn't sent by provider but owned by storageclient is deleted
> k get secret cephfs-provisioner-temp -w
Error from server (NotFound): secrets "cephfs-provisioner-temp" not found
```

internal/controller/storageclient_controller.go
leelavg: other options are 
1. unmarshal the object meta twice, at start or end of the switch
2. unmarshal the object meta once and send it to `reconcileResources`

internal/controller/storageclient_controller.go
leelavg: ok, thought of it but didn't do as we could be accumulating errors, nevertheless, will update.

leelavg: > you are deleting anything that you can find of a type.
- `metav1.IsControlledBy(obj, &r.storageClient)` do you mean something else?

nb-ohad: True, resolving

internal/controller/storageclient_controller.go
leelavg: done.

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go
```diff
@@ -368,42 +372,57 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
 		}
 
+		var resourceObj client.Object = nil
+		shouldReplace := false
 		// Create the received resources, if necessary.
 		switch typeMeta.GroupVersionKind() {
 		case quotav1.SchemeGroupVersion.WithKind("ClusterResourceQuota"):
-			err = r.reconcileResource(&quotav1.ClusterResourceQuota{}, kubeResource, false)
+			resourceObj = &quotav1.ClusterResourceQuota{}
+			shouldReplace = false
```
nb-ohad, line 389: The default is already false, there is no point to set it explicitly for false values
```suggestion
```

internal/controller/storageclient_controller.go
```diff
@@ -419,9 +438,37 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+		for idx := range existingObjList.Items {
+			obj := &existingObjList.Items[idx]
+			if metav1.IsControlledBy(obj, &r.storageClient) && !retainKeys[client.ObjectKeyFromObject(obj)] {
+				if err := r.Delete(r.ctx, obj); client.IgnoreNotFound(err) != nil {
+					multierr.AppendInto(&combinedErr, err)
+				}
+			}
+		}
+	}
+	return combinedErr
```
nb-ohad, line 469: Doesn't an empty combined error is still an error?

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
leelavg: Yes, I thought of it but still made it to be explicit incase the outer code moves or changes meaning.

Anyways, since we already know there is another pr removing this, will make the change.

internal/controller/storageclient_controller.go
leelavg: As zero value of `error` is `nil` I see no issue here (taking into account call site as well).

internal/controller/storageclient_controller.go
```diff
@@ -419,9 +438,37 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	if err := r.deleteStaleResources(resourceMapping); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to delete stale resources: %v", err)
+	}
+
 	return reconcile.Result{}, nil
 }
 
+func (r *storageClientReconcile) deleteStaleResources(resourceMapping gvkToObjectKeys) error {
+	var combinedErr error
+	for gvk, providerSentKeys := range resourceMapping {
```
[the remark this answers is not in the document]
leelavg: Ok, this PR doesn't solve it, need to go back to drawing board. Will that be a blocker for the merge?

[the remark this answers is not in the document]
nb-ohad: Does it make sense to merge this if we need to redesign this?

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go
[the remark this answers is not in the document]
leelavg: done.

leelavg:
rested.

leelavg:
/hold

not move to controller ref from owner ref?

internal/controller/storageclient_controller.go
```diff
@@ -85,6 +87,8 @@ var (
 	}
 )
 
+type gvkToRawresources = map[string][][]byte
```
nb-ohad, line 90:
```suggestion
type gvkToResourceBytes = map[string][][]byte
```

internal/controller/storageclient_controller.go
```diff
@@ -665,6 +710,36 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
+func (r *storageClientReconcile) reconcileResourceByGVK(
+	gvk schema.GroupVersionKind,
+	obj client.Object,
+	resourceMapping gvkToRawresources,
```
nb-ohad, line 716: You don't need the entire map, just the list of resource bytes your are going to reconcile

internal/controller/storageclient_controller.go
```diff
@@ -665,6 +710,36 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+) error {
+	resources := resourceMapping[gvk.String()]
+	retainKeys := make(map[types.NamespacedName]bool, len(resources))
+	for idx := range resources {
+		if err := r.reconcileResource(obj, resources[idx]); err != nil {
+			return err
+		}
+		retainKeys[client.ObjectKeyFromObject(obj)] = true
+	}
```
nb-ohad, line 725: Is there a reason not to use a combined error approach here as well?

internal/controller/storageclient_controller.go
```diff
@@ -665,6 +710,36 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	var combinedErr error
+	for idx := range existingObjList.Items {
+		obj := &existingObjList.Items[idx]
+		if metav1.IsControlledBy(obj, &r.storageClient) && !retainKeys[client.ObjectKeyFromObject(obj)] {
+			if err := r.Delete(r.ctx, obj); client.IgnoreNotFound(err) != nil {
+				multierr.AppendInto(&combinedErr, err)
+			}
+		}
+	}
```
nb-ohad, line 739: If we are already running over the entire list and you already have a need for a map, why not just have one loop running on all of the resources and select the action based of the existence of bytes for that object in the server response?

internal/controller/storageclient_controller.go
```diff
@@ -368,43 +372,84 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to reconcile %s: %v", "Secret", err)
+	}
+	if err := r.reconcileResourceByGVK(
+		nbv1.SchemeGroupVersion.WithKind("NooBaa"),
+		&nbv1.NooBaa{},
+		resourceMapping,
+	); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to reconcile %s: %v", "Noobaa", err)
+	}
```
nb-ohad, line 416: There is no real reason to have these cases hardcoded one after another. We do want the code to dictate which types we should be processing, but we could have it in the form of a master fixed list of just client.Object instances where the type info is already filled in.

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
```diff
@@ -83,8 +87,21 @@ var (
 		templates.RBDDriverName,
 		templates.CephFsDriverName,
 	}
+	knownReconcileObjects = []client.Object{
```
nb-ohad, line 90:
```suggestion
	kindsToReconcile = []client.Object{
```

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +386,31 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
-				continue
-			}
-			err = r.reconcileResource(&groupsnapapi.VolumeGroupSnapshotClass{}, kubeResource)
-		case replicationv1a1.GroupVersion.WithKind("VolumeReplicationClass"):
-			err = r.reconcileResource(&replicationv1a1.VolumeReplicationClass{}, kubeResource)
-		case csiopv1a1.GroupVersion.WithKind("ClientProfile"):
-			err = r.reconcileResource(&csiopv1a1.ClientProfile{}, kubeResource)
+	resourceMapping := gvkToResourceBytes{}
+	for idx := range storageClientResponse.KubeObjects {
+		resource := storageClientResponse.KubeObjects[idx]
```
nb-ohad, line 391: It seems like the items in the slice are pointers, in that case you can just take the pointer itself as the second argument to the for loop. Also `kubeObj` is a better fitting name in this context
```suggestion
	for idx, kubeObj := range storageClientResponse.KubeObjects {
```

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +386,31 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
-		case groupsnapapi.SchemeGroupVersion.WithKind("VolumeGroupSnapshotClass"):
-			if val, _ := r.crdsBeingWatched.Load(VolumeGroupSnapshotClassCrdName); !val.(bool) {
-				continue
-			}
-			err = r.reconcileResource(&groupsnapapi.VolumeGroupSnapshotClass{}, kubeResource)
-		case replicationv1a1.GroupVersion.WithKind("VolumeReplicationClass"):
-			err = r.reconcileResource(&replicationv1a1.VolumeReplicationClass{}, kubeResource)
-		case csiopv1a1.GroupVersion.WithKind("ClientProfile"):
-			err = r.reconcileResource(&csiopv1a1.ClientProfile{}, kubeResource)
+	resourceMapping := gvkToResourceBytes{}
```
nb-ohad, line 389: I think we settled on kube object as our new RPC terminology 
```suggestion
	kubeObjMapping := gvkToKubeObjBytes{}
```

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +386,31 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	for idx := range storageClientResponse.KubeObjects {
+		resource := storageClientResponse.KubeObjects[idx]
+		if resource != nil && resource.GroupVersionKind != "" {
+			resourceMapping[resource.GroupVersionKind] = append(resourceMapping[resource.GroupVersionKind], resource.Bytes)
 		}
+	}
+	dynamicReconcileObjects := []client.Object{}
+	if val, _ := r.crdsBeingWatched.Load(VolumeGroupSnapshotClassCrdName); val.(bool) {
+		dynamicReconcileObjects = append(dynamicReconcileObjects, &groupsnapapi.VolumeGroupSnapshotClass{})
+	}
```
nb-ohad, line 399: This is known at compile time. The information should be encoded outside of the reconcile function

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +386,31 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	var combinedErr error
+	for obj := range ObjectsIterator(knownReconcileObjects, dynamicReconcileObjects) {
```
nb-ohad, line 401: Why not a simple range without iterators?

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +386,31 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+		gvk, err := apiutil.GVKForObject(obj, r.Scheme)
 		if err != nil {
```
nb-ohad, line 403: You are handling all cases inside the `if` and `else` sections. This will allow you to keep the `err` var local
```suggestion
		if gvk, err := apiutil.GVKForObject(obj, r.Scheme); err != nil {
```

internal/controller/storageclient_controller.go
```diff
@@ -697,6 +702,43 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
+func (r *storageClientReconcile) reconcileResourcesByGVK(
+	resources [][]byte,
```
nb-ohad, line 706: I think we decided to define a type name for this complex type

internal/controller/storageclient_controller.go
```diff
@@ -697,6 +702,43 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	obj client.Object,
+	combinedErr error,
+) {
+	retainKeys := make(map[types.NamespacedName]bool, len(resources))
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	for idx := range resources {
+		// object supplied to reconcile mutates it, we either need to send
+		// fresh copy or zero out fields on the mutated object, doing the former
+		// as we don't know the fields that needs to be zeroed for every concrete object
+		var newObj client.Object = reflect.New(reflect.TypeOf(obj).Elem()).Interface().(client.Object)
```
nb-ohad, line 716: Please separate the steps into their own lines. This kind of chaining is unreadable

internal/controller/storageclient_controller.go
```diff
@@ -697,6 +702,43 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+		newObj.GetObjectKind().SetGroupVersionKind(gvk)
```
nb-ohad, line 717: I don't think you need to do that

internal/controller/storageclient_controller.go
```diff
@@ -791,3 +833,15 @@ func (r *storageClientReconcile) getOperatorVersion() (string, error) {
...
+	return func(yield func(T) bool) {
+		for _, slice := range slices {
+			for idx := range slice {
+				if !yield(slice[idx]) {
+					return
+				}
+			}
+		}
+	}
+}
```
nb-ohad, line 847: This complexity is definitely not needed to accomplish what is expected from this PR. Please remove

internal/controller/storageclient_controller.go
```diff
@@ -697,6 +702,43 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
+func (r *storageClientReconcile) reconcileResourcesByGVK(
+	resources [][]byte,
+	obj client.Object,
+	combinedErr error,
+) {
+	retainKeys := make(map[types.NamespacedName]bool, len(resources))
```
nb-ohad, line 710: May I suggest a  more fitting name
```suggestion
	reconciledObjects := make(map[types.NamespacedName]bool, len(resources))
```

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
leelavg: we only have a type alias for a map and this part is only the value of the map which isn't aliased further.

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go
```diff
@@ -697,6 +702,43 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+
+	existingObjList := &metav1.PartialObjectMetadataList{}
+	existingObjList.SetGroupVersionKind(gvk)
+	if err := r.list(existingObjList); err != nil {
+		multierr.AppendInto(&combinedErr, err)
+		r.log.Error(err, "failed to list resources")
+	}
+	for idx := range existingObjList.Items {
+		obj := &existingObjList.Items[idx]
+		if metav1.IsControlledBy(obj, &r.storageClient) && !retainKeys[client.ObjectKeyFromObject(obj)] {
```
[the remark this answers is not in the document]

internal/controller/storageclient_controller.go
nb-ohad: There shouldn't be 2 slices in the first place.

internal/controller/storageclient_controller.go
nb-ohad: preferably have an alias for the type []Bytes, this will allow us to have meaningful names for both the map and the slices

internal/controller/storageclient_controller.go
nb-ohad:
```
objType := reflect.TypeOf(obj).Elem()
copyAsAny := reflect.New(type),Interface()
copy := untypedCopy.(client.Object)
```

leelavg:
/unhold

verified functionality w/ all the changes.

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +393,43 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
-			err = r.reconcileResource(&corev1.Secret{}, kubeResource)
-		case nbv1.SchemeGroupVersion.WithKind("Noobaa"):
-			err = r.reconcileResource(&nbv1.NooBaa{}, kubeResource)
-		case storagev1.SchemeGroupVersion.WithKind("StorageClass"):
-			err = r.reconcileResource(&storagev1.StorageClass{}, kubeResource)
-		case snapapi.SchemeGroupVersion.WithKind("VolumeSnapshotClass"):
-			err = r.reconcileResource(&snapapi.VolumeSnapshotClass{}, kubeResource)
-		case groupsnapapi.SchemeGroupVersion.WithKind("VolumeGroupSnapshotClass"):
-			if val, _ := r.crdsBeingWatched.Load(VolumeGroupSnapshotClassCrdName); !val.(bool) {
+	requestsPerGvk := map[string][]reconcileRequest{}
```
nb-ohad, line 396:
```suggestion
	requestsByGvk := map[string][]reconcileRequest{}
```

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +393,43 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	for _, kubeObj := range storageClientResponse.KubeObjects {
+		if kubeObj != nil {
```
nb-ohad, line 398: Let's loos one indentation
```suggestion
		if kubeObj == nil {
			continue
		}
		
```

internal/controller/storageclient_controller.go
nb-ohad, line 396: Why not `map[string]*metav1.PartialObjectMetadata{}`? 
You are already unmarshlaing and creating these

internal/controller/storageclient_controller.go
```diff
@@ -369,43 +393,43 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+					bytes:     kubeObj.Bytes,
+				},
+			)
+		}
+	}
+	var combinedErr error
+	for idx := range kindsToReconcile {
+		kind := &kindsToReconcile[idx]
+		if kind.requiresCrd != "" {
+			if val, _ := r.crdsBeingWatched.Load(kind.requiresCrd); !val.(bool) {
```
nb-ohad, line 418: Can't we handle that by analyzing the get return error instead?
I am asking because the rest mapper should be the source of truth

internal/controller/storageclient_controller.go
```diff
@@ -83,8 +86,29 @@ var (
 		templates.RBDDriverName,
 		templates.CephFsDriverName,
 	}
+	kindsToReconcile = []struct {
+		object      client.Object
```
nb-ohad, line 90: Object is a bad name because we don't use these as 'objects', more as types

internal/controller/storageclient_controller.go
```diff
@@ -697,15 +721,48 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
-func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject []byte) error {
-	objectMeta := &metav1.PartialObjectMetadata{}
-	if err := json.Unmarshal(rawObject, objectMeta); err != nil {
-		return err
+func (r *storageClientReconcile) reconcileResourcesByGVK(
+	obj client.Object,
```
nb-ohad, line 725: Object is a bad name because we don't use these as 'objects', more as types

internal/controller/storageclient_controller.go
```diff
@@ -697,15 +721,48 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+	reqs []reconcileRequest,
```
nb-ohad, line 726: These are not requests, these are the desired states for some of the object that we need to reconcile. 
Please remember that a result of delete is still considered a reconcile operation

nb-ohad (changes requested):

internal/controller/storageclient_controller.go

internal/controller/storageclient_controller.go
leelavg: it is stored in this syncmap only after doing a get on the clustter

internal/controller/storageclient_controller.go
leelavg: changed to `objectType`

internal/controller/storageclient_controller.go
leelavg: changed to `desiredStates`

internal/controller/storageclient_controller.go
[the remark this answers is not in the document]
leelavg: resolving as this is implemented

internal/controller/storageclient_controller.go
```diff
@@ -370,42 +392,33 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+			return reconcile.Result{}, fmt.Errorf("failed to unmarshal metadata for the Object: %w", err)
+		}
+		gvk := objectMeta.GroupVersionKind().String()
+		kubeObjectsByGvk[gvk] = append(
+			kubeObjectsByGvk[gvk],
+			desiredKubeObject{
+				NamespacedName: types.NamespacedName{
+					Name:      objectMeta.GetName(),
+					Namespace: objectMeta.GetNamespace(),
+				},
```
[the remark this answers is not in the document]

internal/controller/storageclient_controller.go
```diff
@@ -698,21 +711,61 @@ func removeStorageClaimAsOwner(obj client.Object) {
 	}
 }
 
-func (r *storageClientReconcile) reconcileResource(obj client.Object, rawObject []byte) error {
-	objectMeta := &metav1.PartialObjectMetadata{}
-	if err := json.Unmarshal(rawObject, objectMeta); err != nil {
-		return err
+func (r *storageClientReconcile) reconcileResourcesByGVK(
+	objectType client.Object,
```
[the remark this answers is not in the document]

internal/controller/storageclient_controller.go
```diff
@@ -698,21 +711,61 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+
+	objectsToReconcile := desiredObjects[gvk.String()]
+	reconciledObjects := make(map[types.NamespacedName]bool, len(objectsToReconcile))
+	for idx := range objectsToReconcile {
+		// object supplied to reconcile mutates it, we either need to send
+		// fresh copy or zero out fields on the mutated object, doing the former
+		// as we don't know the fields that needs to be zeroed for every concrete object
+		objType := reflect.TypeOf(objectType).Elem()
+		copyAsAny := reflect.New(objType).Interface()
+		objCopy := copyAsAny.(client.Object)
```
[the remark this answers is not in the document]

internal/controller/storageclient_controller.go
```diff
@@ -698,21 +711,61 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+
+		desiredObject := objectsToReconcile[idx]
```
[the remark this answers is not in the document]

internal/controller/storageclient_controller.go
```diff
@@ -698,21 +711,61 @@ func removeStorageClaimAsOwner(obj client.Object) {
...
+				multierr.AppendInto(&combinedErr, err)
+				r.log.Error(err, "failed to delete object", "Name", client.ObjectKeyFromObject(obj))
+			}
+		}
+	}
+}
 
-	obj.SetName(objectMeta.Name)
-	obj.SetNamespace(objectMeta.Namespace)
+func (r *storageClientReconcile) reconcileResource(obj client.Object, desiredObject desiredKubeObject) error {
```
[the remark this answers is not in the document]

PR #367, opened 2025-05-16
add private VolumeGroupSnapshotClass Api in ocs-client-op

ShravaniVangur:
Updating the ocs-client-operator to import the private `volumegroupsnapshotclass` api, enabling access to and utilization of the private fork of the sidecar available at https://github.com/red-hat-storage/external-snapshotter.

//Waiting for the controllers to be available before merging.
//Pending- dependency update


cmd/main.go
```diff
@@ -45,6 +45,7 @@ import (
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	ramenv1alpha1 "github.com/ramendr/ramen/api/v1alpha1"
+	odfgroupsnapapi "github.com/red-hat-storage/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
```
leelavg, line 48: better have the version in the import, after that it'll be too long, then can be curtailed to `odfgsapiv1b1`?

internal/controller/storageclient_controller.go
```diff
@@ -273,7 +290,7 @@ func (r *storageClientReconcile) reconcileDynamicWatches() error {
 	if err := r.controller.Watch(
 		source.Kind(
 			r.cache,
-			client.Object(&groupsnapapi.VolumeGroupSnapshotContent{}),
+			client.Object(obj),
```
leelavg, line 293:
```suggestion
			obj,
```
this should work since the received obj is already a `client.Object`

internal/controller/storageclient_controller.go
leelavg, line 1: the flow is kinda splitting the logic in multiple places, let's say if we want to remove reconcile of private api then we need to remove some top level functions and also some code in the existing functions which is easy to miss.

can we make the changes in a way to treat the apis are entirely different and during removal it'd be easy + we don't need to branch multiple times. wdyt?

ShravaniVangur: The goal of this approach was to avoid unnecessary duplication and have minimal changes, which led to splitting the logic across different parts of the code. I do agree that handling the APIs differently could simplify their removal. If that works then we could keep the flow intact.

internal/controller/storageclient_controller.go
```diff
@@ -286,6 +304,7 @@ func (r *storageClientReconcile) reconcileDynamicWatches() error {
 	}
 	// CRD doesn't exist in the cluster
 	if crd.UID == "" {
+		r.log.Info("VolumeGroupSnapshotClass CRD is not available", "crdName", VolumeGroupSnapshotClassCrdName)
```
leelavg, line 307: this'll add a negative log line in every reconcile and not sure why this is required.

internal/controller/storageclient_controller.go
```diff
@@ -329,6 +348,63 @@ func (r *storageClientReconcile) reconcileDynamicWatches() error {
...
+
+	crd := &metav1.PartialObjectMetadata{}
+	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+	crd.Name = OdfVolumeGroupSnapshotClassCrdName
+	if err := r.get(crd); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+	// CRD doesn't exist in the cluster
+	if crd.UID == "" {
+		r.log.Info("ODF-VolumeGroupSnapshotClass CRD is not available", "crdName", OdfVolumeGroupSnapshotClassCrdName)
```
leelavg, line 364: doesn't add much value. if the crd doesn't exist we anyways get a log when we try to reconcile resource of that kind already.

PR #371, opened 2025-05-20
Avoid updating empty channel on the storage client annotations

malayparida2000:
In case of errors & unexpected situations the subscription channel returned  can be empty. We should not update the empty channel on the storageclient.


leelavg:
/hold

for @nb-ohad review as he approved the server side.

PR #378, opened 2025-05-27
controllers: always set metadata for csi nodeplugin pods

leelavg:
setting the option makes csi to add metadata to the volumes created in ceph and until now this wasn't made default as there is a possibility that external ceph couldn't support it.

however, based on available docs this is supported all the way from RHCEPH 5 and current maintenance ODF version 4.15 supports RHCEPH 6.


leelavg:
/cherrypick release-4.19
via DFBUGS-2395

PR #382, opened 2025-05-28
controllers: delete StorageClaims post upgrade

ShravaniVangur:
After the ownership of resources is transferred from `StorageClaim`, it remains in the cluster. This PR ensures that these `StorageClaims` are properly cleaned up.

Resolves: [DFBUGS-2625](https://issues.redhat.com/browse/DFBUGS-2625)

Test Results for `oc get storageclaim`:
- **Currently**:
```
NAME                          STORAGETYPE   STORAGEPROFILE   STORAGECLIENTNAME    PHASE
ocs-storagecluster-ceph-rbd   block                          ocs-storagecluster   Configuring
ocs-storagecluster-cephfs     sharedfile                     ocs-storagecluster   Configuring
```
- **With the fix**:
```
No resources found
```

### DIFF
--- a/internal/controller/maintenancemode_controller.go
+++ b/internal/controller/maintenancemode_controller.go
@@ -3,6 +3,8 @@ package controller
 import (
 	"context"
 	"fmt"
+	"slices"
+
 	"github.com/go-logr/logr"
 	ramenv1alpha1 "github.com/ramendr/ramen/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
@@ -19,7 +21,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"slices"
 )
 
 const (
@@ -65,7 +66,7 @@ func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 //+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients;storageclaims,verbs=get;list;watch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get;list;watch
 
 func (r *MaintenanceModeReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
 	r.ctx = ctx
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -213,6 +213,7 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotcontents,verbs=get;list;watch
 //+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotclasses,verbs=get;list;watch;create;delete;update
 //+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotcontents,verbs=get;list;watch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclaims,verbs=get;list;watch;delete;patch
 
 func (r *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	handler := storageClientReconcile{StorageClientReconciler: r}
@@ -421,7 +422,38 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 		}
 	}
 
+	// Deleting pre 4.19 built-in storage claims
+	if err = r.deleteStorageClaims(fmt.Sprintf("%s-ceph-rbd", r.storageClient.Name)); err != nil {
+		return reconcile.Result{}, err
+	}
+	if err = r.deleteStorageClaims(fmt.Sprintf("%s-cephfs", r.storageClient.Name)); err != nil {
+		return reconcile.Result{}, err
+	}
 	return reconcile.Result{}, nil
+}
+
+func (r *storageClientReconcile) deleteStorageClaims(claimName string) error {
+	storageClaim := &metav1.PartialObjectMetadata{}
+	storageClaim.SetGroupVersionKind(v1alpha1.GroupVersion.WithKind("StorageClaim"))
+	storageClaim.Name = claimName
+
+	if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); client.IgnoreNotFound(err) != nil {
+		return fmt.Errorf("Failed to get StorageClaim %q: %w", claimName, err)
+	}
+	if storageClaim.UID != "" {
+		if err := r.Client.Delete(r.ctx, storageClaim); err != nil {
+			return fmt.Errorf("Failed to delete StorageClaim %q: %v", claimName, err)
+		}
+
+		original := storageClaim.DeepCopy()
+		if controllerutil.RemoveFinalizer(storageClaim, "storageclaim.ocs.openshift.io") {
+			if err := r.Client.Patch(r.ctx, storageClaim, client.MergeFrom(original)); err != nil {
+				return fmt.Errorf("Failed to patch StorageClaim %q: %v", claimName, err)
+			}
+			r.log.Info("StorageClaim finalizer removed", "name", claimName)
+		}
+	}
+	return nil
 }
 
 func (r *storageClientReconcile) deletionPhase(externalClusterClient *providerClient.OCSProviderClient) (ctrl.Result, error) {
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -217,10 +217,11 @@ rules:
   - ocs.openshift.io
   resources:
   - storageclaims
-  - storageclients
   verbs:
+  - delete
   - get
   - list
+  - patch
   - watch
 - apiGroups:
   - ocs.openshift.io


internal/controller/storageclient_controller.go
```diff
@@ -213,6 +214,7 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotcontents,verbs=get;list;watch
 //+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotclasses,verbs=get;list;watch;create;delete;update
 //+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotcontents,verbs=get;list;watch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclaims,verbs=get;list;watch;delete
```
leelavg, line 217: CI passed but there is no bundle update for this rbac, can you remove the duplicate rbac from other controllers?

internal/controller/storageclient_controller.go
```diff
@@ -239,6 +241,23 @@ func (r *storageClientReconcile) reconcile(ctx context.Context, req ctrl.Request
 
 	result, reconcileErr := r.reconcilePhases()
 
+	storageClaimList := &metav1.PartialObjectMetadataList{}
+	storageClaimList.SetGroupVersionKind(v1alpha1.GroupVersion.WithKind("StorageClaimList"))
+	if err := r.list(storageClaimList); err != nil {
+		r.log.Info("No StorageClaims remainings")
```
leelavg, line 247: we can't deduce a failed api call on a list as no items present and here you should requeue w/ an error.

internal/controller/storageclient_controller.go
```diff
@@ -239,6 +241,23 @@ func (r *storageClientReconcile) reconcile(ctx context.Context, req ctrl.Request
...
+	}
+
+	for _, storageClaim := range storageClaimList.Items {
+		if err := r.Client.Delete(r.ctx, &storageClaim); client.IgnoreNotFound(err) != nil {
```
leelavg, line 251: you are iterating on a list of objects and no need to ignore the not found error, if it indeed gets deleted b/n the list and delete let it requeue w/ an error and it'll be removed eventually.

internal/controller/storageclient_controller.go
```diff
@@ -239,6 +241,23 @@ func (r *storageClientReconcile) reconcile(ctx context.Context, req ctrl.Request
...
+			r.log.Error(err, "Failed to delete StorageClaim", "name", storageClaim.Name)
```
leelavg, line 252: this is a failed api call in the items of a list, pls requeue on error.

internal/controller/storageclient_controller.go
```diff
@@ -239,6 +241,23 @@ func (r *storageClientReconcile) reconcile(ctx context.Context, req ctrl.Request
...
+		}
+
+		controllerutil.RemoveFinalizer(&storageClaim, StorageClaimFinalizer)
```
leelavg, line 255: can you check if this returns a bool? if yes, update only if we actually removed the finalizer on the im-mem obj.

internal/controller/storageclient_controller.go
```diff
@@ -239,6 +241,23 @@ func (r *storageClientReconcile) reconcile(ctx context.Context, req ctrl.Request
...
+		if err := r.Client.Update(r.ctx, &storageClaim); err != nil {
```
rewantsoni, line 256: Let's use patch here, update will fail as this is partial meta

rewantsoni (changes requested):

internal/controller/maintenancemode_controller.go
```diff
@@ -65,7 +66,7 @@ func (r *MaintenanceModeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 //+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=ramendr.openshift.io,resources=maintenancemodes/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients;storageclaims,verbs=get;list;watch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get;list;watch
```
nb-ohad, line 69: Why do we have this addition in this controller? the controller code is not getting changed in this PR

internal/controller/storageclient_controller.go
```diff
@@ -76,6 +76,8 @@ const (
 	vgscClusterIDIndexName = "index:volumeGroupSnapshotContentCSIDriver"
 
 	VolumeGroupSnapshotClassCrdName = "volumegroupsnapshotclasses.groupsnapshot.storage.k8s.io"
+
+	storageClaimFinalizer = "storageclaim.ocs.openshift.io"
```
nb-ohad, line 80: I don't think there is a point in having this here as a constant where the entire code is going away on the next release. I would prefer to just have everything local to a single function.

internal/controller/storageclient_controller.go
```diff
@@ -421,9 +424,35 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+
+		original := &storageClaim
+		if controllerutil.RemoveFinalizer(&storageClaim, storageClaimFinalizer) {
+			if err := r.Client.Patch(r.ctx, &storageClaim, client.MergeFrom(original)); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
nb-ohad, line 454: We should not be deleting arbitrary storage claims, the side effect might be data lost for our customers. The 4.18 storage client owned 2 specific storage claims and our responsibility is to just clean up after these 2 specific ones

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
```diff
@@ -213,6 +214,7 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotcontents,verbs=get;list;watch
 //+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotclasses,verbs=get;list;watch;create;delete;update
 //+kubebuilder:rbac:groups=groupsnapshot.storage.k8s.io,resources=volumegroupsnapshotcontents,verbs=get;list;watch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclaims,verbs=get;list;watch;delete;update;patch
```
nb-ohad, line 217: Is there a reason why we need both update and patch here?

internal/controller/storageclient_controller.go
```diff
@@ -421,9 +423,51 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 		}
 	}
 
+	//Once the owner references are removed from resources, StorageClaim is deleted
+	if err = r.deleteStorageClaims(); err != nil {
+		return reconcile.Result{}, err
+	}
```
nb-ohad, line 429: Is there a reason the clean up is at the end of the reconcile phases? Doesn't it mean that we will not clean up unless we managed to go through an entire successful reconcile? Wouldn't it make more sense to have this cleanup at the start of the reconcile?

internal/controller/storageclient_controller.go
```diff
@@ -421,9 +423,51 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
 }
 
+func (r *storageClientReconcile) deleteStorageClaims() error {
+	storageClaimFinalizer := "storageclaim.ocs.openshift.io"
+	expectedStorageClaimNames := []string{
+		fmt.Sprintf("%s-ceph-rbd", r.storageClient.Name),
+		fmt.Sprintf("%s-cephfs", r.storageClient.Name),
+	}
+
+	for _, claimName := range expectedStorageClaimNames {
```
nb-ohad, line 440: We have 2 very specific claims? Why do we need a list and a loop to clean these?
The strightforward approach is the just have a `deleteStorageClaim` function and call it twice directly from `reconcilePhases`. Something like

```
if err := deleteStorageClaim(fmt.Sprintf("%s-ceph-rbd", r.storageClient.Name)); err != nil {

}
if err

File: internal/controller/storageclient_controller.go [hunk 13]
```diff
@@ -421,9 +423,51 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+		})
+		storageClaim.Name = claimName
+
+		if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); err != nil {
+			if kerrors.IsNotFound(err) {
+				r.log.Info("StorageClaim already deleted", "name", claimName)
+				continue
+			}
+			return fmt.Errorf("failed to get StorageClaim %q: %w", claimName, err)
+		}
```
nb-ohad, line 455: Always prefer non-nested error handling from more specific to less specific
```
		if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); errors.IsNotFound(err) {
			// do something
		} err != nil {
			// do something else
		}
		
		// No error case
```

internal/controller/storageclient_controller.go
```diff
@@ -421,9 +423,51 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
...
+	for _, claimName := range expectedStorageClaimNames {
+		storageClaim := &metav1.PartialObjectMetadata{}
+		storageClaim.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   "ocs.openshift.io",
+			Version: "v1alpha1",
+			Kind:    "StorageClaim",
+		})
+		storageClaim.Name = claimName
+
+		if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); err != nil {
```
nb-ohad, line 449: The get should go after the delete. If there was no finalizer the get might just return a NotFound and we don't need to do anything else after that point

internal/controller/storageclient_controller.go
nb-ohad, line 440: We have 2 very specific claims? Why do we need a list and a loop to clean these?
The strightforward approach is the just have a `deleteStorageClaim` function and call it twice directly from `reconcilePhases`. Something like

```
if err := deleteStorageClaim(fmt.Sprintf("%s-ceph-rbd", r.storageClient.Name)); err != nil {
    ...
}
if err := deleteStorageClaim(fmt.Sprintf("%s-cephfs", r.storageClient.Name)); err != nil {
    ...
}

nb-ohad (changes requested):

leelavg:
LGTM.

PR #383, opened 2025-05-28
controllers: approve all installplans ocs-client-operator namespace

leelavg:
when the subscription approval is marked as manual we missed to approve the dependent installplans which forces user to approve them manually.

with the fix if the top-level operator, ie, ocs-client-operator is asked to update we automatically update the dependents.

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -72,6 +72,7 @@ const (
 	clusterVersionName                 = "version"
 	manageNoobaaSubKey                 = "manageNoobaaSubscription"
 	disableVersionChecksKey            = "disableVersionChecks"
+	disableInstallPlanAutoApprovalKey  = "disableInstallPlanAutoApproval"
 	subscriptionLabelKey               = "managed-by"
 	subscriptionLabelValue             = "webhook.subscription.ocs.openshift.io"
 	generateRbdOMapInfoKey             = "generateRbdOMapInfo"
@@ -184,6 +185,18 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.OnlyMetadata,
 		).
 		Watches(&opv1a1.Subscription{}, enqueueConfigMapRequest, subscriptionPredicates).
+		Watches(
+			&opv1a1.InstallPlan{},
+			enqueueConfigMapRequest,
+			builder.WithPredicates(
+				utils.EventTypePredicate(
+					true,
+					false,
+					false,
+					false,
+				),
+			),
+		).
 		Watches(&admrv1.ValidatingWebhookConfiguration{}, enqueueConfigMapRequest, webhookPredicates).
 		Watches(&v1alpha1.StorageClient{}, enqueueConfigMapRequest, builder.WithPredicates(predicate.AnnotationChangedPredicate{}))
 
@@ -213,6 +226,7 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=*
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=subscriptions,verbs=get;list;watch;update;delete
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=installplans,verbs=get;list;watch;patch
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=delete;list
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=operatorconfigs,verbs=get;list;update;create;watch;delete
@@ -342,6 +356,13 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, err
 		}
 
+		if c.shouldAutoApproveInstallPlans() {
+			if err := c.reconcileInstallPlans(); err != nil {
+				c.log.Error(err, "unable to reconcile InstallPlans")
+				return ctrl.Result{}, err
+			}
+		}
+
 		if err := c.ensureConsolePlugin(); err != nil {
 			c.log.Error(err, "unable to deploy client console")
 			return ctrl.Result{}, err
@@ -384,6 +405,38 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 	}
 	return ctrl.Result{}, nil
+}
+
+func (c *OperatorConfigMapReconciler) shouldAutoApproveInstallPlans() bool {
+	valueAsString, exist := c.operatorConfigMap.Data[disableInstallPlanAutoApprovalKey]
+	if !exist {
+		return true
+	}
+
+	disableInstallPlanAutoApproval, err := strconv.ParseBool(valueAsString)
+	if err != nil {
+		c.log.Error(err, "failed to parse configmap key data", "key", disableInstallPlanAutoApprovalKey)
+		return true
+	}
+
+	return !disableInstallPlanAutoApproval
+}
+
+func (c *OperatorConfigMapReconciler) reconcileInstallPlans() error {
+	approvePatch := client.RawPatch(types.MergePatchType, []byte(`{"spec":{"approved":true}}`))
+	installPlans := &opv1a1.InstallPlanList{}
+	if err := c.list(installPlans, client.InNamespace(c.OperatorNamespace)); err != nil {
+		return err
+	}
+	for idx := range installPlans.Items {
+		installPlan := &installPlans.Items[idx]
+		if !installPlan.Spec.Approved {
+			if err := c.Patch(c.ctx, installPlan, approvePatch); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
 
 func (c *OperatorConfigMapReconciler) errorOnRookOwnedCsi() error {
--- a/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
@@ -7,7 +7,7 @@ metadata:
     categories: Storage
     console.openshift.io/plugins: '["odf-client-console"]'
     containerImage: quay.io/ocs-dev/ocs-client-operator:latest
-    createdAt: "2025-09-03T10:30:43Z"
+    createdAt: "2025-09-15T05:45:11Z"
     description: OpenShift Data Foundation client operator enables consumption of
       storage services from a remote centralized OpenShift Data Foundation provider
       cluster.
@@ -312,6 +312,15 @@ spec:
           - delete
           - get
           - list
+          - watch
+        - apiGroups:
+          - operators.coreos.com
+          resources:
+          - installplans
+          verbs:
+          - get
+          - list
+          - patch
           - watch
         - apiGroups:
           - operators.coreos.com
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -277,6 +277,15 @@ rules:
 - apiGroups:
   - operators.coreos.com
   resources:
+  - installplans
+  verbs:
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - operators.coreos.com
+  resources:
   - subscriptions
   verbs:
   - delete


internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+	return nil
+}
+
+func (c *OperatorConfigMapReconciler) reconcileSubscriptions() error {
+	pkgNames := [][2]string{
+		// d/s and u/s names of packages for fallback
+		{"ocs-client-operator", ""},
+		{"odf-csi-addons-operator", "csi-addons"},
+		{"cephcsi-operator", "ceph-csi-operator"},
+		{"recipe", ""},
```
nb-ohad, line 807: Why to we need the upstream/downstream distinction? I think we can just have a single list which include all names

internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+	installPlansForApproval := map[string]struct{}{}
+	subsMissingInstallPlans := false
+	for _, pkgName := range pkgNames {
+		sub, err := c.getSubscriptionByPackageName(pkgName[0])
+		if pkgName[1] != "" && kerrors.IsNotFound(err) {
+			sub, err = c.getSubscriptionByPackageName(pkgName[1])
+		}
+		if err != nil {
+			return err
+		}
```
nb-ohad, line 821: I don't see a need for such logic. If you find it update it, if not move to the next one. There is no need for hierarchy

internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+		if c.subscriptionChannel != "" && c.subscriptionChannel != sub.Spec.Channel {
+			sub.Spec.Channel = c.subscriptionChannel
+			if err := c.update(sub); err != nil {
+				return fmt.Errorf("failed to update subscription channel of %s to %v: %v", sub.Name, c.subscriptionChannel, err)
+			}
+		}
+		if sub.Spec.InstallPlanApproval == opv1a1.ApprovalManual && sub.Status.CurrentCSV != sub.Status.InstalledCSV {
```
nb-ohad, line 828: I think that it is not important that we will validate that this is a manual install plan

internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+			ipRef := sub.Status.InstallPlanRef
+			if ipRef != nil {
+				installPlansForApproval[ipRef.Name] = struct{}{}
+			} else {
+				subsMissingInstallPlans = true
+				c.log.Info("installplan is not found in subscription %s status", sub.Name)
```
nb-ohad, line 834: This is kind of meaningless logging, install plans can be deleted. And not finding one means we just skip it

internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+		if c.subscriptionChannel != "" && c.subscriptionChannel != sub.Spec.Channel {
+			sub.Spec.Channel = c.subscriptionChannel
+			if err := c.update(sub); err != nil {
+				return fmt.Errorf("failed to update subscription channel of %s to %v: %v", sub.Name, c.subscriptionChannel, err)
+			}
+		}
+		if sub.Spec.InstallPlanApproval == opv1a1.ApprovalManual && sub.Status.CurrentCSV != sub.Status.InstalledCSV {
+			ipRef := sub.Status.InstallPlanRef
+			if ipRef != nil {
+				installPlansForApproval[ipRef.Name] = struct{}{}
```
nb-ohad, line 831: Is the usage of a map is to prevent duplications. In that case can we just mark the one we already processed instead.

```
- Process the install plan
- marked it in the map as processed
```

Of course the map name will have to change to something like `processedInstallPlans`

internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+		} else if installPlan.UID == "" {
+			// if installplan is not found but mentioned in subscription then it's deleted and
+			// we can't do anything in that scenario
+			c.log.Info("Warning! installplan %s not found, probably deleted", installPlan.Name)
+			continue
+		}
+
+		if err := c.Patch(c.ctx, installPlan, approvePatch); err != nil {
+			return err
+		}
```
nb-ohad, line 855: If it is already approved we don't want a patch operation

internal/controller/operatorconfigmap_controller.go
```diff
@@ -890,3 +784,79 @@ func (c *OperatorConfigMapReconciler) deleteDelegatedCSI() error {
...
+	}
+	if subsMissingInstallPlans {
+		// requeue for OLM to update installplan in the subscriptions
+		return fmt.Errorf("one or more subscriptions are missing installplan")
+	}
```
nb-ohad, line 860: I think this is wrong. We do not control the existence of install plan. If OLM decided that an install plan is not needed or a user manually deleted it  I don't think we should error the entire reconcile

nb-ohad (changes requested):

internal/controller/operatorconfigmap_controller.go
```diff
@@ -85,6 +85,18 @@ const (
...
+var (
+	managedPackageNames = []string{
+		"ocs-client-operator",
+		"odf-csi-addons-operator",
+		"csi-addons",
+		"cephcsi-operator",
+		"ceph-csi-operator",
+		"recipe",
+		"odf-external-snapshotter-operator",
+	}
```
nb-ohad, line 97: Why do we need to manually manage package names? any operator within the namespace can be the reason that upgrades are not happening, as such if we are to approve we need to approve all install plans in the namespace regardless of the content of the install plans

internal/controller/operatorconfigmap_controller.go
```diff
@@ -296,8 +321,13 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		if err := c.reconcileClientOperatorSubscription(); err != nil {
-			c.log.Error(err, "unable to reconcile client operator subscription")
+		if err := c.reconcileSubscriptionLabelForWebhook(); err != nil {
+			c.log.Error(err, "unable to reconcile subscription label")
+			return ctrl.Result{}, err
+		}
+
+		if err := c.reconcileSubscriptions(); err != nil {
+			c.log.Error(err, "unable to reconcile subscriptions")
 			return ctrl.Result{}, err
```
nb-ohad, line 331: We should not bundle all subscription under a single reconcile function. There is a future part where we would like to parallaize the reconcile function and in that case you want each one of these to be its own thing and to do its own steps. If you need or want to refactor shared code, use a util function and call it from each of the specific reconcile functions

internal/controller/operatorconfigmap_controller.go
```diff
@@ -810,84 +796,74 @@ func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() erro
...
+	svc.Name = templates.WebhookServiceName
+	svc.Namespace = c.OperatorNamespace
+	err := c.createOrUpdate(svc, func() error {
+		if err := c.own(svc); err != nil {
+			return err
+		}
+		utils.AddAnnotation(svc, "service.beta.openshift.io/serving-cert-secret-name", "webhook-cert-secret")
+		templates.WebhookService.Spec.DeepCopyInto(&svc.Spec)
+		return nil
+	})
```
nb-ohad, line 810: Not sure what change was introduced here or why we even touched the existing function

internal/controller/operatorconfigmap_controller.go
```diff
@@ -810,84 +796,74 @@ func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() erro
...
 	if err != nil {
 		return err
 	}
-	if c.subscriptionChannel != "" && c.subscriptionChannel != cephCsiOperatorSubscription.Spec.Channel {
-		cephCsiOperatorSubscription.Spec.Channel = c.subscriptionChannel
-		if err := c.update(cephCsiOperatorSubscription); err != nil {
-			return fmt.Errorf("failed to update subscription channel of 'cephcsi-operator' to %v: %v", c.subscriptionChannel, err)
+	if utils.AddLabel(sub, subscriptionLabelKey, subscriptionLabelValue) {
+		if err := c.update(sub); err != nil {
+			return err
```
nb-ohad, line 825: I don't see any reason for this code to exist

internal/controller/operatorconfigmap_controller.go
```diff
@@ -810,84 +796,74 @@ func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() erro
...
-		return err
-	}
-	if c.subscriptionChannel != "" && c.subscriptionChannel != recipeOperatorSubscription.Spec.Channel {
-		recipeOperatorSubscription.Spec.Channel = c.subscriptionChannel
-		if err := c.update(recipeOperatorSubscription); err != nil {
-			return fmt.Errorf("failed to update subscription channel of 'recipe' to %v: %v", c.subscriptionChannel, err)
+func (c *OperatorConfigMapReconciler) reconcileSubscriptions() error {
+	installPlansForApproval := map[string]struct{}{}
+
+	for _, pkgName := range managedPackageNames {
```
nb-ohad, line 834: You should not treat all subscription the same, each one might have its own nuances, even if they are not manifested today. Please make sure each subscription is dealt with its own reconcile function

Also, as a general rule, one reconcile function need to only deal with one output resource

internal/controller/operatorconfigmap_controller.go
```diff
@@ -810,84 +796,74 @@ func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() erro
...
-		return err
-	}
-	if c.subscriptionChannel != "" && c.subscriptionChannel != odfSnapshotterSubscription.Spec.Channel {
-		odfSnapshotterSubscription.Spec.Channel = c.subscriptionChannel
-		if err := c.update(odfSnapshotterSubscription); err != nil {
-			return fmt.Errorf("failed to update subscription channel of 'odf-external-snapshotter-operator' to %v: %v", c.subscriptionChannel, err)
+			ipRef := sub.Status.InstallPlanRef
+			if ipRef != nil {
+				installPlansForApproval[ipRef.Name] = struct{}{}
+			}
```
nb-ohad, line 850: This does not make sense, approval of install planes has nothing to do with the subscriptions. The content of the install plan does not matter.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -810,84 +796,74 @@ func (c *OperatorConfigMapReconciler) reconcileNoobaaOperatorSubscription() erro
...
+	for ipName := range installPlansForApproval {
+		installPlan := &opv1a1.InstallPlan{}
+		installPlan.Name = ipName
+		installPlan.Namespace = c.OperatorNamespace
+		if err := c.get(installPlan); client.IgnoreNotFound(err) != nil {
 			return err
+		} else if installPlan.UID != "" && !installPlan.Spec.Approved {
+			if err := c.Patch(c.ctx, installPlan, approvePatch); err != nil {
+				return err
+			}
```
nb-ohad, line 864: Touching an install plane should be done in its own reconcile function. not as part of a reconcile function that deals with changes to subscriptions

nb-ohad (changes requested):

internal/controller/operatorconfigmap_controller.go
leelavg: mistake on my part, there was a missing local change.

internal/controller/operatorconfigmap_controller.go
leelavg: done.

internal/controller/operatorconfigmap_controller.go
leelavg: done.

leelavg:
Depends on red-hat-storage/ocs-operator#3510

internal/controller/operatorconfigmap_controller.go
```diff
@@ -342,6 +356,17 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, err
 		}
 
+		autoApproveInstallPlans, err := strconv.ParseBool(cmp.Or(c.operatorConfigMap.Data[autoApproveInstallPlansKey], "true"))
```
nb-ohad, line 359: 1. by having a key where the value need to be true in order to exec a code path it automaticaly implies that if it does not exists the code path should not be executed. I am not talking about the code being wrong I am talking about API design principles 

2. This code already have an issue as it defaults to false if the value is not proper bool which differs to defulating to true if the value does not exists.

My request is to flip the logic so the key will indicate that we are to "disable" auto approval, it will solve both problems in one go

```suggestion
		disableInstallPlanAutoApproval, err := strconv.ParseBool(c.operatorConfigMap.Data[DisableInstallPlanAutoApproval])
```

nb-ohad (changes requested):

nb-ohad:
\lgtm

leelavg:
/cherrypick release-4.20
via DFBUGS-2631

leelavg:
/unhold

tested below:
1. approve ip if key doesn't exist in cm or value for the key is false
2. do not approve if the value for the key is true or the value is malformed

PR #385, opened 2025-05-29
controllers: apply cni network annotation to csi controller plugins for internal client

rohan47:
when storagecluster is on multus network then csi controller plugin pods need to have cni network annotation. operator config controller applies this annotation to csi controler plugin if storageclient has this annotation.


internal/controller/operatorconfigmap_controller.go
```diff
@@ -433,6 +440,12 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 			return fmt.Errorf("failed to own csi rbd driver: %v", err)
 		}
 		rbdDriver.Spec.GenerateOMapInfo = ptr.To(c.shouldGenerateRBDOmapInfo())
+		if cniNetworkAnnotationExists {
+			rbdDriver.Spec = csiopv1a1.DriverSpec{}
+			rbdDriver.Spec.ControllerPlugin = &csiopv1a1.ControllerPluginSpec{}
+			rbdDriver.Spec.ControllerPlugin.Annotations = make(map[string]string)
+			rbdDriver.Spec.ControllerPlugin.Annotations[cniNetworksAnnotationKey] = cniNetworkAnnotationValue
+		}
```
rewantsoni, line 448: Why are we applying this to each driver and not directly to the operatorConfig?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -423,6 +424,12 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		return fmt.Errorf("failed to reconcile csi operator config: %v", err)
 	}
 
+	storageClients := v1alpha1.StorageClientList{}
+	if err := c.list(&storageClients); err != nil {
+		return fmt.Errorf("failed to get storage clients: %v", err)
+	}
+	cniNetworkAnnotationValue, cniNetworkAnnotationExists := storageClients.Items[0].GetAnnotations()[cniNetworksAnnotationKey]
+
```
rewantsoni, line 432: Should we follow some merge semantics here in case of multiple clients?

rewantsoni (changes requested):

internal/controller/operatorconfigmap_controller.go
```diff
@@ -423,6 +424,12 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		return fmt.Errorf("failed to reconcile csi operator config: %v", err)
 	}
 
+	storageClients := v1alpha1.StorageClientList{}
+	if err := c.list(&storageClients); err != nil {
+		return fmt.Errorf("failed to get storage clients: %v", err)
+	}
+	cniNetworkAnnotationValue, cniNetworkAnnotationExists := storageClients.Items[0].GetAnnotations()[cniNetworksAnnotationKey]
```
Madhu-1, line 431: what if we dont find any storageClients? this will panic, please have a length check

internal/controller/operatorconfigmap_controller.go
Madhu-1: Yes this is preferred

internal/controller/operatorconfigmap_controller.go
```diff
@@ -423,6 +424,12 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		return fmt.Errorf("failed to reconcile csi operator config: %v", err)
 	}
 
+	storageClients := v1alpha1.StorageClientList{}
+	if err := c.list(&storageClients); err != nil {
+		return fmt.Errorf("failed to get storage clients: %v", err)
```
Madhu-1, line 429:
```suggestion
return fmt.Errorf("failed to list storage clients: %v", err)
````

File: internal/controller/operatorconfigmap_controller.go (hunk 3 above)
Comment (rohan47) [reply internal/controller/operatorconfigmap_controller.go]: added

File: internal/controller/operatorconfigmap_controller.go [hunk 5]
```diff
@@ -398,6 +399,16 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		return fmt.Errorf("unable to find the updated cluster version")
 	}
 
+	storageClients := v1alpha1.StorageClientList{}
+	if err := c.list(&storageClients); err != nil {
+		return fmt.Errorf("failed to list storage clients: %v", err)
+	}
```
nb-ohad, line 405: There is no reason the list the clients. The client CRs are already listed somewhere else in this controller

internal/controller/operatorconfigmap_controller.go
```diff
@@ -398,6 +399,16 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
...
+	var cniNetworkAnnotationExists bool
+	var cniNetworkAnnotationValue string
+	if len(storageClients.Items) != 0 {
+		cniNetworkAnnotationValue, cniNetworkAnnotationExists = storageClients.Items[0].GetAnnotations()[cniNetworksAnnotationKey]
```
nb-ohad, line 409: That is really not what we should be doing. We need merge logic here

rohan47: In internal mode with multus, the local client only needs to connect to multus network and since multus is supported only for internal mode, there will not be any other client on multus.

In case of multus, pod has connection to both pod network and multus network.
So any other client connected to different provider will also work as it will not be on multus network.
So do we really need merge logic?
We will not be adding any more annotation from other clients till multus is supported for multicluster deployment mode

internal/controller/operatorconfigmap_controller.go
rohan47: Yes, But I can only use c.list() to list the storageclients, its not directly available in OperatorConfigMapReconciler
Do we want the storageclient list to be in OperatorConfigMapReconciler struct?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -398,6 +405,19 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
...
+		for i := range storageClients.Items {
+			annotationValue := storageClients.Items[i].GetAnnotations()[cniNetworksAnnotationKey]
+			if annotationValue != "" {
+				cniNetworkAnnotationValues = append(cniNetworkAnnotationValues, annotationValue)
+			}
+		}
+		if len(cniNetworkAnnotationValues) != 1 {
+			return fmt.Errorf("only one client with CNI network annotation value is supported, got %d", len(cniNetworkAnnotationValues))
+		}
+	}
```
nb-ohad, line 419: - There is no need to check agains 0 items, the range operator will automatically covers it
- There is no need to have a list of annotation of we are to fail after we find the first one
```suggestion
	cniNetworksAnnotationValue : = ""
	for i := range storageClients.Items {
		if value := storageClients.Items[i].GetAnnotations()[cniNetworksAnnotationKey]; value != "" {
			if cniNetworksAnnotationValue != "" {
				return fmt.Errorf("only one client with CNI network annotation value is supported")
		
			}
			cniNetworksAnnotationValue = annotationValue
		}
	}
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -418,6 +438,12 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 		}
 		csiCtrlPluginHostNetwork, _ := strconv.ParseBool(c.operatorConfigMap.Data[useHostNetworkForCsiControllersKey])
 		csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.HostNetwork = ptr.To(csiCtrlPluginHostNetwork)
+		if len(cniNetworkAnnotationValues) == 1 {
+			if csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.Annotations == nil {
+				csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.Annotations = map[string]string{}
+			}
+			csiOperatorConfig.Spec.DriverSpecDefaults.ControllerPlugin.Annotations[cniNetworksAnnotationKey] = cniNetworkAnnotationValues[0]
+		}
```
nb-ohad, line 446: The entire mutate function will benefit from caching the DriverSpecDefaults in it's own var:
```driverSpecDefaults := &csiOperatorConfig.Spec.DriverSpecDefaults```

This will also make the readability of your logic much better:
```suggestion
		if cniNetworkAnnotationValue != "" {
			utils.AddAnnotation(
				driverSpecDefaults.ControllerPlugin.Annotations, 
				cniNetworkAnnotationKey,
				cniNetworkAnnotationValue,
			)
		}
```

nb-ohad (changes requested):

internal/controller/operatorconfigmap_controller.go
```diff
@@ -411,13 +428,20 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI() error {
 			return fmt.Errorf("failed to own csi operator config: %v", err)
 		}
 		templates.CSIOperatorConfigSpec.DeepCopyInto(&csiOperatorConfig.Spec)
-		csiOperatorConfig.Spec.DriverSpecDefaults.ImageSet = &corev1.LocalObjectReference{Name: cmName}
-		csiOperatorConfig.Spec.DriverSpecDefaults.ClusterName = ptr.To(string(clusterVersion.Spec.ClusterID))
+		driverSpecDefaults := csiOperatorConfig.Spec.DriverSpecDefaults
```
nb-ohad, line 431: This will create a copy and will not mutate the original struct.
```suggestion
		driverSpecDefaults := &csiOperatorConfig.Spec.DriverSpecDefaults
```

nb-ohad (changes requested):

PR #393, opened 2025-06-09
controllers: Update client sub only if client version < provider version

malayparida2000:
This is to avoid early update of client operator csv which can lead to the provider trailing behind the client operator which can cause upgrade deadlock.
Ref-https://issues.redhat.com/browse/DFBUGS-2502


leelavg:
close if not required, thanks.

PR #395, opened 2025-06-11
csi: add secondary name check for ex-snapshotter

Nikhil-Ladha:
add secondary name check for ex-snapshotter being "odf-external-snapshotter-operator"

### DIFF
--- a/config/metadata/dependencies.yaml
+++ b/config/metadata/dependencies.yaml
@@ -17,5 +17,5 @@ dependencies:
       version: "0.0.1"
   - type: olm.package
     value:
-      packageName: odf-snapshot-controller
+      packageName: odf-external-snapshotter-operator
       version: ">=4.20.0"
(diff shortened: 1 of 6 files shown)


internal/controller/operatorconfigmap_controller.go
```diff
@@ -793,6 +793,9 @@ func (c *OperatorConfigMapReconciler) reconcileRecipeOperatorSubscription() erro
 
 func (c *OperatorConfigMapReconciler) reconcileODFSnapshotterSubscription() error {
 	odfSnapshotterSubscription, err := c.getSubscriptionByPackageName("odf-snapshot-controller")
+	if kerrors.IsNotFound(err) {
+		odfSnapshotterSubscription, err = c.getSubscriptionByPackageName("odf-external-snapshotter-operator")
```
leelavg, line 797: pls reverse the search, have the d/s naming first as prod clusters have higher lifetime than u/s or test clusters.

Madhu-1:
@Nikhil-Ladha what is the reason for having 2 checks? we dont have upstream name for this one.

Nikhil-Ladha:
The PR is updated to check only for name `odf-external-snapshot-operator` in the code and also updated the bundle name in the bundle dependencies.

config/metadata/dependencies.yaml
leelavg, line 1: What about https://github.com/red-hat-storage/ocs-client-operator/blob/ca5ec26d0ed7699e657bbbf41aca5f56c6cd7a8c/hack/build-catalog.sh#L68

PR #396, opened 2025-06-11
controllers: verify kind before deletion of StorageClaim

ShravaniVangur:
The PR [#382](https://github.com/red-hat-storage/ocs-client-operator/pull/382) implemented deletion of StorageClaims as a part of the cleanup process. However, it did not account for scenarios where the StorageClaim kind is not present in the cluster. This is handled by the current PR.

Resolves: [DFBUGS-2625](https://issues.redhat.com/browse/DFBUGS-2625)

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -44,6 +44,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


internal/controller/storageclient_controller.go
```diff
@@ -438,6 +439,9 @@ func (r *storageClientReconcile) deleteStorageClaims(claimName string) error {
 	storageClaim.Name = claimName
 
 	if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); client.IgnoreNotFound(err) != nil {
+		if meta.IsNoMatchError(err) {
+			return nil
+		}
 		return fmt.Errorf("Failed to get StorageClaim %q: %w", claimName, err)
```
leelavg, line 445: this is introducing matching of specific errors in two places, can be written as below still maintaining single positive exit at the end
```suggestion
	if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); if err != nil {
		if !(meta.IsNoMatchError(err) || kerrors.IsNotFound(err)) {
                    return fmt.Errorf("Failed to get StorageClaim %q: %w", claimName, err)
		}	
```

internal/controller/storageclient_controller.go
```diff
@@ -438,6 +439,9 @@ func (r *storageClientReconcile) deleteStorageClaims(claimName string) error {
 	storageClaim.Name = claimName
 
 	if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); client.IgnoreNotFound(err) != nil {
+		if meta.IsNoMatchError(err) {
```
leelavg, line 442:
```suggestion
	if err := r.Client.Get(r.ctx, types.NamespacedName{Name: claimName}, storageClaim); err != nil {
		if meta.IsNoMatchError(err) || kerrors.IsNotFound(err) {
```
and also remove the UID check.

internal/controller/storageclient_controller.go
```diff
@@ -438,6 +439,9 @@ func (r *storageClientReconcile) deleteStorageClaims(claimName string) error {
...
+			return nil
+		}
```
nb-ohad, line 444: There is no reason to have another if here - as we already have an if filtering on errors.  
And there is no reason have a separate return here - as we already skipping storage claims that have no uid set

Just have the `IsNotMathError` check be part of the upper condition
(and also move to uses r.get instead r.Client.Get)
```
if err := r.Client.Get(storegeClaim); client.IgnoreNotFound(err) != nil  && !meta.IsNoMatchError(err) {
```

PR #365, opened 2025-05-12
upgrade noobaa along with client-operator

leelavg:
in remote client clusters noobaa should follow client-operator subscription channel. we had tied noobaa feature usage with operator upgrade resulting in operator not upgrading from 4.17.

revert of #219

PR #399, opened 2025-06-12
DFBUGS-2587: [release-4.18] upgrade noobaa along with client-operator

leelavg:
manual cherry-pick of #365


leelavg:
we'll be removing noobaa altogether from client-op, until then this will allow clusters currently stuck at df-client 4.17 to upgrade df-client 4.18

PR #426, opened 2025-07-03
controllers: remove noobaa reconciliation from standard list

leelavg:
there is a possibility that provider is still at older Z stream and client gets upgraded to next Y stream, in that scenario even if we remove existing noobaa cr, csv and sub, a new cr is getting created.


leelavg:
already raised in #424

PR #434, opened 2025-07-17
controllers: manage webhook only if we are running standalone

leelavg:
when we run in hub then the management of subscriptions should be left to odf-operator or else we get stuck during upgrades due to the design of upgrading all dependents of odf-op in a single shot.

this commit doesn't reconcile webhook (and delete if exists) when we are running in hub.

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -253,9 +253,28 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, err
 	}
 
-	var err error
-	if c.subscriptionChannel, err = c.getDesiredSubscriptionChannel(storageClients); err != nil {
+	// TODO: this is an interim fix until we decide on who manages the subscription of client-op
+	// including it's dependents when we run in hub
+	//
+	// since odf-op is the top level operator it is guaranteed that it's existence indicates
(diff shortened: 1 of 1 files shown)


leelavg:
tested.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -253,9 +253,30 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	if err := c.list(
+		csvList,
+		client.InNamespace(c.OperatorNamespace),
+		client.MatchingLabels{odfOperatorLabel: ""},
+		client.Limit(1),
+	); err != nil {
 		return reconcile.Result{}, err
+	} else if len(csvList.Items) == 1 {
+		runningOnHub = true
+	}
```
nb-ohad, line 273:
```suggestion
	csvList := &opv1a1.ClusterServiceVersionList{}
	odfOperatorLabel := fmt.Sprintf("operators.coreos.com/odf-operator.%s", c.OperatorNamespace)
	if err := c.list(
		csvList,
		client.InNamespace(c.OperatorNamespace),
		client.MatchingLabels{odfOperatorLabel: ""},
		client.Limit(1),
	); err != nil {
		return reconcile.Result{}, err
	}
	runningOnHub := len(csvList.Items) > 0 
```

nb-ohad (changes requested):

PR #441, opened 2025-07-29
csi: update ceph-csi-operator api to v1

black-dragon74:
This patch updates ceph-csi-operator API types
from v1alpha1 to v1


internal/controller/storageclient_controller.go
```diff
@@ -93,14 +93,14 @@ var (
 	}
 	kindsToReconcile = []client.Object{
 		&quotav1.ClusterResourceQuota{},
-		&csiopv1a1.CephConnection{},
-		&csiopv1a1.ClientProfileMapping{},
```
leelavg, line 97: do not remove these for 4.20, this'll also act a pruning mechanism, the idea is as follows:
t0: hub 4.19 sends v1a1 to client 4.19
t1: hub 4.20 should send v1a1 to client 4.19
t2/a: hub 4.20 should send v1 to client 4.20
t2/b: client 4.20 will prune v1a1 that were created in 4.19

the last line, t2/b depends on having the kind in this list. now, the question is, what are the effects of deleting the v1a1 CRs?

PR #442, opened 2025-07-29
Csi/enable fencing

iPraveenParihar:
This PR includes -

- updates the API version of ceph-csi-operator
from v1alpha1 to v1.

- set `EnableFencing` field in OperatorConfig.DriverSpecDefaults
to true.


internal/controller/operatorconfigmap_controller.go
```diff
@@ -520,6 +520,7 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 			}
 		}
 		driverSpecDefaults.GenerateOMapInfo = ptr.To(c.shouldGenerateRBDOmapInfo())
+		driverSpecDefaults.EnableFencing = ptr.To(true)
```
Madhu-1, line 523: please address https://github.com/ceph/ceph-csi-operator/pull/281/files#r2241692992 or else this change will not work

Madhu-1 (changes requested):
change it here https://github.com/red-hat-storage/ocs-client-operator/blob/main/pkg/templates/csi.go#L91

PR #452, opened 2025-08-13
controllers: ensure the storageclient status updates after onboard

leelavg:
when the hub signals success for onboard and spoke fails to persist the update, spoke will never get a positive response.

we will try to patch the status till it persists or fails (unlikely).

below is not for interpretation only as a note:

at hub:
t0: cluster controller created consumer
t1: consumer controller added annotation
t2: api-server got the obj at t0 and failed trying to enable after t1 t3: api-server enabled the consumer for t1 object
t4: consumer is already enabled and don't onboard again

at spoke:
t0: cluster controller created client
t1: client controller added finalizer
t2: onboard failed due to stale consumer at hub
t3: onboard successful with t0 object but failed trying to update status t4: not onboarded and stuck in a loop due to loosing memory that is was successful at t3

this commit fixes t3 at spoke as not to hit t4.


nb-ohad:
Summarising our offline discussion: 
The current solution on the PR is a bandage that has the potential to create even more timing issues. 

The underlaying issue is that post finalizer update we continue the reconciliation where the cache and the object or desync. A better solution is to requeue after adding the finalizer.

internal/controller/storageclient_controller.go
```diff
@@ -411,6 +410,15 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 		return reconcile.Result{}, err
 	}
 
+	if controllerutil.AddFinalizer(&r.storageClient, storageClientFinalizer) {
+		r.storageClient.Status.Phase = v1alpha1.StorageClientInitializing
```
nb-ohad, line 414: Why are we setting this here? shouldn't we be in initializing phase unrelated to the finalizer?

PR #460, opened 2025-09-03
controllers: remove deprecated code for 4.20

rewantsoni:
This PR removes the deprecated code and permissions no longer required in 4.20+


leelavg:
@rewantsoni i think this missed removal of one time transfer of tolerations?

PR #472, opened 2025-09-22
Use lower cpu, memory requests for csi sidecars

Nikhil-Ladha:
Use lower cpu, memory requqets for csi sidecars, to allow deployment in minimum supported lower spec clusters.
Ref: https://issues.redhat.com/browse/DFBUGS-4052


Madhu-1:
> @Madhu-1 should this be approved as is? I don't know why we want to have restrictions for CSI pods as they are scaled as per cluster requirements.
> 

Yes this sounds good. I have not seen this much usage in a cluster with 20k PVC

> This PR has a 10x diff b/n requests & limits which is throttled when the pod lands on a busy node.

PR #500, opened 2025-11-27
controllers: delete nfs driver if nfs is not enabled

rohan47:
delete nfs driver if nfs is not enabled and nfs driver exists, also there are no PVCs using nfs storageclass
[DFBUGS-4704](https://issues.redhat.com//browse/DFBUGS-4704)

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -39,6 +39,7 @@ import (
 
 	csiopv1 "github.com/ceph/ceph-csi-operator/api/v1"
 	"github.com/go-logr/logr"
+	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
 	configv1 "github.com/openshift/api/config/v1"
 	secv1 "github.com/openshift/api/security/v1"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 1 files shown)


internal/controller/operatorconfigmap_controller.go
```diff
@@ -687,6 +713,23 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 		}); err != nil {
 			return fmt.Errorf("failed to reconcile nfs driver: %v", err)
 		}
+	} else {
+		nfsDriver := &csiopv1.Driver{}
+		nfsDriver.Name = templates.NfsDriverName
+		nfsDriver.Namespace = c.OperatorNamespace
```
Nikhil-Ladha: We can move this from here and the `if` condition above, to the outside block in someway like this:
```
nfsDriver := &csiopv1.Driver{
    Name: templates.NfsDriverName,
    Namespace: c.OperatorNamespace
}
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -643,6 +669,22 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 		}); err != nil {
 			return fmt.Errorf("failed to reconcile nfs driver: %v", err)
 		}
+	} else {
+		if hasPvs, err := c.hasPersistentVolumesWithNfsDriver(); err != nil {
+			return fmt.Errorf("failed to check if NFS driver has PVs: %v", err)
+		} else if hasPvs {
+			c.log.Info("NFS driver has PVs, skipping deletion")
```
ezio-auditore: Do we really need to log this info?

leelavg (approved):

PR #524, opened 2026-03-02
Configure nginx reverse proxy for the external endpoints

SanjalKatiyar:
**Design doc:** https://docs.google.com/document/d/1yId5AWQ-Og-FXEGIPFK4EKop-KEqffFxC3ZVFUKnWgc/edit?tab=t.ns83vyascte9

**Corresponding OCS operator changes:** https://github.com/red-hat-storage/ocs-operator/pull/3695

- Mounted per client proxy config to the nginx (console) pod (using `proxy-<S3_ENDPOINTS_CONFIGMAP_NAME>.conf` ConfigMap key for this).
- Added optional Secret for injecting custom CA certs per endpoint (using `<S3_ENDPOINTS_CONFIGMAP_NAME>-noobaaS3.crt` Secret key for this).
- Updated nginx config.
- Added rate and connection limits.
- Added whitelisting to allow management operations only.
- Cleanup while client deletion.

https://issues.redhat.com/browse/RHSTOR-8528
https://issues.redhat.com/browse/RHSTOR-8529
https://issues.redhat.com/browse/RHSTOR-8530

**Testing:**
_On an existing 4.22 cluster (operator already deployed), updated CSV with custom image, RBACs and new volume mounts for the console pod -_
##### Client Object Browser (UI), accessing S3 deployed on Provider, recording: 
https://drive.google.com/file/d/18Y_W4yB2rZiwSAqfV8JoPqY4jZOocX7L/view?usp=sharing

_On an existing 4.21 cluster (operator already deployed), updated CSV with custom image, RBACs and new volume mounts for the console pod  (had to add few more RBACs for OB/OBC/Infrastructure CRDs/CRs due to recent additions in 4.22) -_
##### Created nginx ConfigMapData:
```
data:
  nginx.conf: <ENTIRE_ROOT_CONFIG_FILE>
  proxy-ocs-s3-endpoints-list-f603c982-27fb-4a52-97c5-5ddda6f3bc99.conf: |
    location /ocs-s3-endpoints-list-f603c982-27fb-4a52-97c5-5ddda6f3bc99/noobaaS3/ {
        ...
        ...
        proxy_pass https://s3-openshift-storage.apps.dr2-apr-8b-26.devcluster.openshift.com/;
        proxy_set_header Host s3-openshift-storage.apps.dr2-apr-8b-26.devcluster.openshift.com;
        proxy_request_buffering off;
        proxy_buffering off;
        proxy_ssl_name s3-openshift-storage.apps.dr2-apr-8b-26.devcluster.openshift.com;
        proxy_ssl_verify on;
        proxy_ssl_server_name on;
        proxy_ssl_trusted_certificate /etc/ssl/certs/s3-endpoint-ca-certs/ocs-s3-endpoints-list-f603c982-27fb-4a52-97c5-5ddda6f3bc99-noobaaS3.crt;
    }
```
##### Console pod's directories:
Screenshot 2026-04-08 at 8 57 40 PM


api/v1alpha1/storageclient_types.go
```diff
@@ -55,6 +55,9 @@ type StorageClientStatus struct {
 	// ConsumerID will hold the identity of this cluster inside the attached provider cluster
 	ConsumerID string `json:"id,omitempty"`
 
+	// ExternalEndpoints holds info about external endpoints deployed on the provider (e.g. NooBaa S3, NooBaa IAM, RGW S3 etc.)
+	ExternalEndpoints map[string]string `json:"externalEndpoints,omitempty"`
+
```
SanjalKatiyar: We need to advertise the Routes to the clients, hence adding it to the StorageClient CR's status.
1. Console itself can't send gRPC calls to the Provider server, so we need to persist these Route somewhere inside the cluster, we need direct Routes for "data operations" (only "management operations" are proxied via nginx).
2. AFAIU this will be helpful for the Lakehouse use case as well (there also we have a need for endpoint advertise ?).

pkg/console/nginx_conf.go
```diff
@@ -37,33 +72,115 @@ http {
...
+        "~*^DELETE:/[^/]+/[^/]+/[^/]+/.+:.*x-id=AbortMultipartUpload"  1;
+
+        # Bucket-level and Object-level HEAD (metadata only)
+        "~*^HEAD:/[^/]+/[^/]+/[^/]+:?.*"            1;
+        "~*^HEAD:/[^/]+/[^/]+/[^/]+/.+"             1;
+
+        # Service-level, Bucket-level and Object-level OPTIONS (CORS preflight)
+        "~*^OPTIONS:/[^/]+/[^/]+/?:?.*"             1;
+        "~*^OPTIONS:/[^/]+/[^/]+/[^/]+:?.*"         1;
+        "~*^OPTIONS:/[^/]+/[^/]+/[^/]+/.+"          1;
```
SanjalKatiyar: did some initial testing (using AWS CLI tool) and it was working as expected...
but will do the final testing once console (AWS SDK) changes are ready and update (if needed) these whitelisting config accordingly...

pkg/console/nginx_conf.go
```diff
@@ -37,33 +72,115 @@ http {
...
+            limit_conn root_conn 400;
+            send_timeout 60s;  # Time to send response to client
+
             root   /opt/app-root/src;
+            ssi on;
+            add_header Last-Modified $date_gmt;
+            add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
+            if_modified_since off;
+            expires off;
+            etag off;
```
SanjalKatiyar: these settings are unrelated to this PR, it's already present for these "location(s)" currently... we can look into removing them (if needed) later as a separate (unrelated to this PR) issue...

internal/controller/storageclient_controller.go
leelavg: don't particularly like reconciliation of console in storageclient-controller, how difficult it'd be to move this to configmap controller? the unwritten kinda rule is, operator level singletons (destinations) are reconciled in configmap controller, not strictly enforced but this pattern emerged over the course of years writing these controllers.

like:
there can only be one subscription, one installplan, one csi-controller, one console deployment, webhook etc which are reconciled by configmap controller.

multiple *classes, secrets, clientprofiles, cephconnection etc reconciled by storageclient controller.

I feel nginx config, the reverse proxy configurations all just need to be created in a configmap (referring to the ocs-op pr) and be reconciled by cm controller? wdyt?

leelavg (commented):
will provide an indepth review once a decision is reached on the flow in ocs-op and client-op PRs.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -90,8 +94,30 @@ const (
 	cniNetworksAnnotationKey   = "k8s.v1.cni.cncf.io/networks"
 	noobaaCrdName              = "noobaas.noobaa.io"
 	noobaaCrName               = "noobaa-remote"
+
+	// disableExternalEndpointProxyKey, if true, disables deploying the external endpoint reverse proxy for the local/internal client
+	disableExternalEndpointProxyKey = "disableExternalEndpointProxy"
+	externalEndpointsLabelKey       = "ocs.openshift.io/external-endpoints"
+	externalEndpointsLabelValue     = "true"
```
nb-ohad: Please don't use the term external here. External means something different in this context. As suggested on the other PR, the correct terminology here should be

```suggestion
	ocsHubS3EndpointConfigMapLabelKey       = "ocs.openshift.io/hub-s3-endpoints
```

And there is truly no need to have a constant for true, especially when you only care about the existence of the label key and not the value. And even if you did care, you can always just use strconv.FormatBool(true) for consistency

internal/controller/operatorconfigmap_controller.go
```diff
@@ -90,8 +94,30 @@ const (
...
+
+	// nginx proxy config key pattern per client: proxy-<clientuid>.conf (all locations for this client in one key).
+	nginxProxyConfigKeyPrefix = "proxy-"
+	nginxProxyConfigKeySuffix = ".conf"
+
+	// default system CA bundle used by nginx for TLS verification.
+	defaultCertsPath = "/etc/pki/tls/certs/ca-bundle.crt"
```
nb-ohad, line 105: This is not that path to get the CA from.
The correct path should be: `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` and even then the expected way to reach this ca is by expending the `KUBERNETES_SERVICE_ACCOUNT_DIR` env var

internal/controller/operatorconfigmap_controller.go
```diff
@@ -90,8 +94,30 @@ const (
...
+	// optional Secret for custom CA certs; keys are "<clientuid>-<exposeas>.crt".
+	externalEndpointCASecretName  = "ocs-client-operator-console-external-endpoint-ca-certs"
+	externalEndpointCertKeySuffix = ".crt"
+	// the mount path for custom CA secret in the console pod.
+	externalEndpointCertsMountPath = "/etc/ssl/certs/external-endpoint-ca-certs"
```
nb-ohad: Don't use the external terminology, it means something else in the context of clients.
Please apply the request here to the rest of the changes in this pr

internal/controller/operatorconfigmap_controller.go
```diff
@@ -164,6 +190,23 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		),
 	)
 
+	externalEndpointsConfigMapPredicates := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(obj client.Object) bool {
+				return obj.GetNamespace() == c.OperatorNamespace &&
+					obj.GetLabels()[externalEndpointsLabelKey] == externalEndpointsLabelValue
+			},
```
nb-ohad: 1. you will get a nil pointer exception if you don't check for the existence of a map for the labels 
2. you don't care about the value of the label, just the existence of it


```suggestion
			func(obj client.Object) bool {
			    if obj.GetNamespace() == c.OperatorNamespace {
			        if labels := obj.GetLabels(); labels != nil {
			            _, ok := labels[externalEndpointsLabelKey]
			            return ok
			        }
			    }
                return false
			},
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -219,6 +262,16 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					storageClientStatusPredicate,
 				),
 			),
+		).
+		Watches(
+			&corev1.ConfigMap{},
+			enqueueConfigMapRequest,
+			externalEndpointsConfigMapPredicates,
+		).
```
nb-ohad: This does not make a lot of sense, we are already watching configmaps, why can we just amend the original config map watch with a predicate that will cover all cases? (maybe using the Or Predicate)

internal/controller/operatorconfigmap_controller.go
```diff
@@ -750,28 +805,37 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		nginxConfigMap.Data = desiredData
 		return controllerutil.SetControllerReference(c.consoleDeployment, nginxConfigMap, c.Scheme)
 	})
-
 	if err != nil {
 		c.log.Error(err, "failed to create nginx config map")
 		return err
 	}
 
+	if !maps.Equal(priorNginxData, nginxConfigMap.Data) {
```
nb-ohad: There is no point in trying to calculate it by yourself. CreateOrUpdate has a result return value that would tell you if an update or create happened. We do not expose it with our c.CreateOrUpdate util, but you can easily change that. If there was no change in the actual definition of the configmap then you will get a result that indicate so so your if here would just check that the result is either Creates or Updated

internal/controller/operatorconfigmap_controller.go
```diff
@@ -740,28 +784,37 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		c.log.Info("nginx ConfigMap data changed, restarting console pod to pick up new config")
+		// Console pod must restart (or nginx must reload) to pick up updated mounted config files.
+		// Restart failures are currently logged only. If restart fails once and config does not change again,
+		// subsequent reconciles will not retry because this block is gated by data-diff check.
+		// ToDo: If issue is prominent, explore reliable retry semantics, like restart based on marker/hash or graceful nginx reload.
+		// Note: checksum rollout via Deployment pod-template annotation is not reliable as Deployment is currently OLM/CSV-owned.
```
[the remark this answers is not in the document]
nb-ohad: pod restarts are problematic because nothing ensure that the pod will get scheduled again, unless you are killing the pod using a kill signal. What is the issue of the pod to re-read the file when it gets updated?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		c.log.Error(err, "failed to list console pods by label selector", "namespace", c.OperatorNamespace)
+		return
+	}
+	if len(podList.Items) == 0 {
+		c.log.Info("no console pods found for restart", "namespace", c.OperatorNamespace)
+		return
+	}
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		if err := c.delete(pod); err != nil {
```
nb-ohad, line 905: This is dangerous because you are not killing the pod, you are deleting the pod expecting that another pod will come up to take its place. But the way k8s works is that the new pod need to be scheduled and nobody can guarantee that. What if some other application pods are already waiting on the needed resources (mem, cpu, etc.)?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+	}
+}
+
+func (c *OperatorConfigMapReconciler) buildDesiredNginxDataWithProxies() (map[string]string, error) {
+	out := map[string]string{
+		// Root config is mandatory for nginx to start. Proxy configs (per client) are optional.
+		"nginx.conf": console.GetNginxRootConf(),
+	}
+
+	if strings.ToLower(strings.TrimSpace(c.operatorConfigMap.Data[disableExternalEndpointProxyKey])) == "true" {
```
nb-ohad:
```suggestion
	if strconv.ParseBool(c.operatorConfigMap.Data[disableExternalEndpointProxyKey])) {
```

In addition you cannot make an assumption that the map has a data key in it. Someone can mess up the configmap and this will cause a nil pointer exception here

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		c.log.Info("external endpoint proxy disabled via operator ConfigMap", "key", disableExternalEndpointProxyKey)
+		return out, nil
+	}
+
+	desiredProxies, err := c.computeDesiredProxyConfigByKey()
+	if err != nil {
+		return nil, err
+	}
+	maps.Copy(out, desiredProxies)
+	return out, nil
```
nb-ohad: Instead of doing a maps copy, I would prefer if you just send `out` to the `computeDesiredProxyConfigsByKey` function


```suggestion
	err := c.computeDesiredProxyConfigByKey(out)
    return out, err
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+}
+
+func (c *OperatorConfigMapReconciler) computeDesiredProxyConfigByKey() (map[string]string, error) {
```
nb-ohad: Prefer sending a map to be filled then returning a new map

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		cm := &extList.Items[i]
+		ownerRef := metav1.GetControllerOf(cm)
+		if ownerRef == nil ||
+			ownerRef.Kind != storageClientGVK.Kind ||
+			ownerRef.APIVersion != storageClientGVK.GroupVersion().String() ||
+			ownerRef.UID == "" {
+			c.log.Info("skipping external endpoints ConfigMap without StorageClient controller ownerRef",
+				"configmap", client.ObjectKeyFromObject(cm).String())
+			continue
+		}
```
nb-ohad: There is no point in doing this. If you are depending on a label to indicate that this config map is part of your set of endpoint configs you depend on the label. No additional checks are required.

```suggestion
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		sc := &v1alpha1.StorageClient{}
+		sc.Name = ownerRef.Name
+		if err := c.get(sc); err != nil {
+			if kerrors.IsNotFound(err) {
+				c.log.Info("skipping external endpoints ConfigMap, owner StorageClient not found",
+					"configmap", client.ObjectKeyFromObject(cm).String())
+				continue
+			}
+			return nil, fmt.Errorf("get StorageClient %q: %w", ownerRef.Name, err)
+		}
```
nb-ohad: As of the above, this is not important as well, and just extra code that complicate the solution

```suggestion
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+
+		if !sc.GetDeletionTimestamp().IsZero() {
+			c.log.Info("skipping external endpoints ConfigMap, owner StorageClient is being deleted",
+				"configmap", client.ObjectKeyFromObject(cm).String())
+			continue
+		}
```
nb-ohad: This is wrong! let the StorageClient delete the cm and then act on the deletion. don't try to codify the lifecycle of a client and its attached resources into this controller. You are breaking the separation of concerns rule and are coupling the lifetime of clients with the operator config controller logic

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+
+		if sc.UID != ownerRef.UID {
+			c.log.Info("skipping external endpoints ConfigMap with stale StorageClient ownerRef UID",
+				"configmap", client.ObjectKeyFromObject(cm).String(),
+				"ownerName", ownerRef.Name,
+				"ownerUID", ownerRef.UID,
+				"storageClientUID", sc.UID)
+			continue
+		}
```
nb-ohad: Similar to the above this is code that should be eliminated

```suggestion
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+
+		clientUID := string(sc.UID)
+		configKey := nginxProxyConfigKeyPrefix + clientUID + nginxProxyConfigKeySuffix
```
nb-ohad: Why use the clientUID here and not the configmap name or uid? The client here should not be a concern. This code deals with configmaps that bare a known label, nothing more, nothing less

internal/controller/storageclient_controller.go
```diff
@@ -98,6 +98,7 @@ var (
 		&csiopv1.CephConnection{},
 		&csiopv1.ClientProfileMapping{},
 		&corev1.Secret{},
+		&corev1.ConfigMap{},
```
nb-ohad: storage client operator already watches for config maps. there is no point in adding it again here

pkg/console/console.go
```diff
@@ -66,17 +69,29 @@ func GetConsolePlugin(consolePort int32, serviceNamespace string) *consolev1.Con
 				Service: &consolev1.ConsolePluginService{
 					Name:      DeploymentName,
 					Namespace: serviceNamespace,
-					Port:      int32(consolePort),
+					Port:      consolePort,
```
nb-ohad, line 79: What is the reason for this change?

pkg/console/console.go
```diff
@@ -66,17 +69,29 @@ func GetConsolePlugin(consolePort int32, serviceNamespace string) *consolev1.Con
...
+			Proxy: getConsolePluginProxy(consolePort, serviceNamespace),
 		},
 	}
 }
 
-func GetNginxConf() string {
-	return nginxConf
+func GetNginxRootConf() string {
+	return nginxRootConf
+}
```
nb-ohad, line 90: Do you really need a function to return a global?

pkg/console/nginx_conf.go
```diff
@@ -1,6 +1,41 @@
...
+type NginxProxyConfData struct {
+	ClientUID    string
+	ExposeAs     string
+	EndpointURL  string
+	EndpointHost string
+	CertsPath    string
+}
+
+var nginxProxyConf = `
+location /{{.ClientUID}}/{{.ExposeAs}}/ {
```
nb-ohad: Instead of doing this, use text file based embedding with the embed directive
see example here: https://gobyexample.com/embed-directive

pkg/console/nginx_conf.go
```diff
@@ -1,6 +1,41 @@
...
+    proxy_request_buffering off;
+    proxy_buffering off;
+    proxy_ssl_name {{.EndpointHost}};
+    proxy_ssl_verify on;
+    proxy_ssl_server_name on;
+    proxy_ssl_trusted_certificate {{.CertsPath}};
+}
+`
+
+var nginxRootConf = `
```
nb-ohad: Same as the last comment, use the embed directive

pkg/console/nginx_root.conf
nb-ohad, line 1: This entire file is unnecessary. The content should be moved to text file and embedded into the console.go file's global var space using the embed directive. The struct itself should go that file as well and can even be defined inside the function scope to prevent name pollution of the global scope

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+			result[configKey] = content
+		}
+	}
+	return result, nil
+}
+
+func parseEndpointConfigs(data map[string]string) (map[string]endpointConfig, error) {
+	if len(data) == 0 {
+		return nil, nil
+	}
```
nb-ohad: You are running over the items in a loop, this already cover the case of data of len 0

```suggestion
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+	out := make(map[string]endpointConfig, len(data))
+	for exposeAs, rawCfg := range data {
+		exposeAs = strings.TrimSpace(exposeAs)
+		rawCfg = strings.TrimSpace(rawCfg)
```
nb-ohad: If rawCfg is meant to be marshaled byte you should not be trimming anything. Marshalled information should be sent as is to the unmarshal command.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +868,207 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+	secret.Namespace = c.OperatorNamespace
+	secret.Name = externalEndpointCASecretName
+	hasCustomCA := false
+	if err := c.get(secret); err != nil {
+		if !kerrors.IsNotFound(err) {
+			return "", fmt.Errorf("failed to get external endpoint CA secret %q: %w", externalEndpointCASecretName, err)
+		}
+	} else {
+		hasCustomCA = true
+	}
```
nb-ohad:
```suggestion
	if err := c.get(secret); client.IgnoreNotFound(err)
		return "", fmt.Errorf("failed to get external endpoint CA secret %q: %w", externalEndpointCASecretName, err)
	}
	hasCustomCA := secret.UID != ""
```

nb-ohad (changes requested):

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: @nb-ohad as per my testing and understanding, both `/etc/pki/tls/certs/ca-bundle.crt` and `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` serves different purpose.

We use two different trust stores for two different jobs:
1. `/etc/pki/tls/certs/ca-bundle.crt` or `/etc/ssl/certs/ca-bundle.crt` is the system/public bundle from the RHEL/UBI image, which is what we need for nginx to verify external s3 HTTPS backends which are using certs signed by a **public/known CAs** (eg: letsEncrypt etc). This is the default behaviour, I confirmed the bundle path already exists in our pod's image (please ref: https://betterstack.com/community/questions/ssl-cert-location-on-linux/).
And for backends which are using **custom CAs**, we are using an optional Secret mounted at specific path, nginx will use that as its trust store in such cases.

2. `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (OpenShift) is the Service CA for in-cluster service-serving certs, not a substitute for trusting public (external) internet TLS.

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: ack !

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: ack !

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: @nb-ohad it was intentional on my end, having:
1. Primary "For" which handles the operator configmap object and corresponding "Owns" for it's resources (what we have today).
2. Separate "Watches(&corev1.ConfigMap{}, enqueueConfigMapRequest, s3Endpoints...)" which maps S3 CM events back to the fixed operator configmap request (added new in this PR).

To me it looked clearer, but I maybe missing something. Can u plz elaborate on this or share a snippet, I didn't get what should be the new behaviour ??

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated both...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: removed...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: removed...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: removed...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: removed...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: ack, using configmap name instead...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated...

internal/controller/storageclient_controller.go
SanjalKatiyar: AFAIR, when I initially added it, `&corev1.ConfigMap{}` was not part of `kindsToReconcile` entry, guess some other PR added it recently, I removed it now. Thanks for pointing it out.

pkg/console/console.go
SanjalKatiyar: isn't it redundant ??
```
func GetConsolePlugin(consolePort int32, serviceNamespace string) *consolev1.ConsolePlugin {
```
consolePort argument is already `int32` typed, so I will anyway get compile-time error if I tried passing string or anything else, right ??

pkg/console/console.go
SanjalKatiyar: I don't have any strong reasoning for defining `GetNginxRootConf`, but I kept just to be consistent with `GetNginxProxyConf` util (one for returning root configuration and other for the proxy configuration).
Please let me know if that's an issue, I will make `nginxRootConf` global and import directly whereever needed.

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: removed...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: yeah, make sense... my bad...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated...

internal/controller/operatorconfigmap_controller.go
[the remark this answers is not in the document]
SanjalKatiyar: When the nginx configuration changes (ConfigMap updates), the pod does not continuously re-parse configuration files from the disk (or mounted directory). Instead, nginx loads the configuration at startup and continues running with that loaded configuration until it is reloaded or the pod is restarted.

Apart from restarting the pod, another option is to exec into the pod and run `nginx -s reload` (known as "graceful" restart). However, I am not sure if this is recommended Kubernetes controller pattern (i.e., running commands inside the pod during reconcile) ??

I do plan to investigate this further and implement a proper fix. But, for now, given that we are very close to the feature freeze, I request merging this PR (I will resolve all other review comments on this PR though). I will revisit and address the pod restart behaviour in a separate PR after the freeze (as a blocker bug fix).

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3022428768

pkg/console/nginx_root.conf
SanjalKatiyar: updated...

internal/controller/operatorconfigmap_controller.go
leelavg:
```
In fact, For() and Owns() , behind the scenes, are plainly implementations of Watches() only

For example, in our MyController ‘s example:

For(&corev1alpha1.MyCustomResource{}) == Watches(&source.Kind{Type: &corev1alpha1.MyCustomResource{}}, &handler.EnqueueRequestForObject{})

Owns(&corev1.Pod{}) == Watches(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &corev1alpha1.MyCustomResource{}})
```

ref https://yash-kukreja-98.medium.com/develop-on-kubernetes-series-demystifying-the-for-vs-owns-vs-watches-controller-builders-in-c11ab32a046e

internal/controller/operatorconfigmap_controller.go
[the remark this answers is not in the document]
leelavg: why can't a small liveness probe be used to have kubelet restart the container? something naive like, verify the process start and file modification timestamp?

config/console/console_init.yaml
```diff
@@ -58,6 +65,10 @@ spec:
         - name: ocs-client-operator-console-nginx-conf
           configMap:
             name: ocs-client-operator-console-nginx-conf
+        - name: ocs-client-operator-console-s3-endpoint-ca-certs
+          secret:
+            secretName: ocs-client-operator-console-s3-endpoint-ca-certs
+            optional: true
```
leelavg, line 71: should these certs be available before nginx starts or it can refresh during runtime?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -90,8 +93,29 @@ const (
...
 	noobaaCrdName              = "noobaas.noobaa.io"
 	noobaaCrName               = "noobaa-remote"
+
+	// disableS3EndpointProxyKey, if true, disables deploying the s3 endpoint reverse proxy for the local/internal client.
+	disableS3EndpointProxyKey    = "disableS3EndpointProxy"
+	s3EndpointsConfigMapLabelKey = "ocs.openshift.io/hub-s3-endpoints"
+
+	// nginx proxy config key pattern per client: proxy-<clientuid>.conf (all locations for this client in one key).
+	nginxProxyConfigKeyPrefix = "proxy-"
+	nginxProxyConfigKeySuffix = ".conf"
```
leelavg: seeing the usage of this, why can't this be a format `nginxProxyConfigKeyFmt = "proxy-%s.conf"`?

internal/controller/operatorconfigmap_controller.go
leelavg: I believe reads on an empty map is fine, anyways updated code is much better for reading.

internal/controller/operatorconfigmap_controller.go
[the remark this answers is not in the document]
leelavg: > is not reliable as Deployment is currently OLM/CSV-owned.
- I don't think so, or else `kubectl rollout` wouldn't work?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +876,154 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+	}
+}
+
+func (c *OperatorConfigMapReconciler) buildDesiredNginxDataWithProxies() (map[string]string, error) {
+	out := map[string]string{
+		// Root config is mandatory for nginx to start. Proxy configs (per client) are optional.
+		"nginx.conf": console.GetNginxRootConf(),
+	}
+
+	if c.operatorConfigMap.Data != nil {
```
leelavg, line 917: `data` is a map and you can directly read from it without this check.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +876,154 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+		if disableS3EndpointProxy, _ := strconv.ParseBool(c.operatorConfigMap.Data[disableS3EndpointProxyKey]); disableS3EndpointProxy {
```
leelavg: why to silence err, bubble it up, the setting is done by operator on the other side and not human with a default, silencing it here doesn't surface the bug.

config/console/console_init.yaml
SanjalKatiyar: it can be added during runtime... by default (if not present) nginx will use its system's public CA certs (https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3020915903) and during reconcile if this Secret and corresponding key is found we then use that as a trust store...

Secret creation/updation is admin driven manual + optional step...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: initially I was not using this check, but added it later as per https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3018679566

internal/controller/operatorconfigmap_controller.go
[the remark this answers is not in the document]
SanjalKatiyar: My understanding is that a rollout or any pod template based change **might** trigger a pod restart (or create a new ReplicaSet).
As it's owned by OLM, the CSV will eventually revert those changes back to the original state:
- If this reversion happens before the rollout completes, the pod might not restart.
- If it happens after, you could briefly get a new ReplicaSet and then another revert rollout (which ideally resolves our problem as pod indeed gets restarted, but we will have an extra/redundant rollout).

That said, this is **just my assumption, without any actual testing**. That's why I want to investigate further and implement a proper fix. For now, I suggest/request treating this as a bug, as I don't have enough time to investigate it before the feature freeze.

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: Thanks !
What I understand, suggestion here is to remove the "For" and "Owns" and replaces them with "Watches" ?? I made the changes accordingly, please let me know in case that's not what we want and I misunderstood.

internal/controller/operatorconfigmap_controller.go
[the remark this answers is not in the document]
leelavg: you concentrated only on the OLM part but didn't ack the liveness probe?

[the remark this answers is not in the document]
SanjalKatiyar: Haven't looked into it yet, will investigate that and any other possible way after the feature freeze.
I have created `DFBUGS-6311` for fixing it.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -750,28 +825,35 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
 	}
 
+	if nginxConfigMapResult == controllerutil.OperationResultCreated || nginxConfigMapResult == controllerutil.OperationResultUpdated {
+		c.log.Info("nginx ConfigMap data changed, restarting console pod to pick up new config")
+		// Console pod must restart (or nginx must reload) to pick up updated mounted config files.
+		// Restart failures are currently logged only. If restart fails once and config does not change again,
+		// subsequent reconciles will not retry because this block is gated by data-diff check.
+		// ToDo: If issue is prominent, explore reliable retry semantics, like restart based on marker/hash or graceful nginx reload.
+		// Note: checksum rollout via Deployment pod-template annotation is not reliable as Deployment is currently OLM/CSV-owned.
+		c.restartConsolePodsByLabel()
```
Madhu-1, line 854: why do we need to do it? This is generic case where it takes some time for the mounted configures to be visible in the pod.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +886,160 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
...
+			return err
+		}
+		if content != "" {
+			out[configKey] = content
+		}
+	}
+	return nil
+}
+
+func parseEndpointConfigs(data map[string]string) (map[string]s3EndpointConfig, error) {
```
Madhu-1, line 962: please add unit testing

internal/controller/operatorconfigmap_controller.go
```diff
@@ -804,6 +886,160 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) restartConsolePodsByLabel() {
```
Madhu-1, line 889: cant you use ` DeleteAllOf` existing client Method?

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: @Madhu-1 https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3022428768 was my observation... pod's filesystem itself gets the new config file, but nginx keeps on using older one... maybe eventually/randomly it might read again, but it's not very frequent, I did wait for quite a while...

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: @Madhu-1 I am going to remove this entirely as a bug fix: https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r2981056704 after feature freeze and shift to some other approach.

SanjalKatiyar: currently, even though what seems like I am looping over and deleting each pod one by one, in reality we only have one console pod which we never scale up... so it's anyway one API call to the sever only...
and I am going to remove that too... so shift to `DeleteAllOf` will be kind of redundant to do so...

SanjalKatiyar: I have created `DFBUGS-6311` for fixing it.

internal/controller/operatorconfigmap_controller.go
leelavg: sry, `eventually/randomly` isn't a good argument warranting a code for pod restart to put in place, I'm always on the assumption that it didn't work, didn't the nginx docs state the behavior? kubelet would update the data, it's a given, maybe start a nginx locally pointing to a sample configmap? I believe it wouldn't read again because once the server starts something has to intervene to make it re-read or in the config itself mention it should re-read on change etc?

SanjalKatiyar: @leelavg I mentioned "maybe eventually/randomly" only because I haven't looked at the nginx code itself, a system can have it's internal implementation which can cause re-read under certain scenarios. In my testing, which is normal ODF user flow, that never happened. What we really need is consistent and predictable behaviour, so we can ensure nginx picks up config changes reliably/always.

I agree that restarting the pod is not the ideal solution. As I have mentioned earlier, this is just a temporary workaround to enable happy-path testing for QE. I am planning to address this properly after the feature freeze, and this "pod restart" code will not go into production. I have already created a bug for it and will mark it as a blocker.

We don't need to finalise the correct solution right now, we can discuss and decide on the correct solution in a separate PR and I will reply back with more details.

SanjalKatiyar: I get the concern, which is valid and I am indeed going to fix that separately as a bug.

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: updated...

Madhu-1:
LGTM to address the concern as followup, leaving final approval to @leelavg as he reviewed PR

internal/controller/operatorconfigmap_controller.go
```diff
@@ -120,10 +143,17 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				if obj.GetNamespace() != c.OperatorNamespace {
+					return false
+				}
+
+				if obj.GetName() == operatorConfigMapName {
+					return true
+				}
+
+				labels := obj.GetLabels()
+				return labels != nil && labels[s3EndpointsConfigMapLabelKey] == strconv.FormatBool(true)
```
leelavg, line 156: this was the only new addition you require nothing else like rewriting the For/Owns a few lines down.

SanjalKatiyar: @leelavg sorry but I don't get it, I already had this label check earlier as well (when I was using for/owns), it's not a new addition, then I re-wrote because of discussion here: https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3039135026

SanjalKatiyar: as my comment here: https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3021457053

I was already using proper label based predicate for watch on these configmaps...

leelavg: this wasn't needed, ie, a specific watches for configmap, updating the existing predicate was good enough.
``` go
	Watches(
			&corev1.ConfigMap{},
			enqueueConfigMapRequest,
			externalEndpointsConfigMapPredicates,
		).
```

internal/controller/operatorconfigmap_controller.go
leelavg: > it's internal implementation
- I respectfully disagree, I never mentioned to look at implementation. I took my own comment and tried to see how it goes:
1. did a search on nginx reload config and landed at https://nginx.org/en/docs/beginners_guide.html
2. checked for `new config` and read `To apply the new configuration, start nginx if it is not yet started or send the reload signal to the nginx’s master process, by executing: `
3. a sample prompt to add this to yaml manifest got me this
``` yaml
diff --git a/config/console/console_init.yaml b/config/console/console_init.yaml
index f186ca5a..047b4e7f 100644
--- a/config/console/console_init.yaml
+++ b/config/console/console_init.yaml
@@ -23,12 +23,27 @@ spec:
               cpu: 50m
               memory: 256Mi
           livenessProbe:
-            httpGet:
-              path: /plugin-manifest.json
-              port: 9001
-              scheme: HTTPS
-            initialDelaySeconds: 1000
-            periodSeconds: 60
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - |
+                  # Check if nginx config is newer than nginx process and reload if needed
+                  CONFIG_TIME=$(stat -c %Y /opt/app-root/etc/nginx.d/nginx.conf 2>/dev/null || echo 0)
+                  NGINX_PID=$(pgrep -o nginx)
+                  if [ -n "$NGINX_PID" ]; then
+                    NGINX_START=$(stat -c %Y /proc/$NGINX_PID 2>/dev/null || echo 0)
+                    if [ $CONFIG_TIME -gt $NGINX_START ]; then
+                      echo "Config updated, reloading nginx..."
+                      nginx -s reload
+                    fi
+                  fi
+                  # Validate nginx is healthy
+                  nginx -t 2>&1 && curl -k -f https://localhost:9001/plugin-manifest.json
+            initialDelaySeconds: 30
+            periodSeconds: 30
+            timeoutSeconds: 10
+            failureThreshold: 3
           ports:
             - containerPort: 9001
               protocol: TCP
```
I am not saying to take above as is, it's just to show one possibility [hinted](https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3039232544) 2 days ago and above can be tested on a running cluster, like update the csv and edit the configmap and observe.

If 3 reviewers found it odd then some consideration has to be given that a change would be required even before FF. What I totally agree is to the delay in getting the reviews for which I've no answer and should we take it as is, I don't think we can wait further.

leelavg (commented):
/hold
pls provide the test results for posterity and unhold

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: I agree with you on this.

Initially I did some research during readouts (https://docs.google.com/document/d/1yId5AWQ-Og-FXEGIPFK4EKop-KEqffFxC3ZVFUKnWgc/edit?disco=AAABy8Ac1g8), copying my comment from the doc:

> Though on searching online, I see "graceful reload" approach ([https://serverfault.com/questions/378581/nginx-config-reload-without-downtime](https://www.google.com/url?q=https://serverfault.com/questions/378581/nginx-config-reload-without-downtime&sa=D&source=docs&ust=1775659585939872&usg=AOvVaw1SIg2xZGz5L6YFv_9IY71g)) seems possible, it would require the controller to exec into the nginx pod and run commands there. I'm not sure if that's an acceptable pattern for a controller.
> 
> Another option would be something like "inotifywait" ([https://stackoverflow.com/questions/39648684/using-inotifywait-for-nginx-restart](https://www.google.com/url?q=https://stackoverflow.com/questions/39648684/using-inotifywait-for-nginx-restart&sa=D&source=docs&ust=1775659585940005&usg=AOvVaw0AOB9KUaCGdlEzCNKnncbn)), but there's no official image for this in the Red Hat registry and we will have to install it as an additional third-party dependency inside the pod. This doesn't feel ideal to me too (but let me know if u think otherwise).
> 
> Additionally, if we allow users to manually add a CA bundle, we need to ensure the same kind of graceful reload happens in that case as well, without causing UI downtime for other users on the same cluster.

So, yeah there are different ways that we can explore, but I somehow stuck with pod restart at that time (maybe a mistake for not discussing it rigorously before starting the implementation). I acknowledge the review comments and suggestions (and am thankful as well). For time being we can merge current solution so that QE is not blocked with the testing and meanwhile I will fix this asap.

SanjalKatiyar:
> /hold pls provide the test results for posterity and unhold

@leelavg I have updated the description, ptal. I also added an e2e UI recording I captured yesterday on a 4.22 QE cluster.
Since I didn't capture logs at the time, I redeployed the changes on a temporary 4.21 dev cluster I had and included those details as well.

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: had an offline discussion:
- if we had kept `For` and added the extra label check there, the controller would have reconciled the S3 endpoints ConfigMap... this controller is intended to handle only the operator ConfigMap.
- `Owns` depends on `For`... so if we remove `For`, we need to remove `Owns` as well and replace it with the appropriate `Watches` configuration.

leelavg (approved):

internal/controller/operatorconfigmap_controller.go
SanjalKatiyar: https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3052580880

PR #530, opened 2026-03-09
metrics: Enable client side alerting

ShravaniVangur:
This PR creates:
- Alert Collector: to collect the name,labels and value of the alerts that are being retrieved through the GetClientAlerts() RPC call.
- Static PrometheusRules: which map the labels to the rule CRs and build the message/description from labels.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -27,6 +27,7 @@ import (
 
 	apiv1alpha1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/internal/controller"
+	"github.com/red-hat-storage/ocs-client-operator/internal/controller/alert"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/templates"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
 	admwebhook "github.com/red-hat-storage/ocs-client-operator/pkg/webhook"
@@ -65,6 +66,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -308,12 +310,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	alertRunnable := alert.NewRunnable(
+		mgr.GetClient(),
+		utils.GetOperatorNamespace(),
+		ctrl.Log.WithName("alert"),
+		alert.DefaultPollInterval,
+	)
+	if err := mgr.Add(alertRunnable); err != nil {
+		setupLog.Error(err, "unable to add alert runnable to manager")
+		os.Exit(1)
+	}
+
 	if err = (&controller.OperatorConfigMapReconciler{
-		Client:            mgr.GetClient(),
-		Scheme:            mgr.GetScheme(),
-		OperatorNamespace: utils.GetOperatorNamespace(),
-		ConsolePort:       int32(consolePort),
-		AvailableCrds:     availCrds,
+		Client:                  mgr.GetClient(),
+		Scheme:                  mgr.GetScheme(),
+		OperatorNamespace:       utils.GetOperatorNamespace(),
+		ConsolePort:             int32(consolePort),
+		AvailableCrds:           availCrds,
+		UpdateAlertPollInterval: alertRunnable.SetPollInterval,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "OperatorConfigMapReconciler")
 		os.Exit(1)
@@ -338,6 +352,10 @@ func main() {
 			os.Exit(1)
 		}
 	}
+
+	alertCollector := alert.NewCollector(alertRunnable)
+	resourceCollector := alert.NewResourceCollector(mgr.GetClient())
+	metrics.Registry.MustRegister(alertCollector, resourceCollector)
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -65,3 +65,5 @@ resources:
 - ../rbac
 - ../manager
 - ../crd
+- ../prometheus
+- metrics_service.yaml
--- a/go.mod
+++ b/go.mod
@@ -115,7 +115,7 @@ require (
 	github.com/noobaa/noobaa-operator/v5 v5.15.4
 	github.com/nxadm/tail v1.4.8 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_golang v1.23.2 // indirect
+	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/prometheus/common v0.67.3 // indirect
 	github.com/prometheus/procfs v0.17.0 // indirect
--- a/internal/controller/alert/alert.go
+++ b/internal/controller/alert/alert.go
@@ -1,0 +1,172 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package alert
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+	pb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
+	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	OperatorConfigMapName = "ocs-client-operator-config"
+	DefaultPollInterval   = 1 * time.Minute
+)
+
+// Runnable is a manager.Runnable that periodically fetches firing alerts
+// from the storage provider for each connected StorageClient.
+type Runnable struct {
+	client.Client
+	namespace    string
+	logger       logr.Logger
+	pollInterval atomic.Int64
+
+	mu         sync.RWMutex
+	lastAlerts map[string][]*pb.AlertInfo
+}
+
+// NewRunnable creates a new alert Runnable.
+func NewRunnable(c client.Client, namespace string, logger logr.Logger, pollInterval time.Duration) *Runnable {
+	r := &Runnable{
+		Client:     c,
+		namespace:  namespace,
+		logger:     logger,
+		lastAlerts: make(map[string][]*pb.AlertInfo),
+	}
+	r.pollInterval.Store(int64(pollInterval))
+	return r
+}
+
+// SetPollInterval atomically updates the poll interval.
+func (ca *Runnable) SetPollInterval(d time.Duration) {
+	ca.pollInterval.Store(int64(d))
+}
+
+// GetPollInterval atomically reads the poll interval.
+func (ca *Runnable) GetPollInterval() time.Duration {
+	return time.Duration(ca.pollInterval.Load())
+}
+
+// GetAllAlerts returns all currently cached alerts keyed by StorageClient name.
+func (ca *Runnable) GetAllAlerts() map[string][]*pb.AlertInfo {
+	ca.mu.RLock()
+	defer ca.mu.RUnlock()
+	copied := make(map[string][]*pb.AlertInfo, len(ca.lastAlerts))
+	for key, alerts := range ca.lastAlerts {
+		newAlerts := make([]*pb.AlertInfo, len(alerts))
+		copy(newAlerts, alerts)
+		copied[key] = newAlerts
+	}
+	return copied
+}
+
+// Start implements manager.Runnable. It periodically polls for client alerts.
+func (ca *Runnable) Start(ctx context.Context) error {
+	currentInterval := ca.GetPollInterval()
+	ca.logger.V(5).Info("starting client alert runnable", "pollInterval", currentInterval)
+
+	ticker := time.NewTicker(currentInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			ca.logger.V(5).Info("stopping client alert runnable")
+			return nil
+		case <-ticker.C:
+			if newInterval := ca.GetPollInterval(); newInterval != currentInterval {
+				ca.logger.V(5).Info("updating poll interval", "pollInterval", newInterval)
+				ticker.Reset(newInterval)
+				currentInterval = newInterval
+			}
+			ca.fetchAlerts(ctx)
+		}
+	}
+}
+
+// fetchAlerts fetches alerts for all connected StorageClients.
+func (ca *Runnable) fetchAlerts(ctx context.Context) {
+	storageClientList := &v1alpha1.StorageClientList{}
+	if err := ca.List(ctx, storageClientList); err != nil {
+		ca.logger.Error(err, "failed to list StorageClients")
+		return
+	}
+
+	newAlerts := make(map[string][]*pb.AlertInfo, len(storageClientList.Items))
+
+	for i := range storageClientList.Items {
+		sc := &storageClientList.Items[i]
+
+		if sc.Status.Phase != v1alpha1.StorageClientConnected {
+			ca.logger.V(5).Info("skipping StorageClient not in Connected phase", "storageClient", sc.Name, "phase", sc.Status.Phase)
+			continue
+		}
+
+		if sc.Status.ConsumerID == "" {
+			ca.logger.V(5).Info("skipping StorageClient with empty consumerID", "storageClient", sc.Name)
+			continue
+		}
+
+		alerts, err := ca.fetchAlertsForClient(ctx, sc)
+		if err != nil {
+			ca.logger.Error(err, "failed to fetch alerts for StorageClient", "storageClient", sc.Name)
+			continue
+		}
+
+		newAlerts[sc.Name] = alerts
+		ca.logger.V(5).Info("fetched alerts", "storageClient", sc.Name, "alertCount", len(alerts))
+
+		for _, alert := range alerts {
+			ca.logger.V(5).Info("alert detail",
+				"storageClient", sc.Name,
+				"alertName", alert.AlertName,
+				"value", alert.Value,
+				"labels", alert.Labels,
+			)
+		}
+	}
+
+	ca.mu.Lock()
+	ca.lastAlerts = newAlerts
+	ca.mu.Unlock()
+}
+
+// fetchAlertsForClient calls the GetClientAlerts gRPC RPC for a single StorageClient.
+func (ca *Runnable) fetchAlertsForClient(ctx context.Context, sc *v1alpha1.StorageClient) ([]*pb.AlertInfo, error) {
+	ocsProviderClient, err := providerClient.NewProviderClient(
+		ctx, sc.Spec.StorageProviderEndpoint, utils.OcsClientTimeout,
+	)
+	if err != nil {
+		return nil, err
+	}
+	defer ocsProviderClient.Close()
+
+	resp, err := ocsProviderClient.GetClientAlerts(ctx, sc.Status.ConsumerID)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.Alerts, nil
+}
--- a/internal/controller/alert/collector.go
+++ b/internal/controller/alert/collector.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2024 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package alert
+
+import (
+	"sort"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var _ prometheus.Collector = &Collector{}
+
+// Collector exposes alerts fetched from the storage provider as
+// Prometheus metrics. Each firing provider-side alert becomes a gauge:
+//
+//	ocs_client_alert{alert_name="X", severity="Y", storage_client="Z", ...} = value
+//
+// The "alertname" key from the provider labels is renamed to "alert_name" to
+// avoid conflicting with Prometheus' reserved alertname label.
+//
+// Describe sends no descriptors so the collector is registered as "unchecked",
+// which allows dynamic label sets per alert.
+type Collector struct {
+	runnable *Runnable
+}
+
+// NewCollector creates a collector that reads from ca's in-memory alert cache.
+func NewCollector(ca *Runnable) *Collector {
+	return &Collector{runnable: ca}
+}
+
+// Describe implements prometheus.Collector.
+// Empty on purpose -- signals an unchecked collector to the registry because
+// label names vary per alert.
+func (c *Collector) Describe(chan<- *prometheus.Desc) {}
+
+// Collect implements prometheus.Collector.
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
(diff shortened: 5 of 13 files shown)


ShravaniVangur:
Depends on: https://github.com/red-hat-storage/ocs-operator/pull/3730

cmd/main.go
```diff
@@ -356,6 +358,20 @@ func main() {
 		os.Exit(1)
 	}
 
+	clientAlertRunnable := clientalert.NewClientAlert(
```
leelavg: too many `clientAlert` in the name, why can't this be directly inlined with `mgr.Add` and the pkg can just be `alerts` no need for `clientalert` as you are already in `client-op` and the method be `NewRunnable`, so it'll be `mgr.Add(alert.NewRunnable(...))`

internal/controller/client-alert-rules.yaml
```diff
@@ -0,0 +1,98 @@
...
+    - alert: CephOSDNearFull
+      annotations:
+        description: Provider alert CephOSDNearFull is firing for storage client {{ $labels.storage_client }}.
+        message: Ceph OSD near full reported by storage provider for client {{ $labels.storage_client }}.
+        severity_level: warning
+        storage_type: ceph
+      expr: |
+        ocs_client_alert{alert_name="CephOSDNearFull"} > 0
+        or
+        (ceph_osd_metadata * on (ceph_daemon, namespace, managedBy) group_right(device_class,hostname) (ceph_osd_stat_bytes_used / ceph_osd_stat_bytes)) >= 0.75
```
leelavg: how do you get ceph metrics in a remote cluster? same for the rest, or the idea is, when we are running with provider the second part of the rule wins? do we stop alert forwarding if running standalone then?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1184,6 +1219,62 @@ func (c *OperatorConfigMapReconciler) removeNoobaaOperator() error {
 	return nil
 }
 
+const (
+	metricsServiceName        = "ocs-client-operator-metrics"
+	metricsServiceMonitorName = "ocs-client-operator-metrics-monitor"
+	metricsPortName           = "metrics"
+	metricsPort               = 8080
+)
```
leelavg: can move to top level, no need to segment.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1184,6 +1219,62 @@ func (c *OperatorConfigMapReconciler) removeNoobaaOperator() error {
 	return nil
 }
 
+const (
+	metricsServiceName        = "ocs-client-operator-metrics"
+	metricsServiceMonitorName = "ocs-client-operator-metrics-monitor"
+	metricsPortName           = "metrics"
+	metricsPort               = 8080
```
leelavg: this assumes the flag isn't changed, you need to get the flag from main.

internal/controller/alert/alert.go
```diff
@@ -0,0 +1,193 @@
...
+				"storageClient", sc.Name,
+				"alertName", alert.AlertName,
+				"value", alert.Value,
+				"labels", alert.Labels,
+			)
+		}
+	}
+
+	ca.mu.Lock()
+	ca.lastAlerts = newAlerts
```
leelavg, line 152: can use maps.Update, isn't it?

pkg/clientalert/clientalert.go
```diff
@@ -0,0 +1,193 @@
...
+	return &ClientAlert{
+		Client:     c,
+		namespace:  namespace,
+		logger:     logger,
+		lastAlerts: make(map[string][]*pb.AlertInfo),
+	}
+}
+
+// GetLastAlerts returns the most recently fetched alerts for the given StorageClient.
+func (ca *ClientAlert) GetLastAlerts(storageClientName string) []*pb.AlertInfo {
```
leelavg: where is this being used?

internal/controller/alert/resource_collector.go
leelavg, line 1: only cephfs and where is this being used? in another PR to calculate the diff?

pkg/clientalert/clientalert.go
ShravaniVangur: Last alerts are being collected in the ClientAlert and this is not being used. 
Removing this.

internal/controller/alert/alert.go
ShravaniVangur: maps.Update is not a pre-defined function in the maps package.

internal/controller/alert/resource_collector.go
ShravaniVangur: The values are being sent back to the ocs-operator to compute and trigger the alert.

internal/controller/client-alert-rules.yaml
ShravaniVangur: The idea is to avoid duplicate alerts for local cluster.

go.mod
```diff
@@ -1,10 +1,11 @@
 module github.com/red-hat-storage/ocs-client-operator
 
-go 1.25.7
+go 1.24.6
 
 replace (
 	github.com/portworx/sched-ops => github.com/portworx/sched-ops v0.20.4-openstorage-rc3 // required by Rook v1.12
 	github.com/red-hat-storage/ocs-client-operator/api => ./api
+	github.com/red-hat-storage/ocs-operator/services/provider/api/v4 => /Users/home/code/ocs-operator/services/provider/api
```
aruniiird: Remove the line and run `make godeps-update`
```suggestion
```

vendor/modules.txt
```diff
@@ -276,8 +276,8 @@ github.com/red-hat-storage/external-snapshotter/client/v8/apis/volumegroupsnapsh
 # github.com/red-hat-storage/ocs-client-operator/api v0.0.0-00010101000000-000000000000 => ./api
 ## explicit; go 1.24.6
 github.com/red-hat-storage/ocs-client-operator/api/v1alpha1
-# github.com/red-hat-storage/ocs-operator/services/provider/api/v4 v4.0.0-20260331105223-62d56534e987
-## explicit; go 1.25.7
+# github.com/red-hat-storage/ocs-operator/services/provider/api/v4 v4.0.0-20260218112205-69b82919ef95 => /Users/home/code/ocs-operator/services/provider/api
```
aruniiird: All the `/Users/...` local references should be fixed (by itself), when you run `make godeps-update` command after removing the `/User/...` line from go.mod file (change requested in the go.mod file)

internal/controller/client-alert-rules.yaml
```diff
@@ -85,3 +85,16 @@ spec:
...
+      annotations:
+        description: Provider alert CephFSOrphanedSnapshot is firing for storage client {{ $labels.storage_client }}.
+        message: There could be orphaned snapshots that are not claimed or utilized on the storage provider for client {{ $labels.storage_client }}. Please investigate for possible orphaned snapshots.
+        severity_level: warning
+        storage_type: ceph
+      expr: |
+        ocs_client_alert{alert_name="CephFSOrphanedSnapshot"} > 0
+        or
+        ocs_cephfs_snapshot_count > on(consumer_name) ocs_storage_client_cephfs_snapshot_count
+        for: 5s
```
aruniiird: `for` line should come out of the expression block (it should be in the same line as the `expr`)

pkg/alert/alert.go
```diff
@@ -0,0 +1,186 @@
...
+		lastAlerts: make(map[string][]*pb.AlertInfo),
+	}
+}
+
+// GetAllAlerts returns all currently cached alerts keyed by StorageClient name.
+func (ca *Runnable) GetAllAlerts() map[string][]*pb.AlertInfo {
+	ca.mu.RLock()
+	defer ca.mu.RUnlock()
+	copied := make(map[string][]*pb.AlertInfo, len(ca.lastAlerts))
+	maps.Copy(copied, ca.lastAlerts)
```
aruniiird: Can we avoid the `map.Copy()` method here?
Why ?
`map.Copy()` is perfectly good if the key is pointing to a simple/basic types (like int/string/float64 etc)
But here each string key is pointing to a slice (of any type, it doesn't matter), which is a memory reference.
Even though you did a lock and made this copy (map.Copy), the underlying `ca.lastAlerts` slice may change (after the return of this function).

```suggestion
	for key, alerts := range ca.lastAlerts {
		newAlerts := make([]*pb.AlertInfo, len(alerts))
		copy(newAlerts, alerts)
		copied[key] = newAlerts
	}
```

pkg/alert/alert.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+	pb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
+	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	operatorConfigMapName = "ocs-client-operator-config"
+	alertPollIntervalKey  = "alertPollInterval"
```
aruniiird: Can we reuse the exported const `AlertPollIntervalKey` from `internal/controller/operatorconfigmap_controller.go` file

aruniiird (changes requested):
Some initial reviews

internal/controller/alert/collector.go
```diff
@@ -0,0 +1,105 @@
...
+			desc := prometheus.NewDesc(
+				"ocs_client_alert",
+				"Alert relayed from the storage provider",
+				labelNames, nil,
+			)
+
+			value := alert.Value
+			if value == 0 {
+				value = 1
+			}
```
aruniiird, line 67: Currently the `value` variable is used for TWO things,
a. In a normal scenario, it will indicate the VALUE of that alert
b. if the actual VALUE of the alert is ZERO, it is changed to ONE in order to match the condition provided in the `client-alert-rules.yaml` file (eg: `ocs_client_alert{alert_name="CephOSDNearFull"} > 0`

This has a drawback, by looking at the `ocs_client_alert` metric's value, user will never know whether the actual value is ONE or whether it is changed to ONE (from ZERO). User may have to look into alert's `value` label to see the actual value (and that should still be fine).

At present, we don't have to change anything here (IMO), but I would suggest to change the conditions in the `client-alert-rules.yaml` file to `ocs_client_alert{alert_name="<alert_name>"} != 0` (instead of the current `greater-than` ZERO). This can also go in as bug fix.
Why this change?
Theoretically `ocs_client_alert` metric is a `Gauge` metric (see the below line#69) which can go up or down (even below ZERO). So (for correctness and a) safer side will be to change the condition to `not-equal-to` ZERO.

cmd/main.go
```diff
@@ -117,15 +119,16 @@ func main() {
...
 	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.BoolVar(&secureMetrics, "metrics-secure", true,
+	flag.BoolVar(&secureMetrics, "metrics-secure", false,
```
Madhu-1: why this is set to false?

cmd/main.go
```diff
@@ -117,15 +119,16 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var consolePort int
+	var metricsPort int
 	var webhookPort int
 	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metrics endpoint binds to. "+
```
Madhu-1: this should be :8443 for secure access right?

cmd/main.go
```diff
@@ -134,6 +137,7 @@ func main() {
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics")
 	flag.IntVar(&webhookPort, "webhook-port", 7443, "The port the webhook sever binds to.")
+	flag.IntVar(&metricsPort, "metrics-port", 8080, "The port where the metrics endpoint is exposed.")
```
Madhu-1: why do we need one more variable for port? isnt `metrics-bind-address` enough?

internal/controller/client-alert-rules.yaml
Madhu-1: Are we moving all the alters from the ocs-operator to the client operator? these alerts doesnt make sense for remote client, its not per client its per storage cluster

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1194,6 +1234,55 @@ func (c *OperatorConfigMapReconciler) removeNoobaaOperator() error {
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) reconcileMetricsService() error {
```
Madhu-1: isnt this service need to be part of the yaml/bundle not to create using code?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1194,6 +1234,55 @@ func (c *OperatorConfigMapReconciler) removeNoobaaOperator() error {
...
+				Name:     metricsPortName,
+				Port:     c.MetricsPort,
+				Protocol: corev1.ProtocolTCP,
+			},
+		}
+		return nil
+	})
+}
+
+func (c *OperatorConfigMapReconciler) reconcileMetricsServiceMonitor() error {
```
Madhu-1: same for this one as well. why this is not part of the bundle?

pkg/alert/alert.go
```diff
@@ -0,0 +1,189 @@
...
+				interval = newInterval
+				ticker.Reset(interval)
+			}
+		}
+	}
+}
+
+// getPollInterval reads the poll interval from the operator ConfigMap.
+// Returns defaultPollInterval if the ConfigMap or key is not found or cannot be parsed.
+func (ca *Runnable) getPollInterval(ctx context.Context) time.Duration {
```
Madhu-1: helper for each key doesnt make sense please keep it generic, read from configmap once

pkg/alert/alert.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			ca.logger.Error(err, "failed to fetch alerts for StorageClient", "storageClient", sc.Name)
+			continue
+		}
+
+		newAlerts[sc.Name] = alerts
+		ca.logger.V(1).Info("fetched alerts", "storageClient", sc.Name, "alertCount", len(alerts))
+
+		for _, alert := range alerts {
+			ca.logger.V(1).Info("alert detail",
```
Madhu-1: this should be log level 5 not 1 , the log 1 should log only minimal details not trace or debug details

internal/controller/alert/collector.go
Madhu-1: @aruniiird any changes required for this one?

pkg/alert/resource_collector.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Collect implements prometheus.Collector.
+func (c *ResourceCollector) Collect(ch chan<- prometheus.Metric) {
+	ctx := context.Background()
+
+	c.collectCephFsPVCount(ctx, ch)
+	c.collectVSCCount(ctx, ch)
+}
+
+func (c *ResourceCollector) collectCephFsPVCount(ctx context.Context, ch chan<- prometheus.Metric) {
```
Madhu-1: keep this generic and pass the driver name as input

pkg/alert/resource_collector.go
```diff
@@ -0,0 +1,105 @@
...
+		nil, nil,
+	)
+	m, err := prometheus.NewConstMetric(desc, prometheus.GaugeValue, float64(count))
+	if err != nil {
+		return
+	}
+	ch <- m
+}
+
+func (c *ResourceCollector) collectVSCCount(ctx context.Context, ch chan<- prometheus.Metric) {
```
Madhu-1: try to keep this as generic as well

internal/controller/alert/collector.go
aruniiird: At present no changes are required. Let it be like this

cmd/main.go
```diff
@@ -119,13 +122,13 @@ func main() {
 	var consolePort int
 	var webhookPort int
 	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metrics endpoint binds to. "+
```
Madhu-1: no need to change this one as the flag is set in the deployment now

cmd/main.go
```diff
@@ -119,13 +122,13 @@ func main() {
...
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.BoolVar(&secureMetrics, "metrics-secure", true,
+	flag.BoolVar(&secureMetrics, "metrics-secure", false,
```
Madhu-1: lets not change the default in the flag, set this to required field in deployment.yaml

cmd/main.go
```diff
@@ -339,6 +342,35 @@ func main() {
...
+	cm.Namespace = utils.GetOperatorNamespace()
+	if err := apiClient.Get(context.Background(), client.ObjectKeyFromObject(cm), cm); err == nil {
+		if val := cm.Data[controller.AlertPollIntervalKey]; val != "" {
+			if d, err := time.ParseDuration(val); err == nil {
+				alertPollInterval = d
+			} else {
+				setupLog.Error(err, "failed to parse alertPollInterval, using default", "value", alert.DefaultPollInterval)
+			}
+		}
+	}
```
leelavg: nope, the idea was to load configmap during reconcile and reuse but not to load even before controller starts, there is no diff b/n doing above and using env.

the implicit expectation that a value will be set when you read the cm is wrong as the controllers in ocs-op is not synced to controllers in client-op.

config/default/kustomization.yaml
```diff
@@ -18,9 +18,9 @@ namePrefix: ocs-client-operator-
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 #- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-#- ../prometheus
+# - ../prometheus (moved to resources below)
 # [METRICS] Expose the controller manager metrics service.
-# - metrics_service.yaml
+# - metrics_service.yaml (moved to resources below)
```
leelavg: not required to add one more comment, we can see in the diff, just remove the lines.

Madhu-1 (commented):
small nits, LGTM @leelavg PTAL

internal/controller/storageclient_controller.go
```diff
@@ -166,10 +186,8 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			slices.Contains(csiDrivers, vsc.Spec.Driver) &&
 			vsc.Status != nil &&
 			vsc.Status.SnapshotHandle != nil {
-			parts := strings.Split(*vsc.Status.SnapshotHandle, "-")
-			if len(parts) == 9 {
-				// second entry in the volumeID is clusterID which is unique across the cluster
-				return []string{parts[2]}
+			if clusterID := extractClusterID(*vsc.Status.SnapshotHandle); clusterID != "" {
+				return []string{clusterID}
```
leelavg, line 190: what's the problem with existing code?

internal/controller/alert/alert.go
```diff
@@ -0,0 +1,154 @@
...
+	ticker := time.NewTicker(ca.pollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			ca.logger.V(5).Info("stopping client alert runnable")
+			return nil
+		case <-ticker.C:
+			ca.fetchAlerts(ctx)
```
leelavg, line 104: there used to be a code resetting the ticker if the value in cm changes, isn't it?

internal/controller/alert/resource_collector.go
```diff
@@ -0,0 +1,153 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// These must match the index names registered in
+// internal/controller/storageclient_controller.go SetupWithManager.
+const (
+	ownerUIDIndexName     = "index:ownerUID"
+	pvClusterIDIndexName  = "index:persistentVolumeClusterID"
+	vscClusterIDIndexName = "index:volumeSnapshotContentCSIDriver"
+)
```
leelavg, line 38: fragile, all these files should be in `internal/alerts/`

internal/alerts/resource_collector.go
internal/alerts/federated_collector.go
internal/alerts/runnable.go <- registered in main and should act like a controller.

internal/controller/alert/collector.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// NewCollector creates a collector that reads from ca's in-memory alert cache.
+func NewCollector(ca *Runnable) *Collector {
+	return &Collector{runnable: ca}
+}
+
+// Describe implements prometheus.Collector.
+// Empty on purpose -- signals an unchecked collector to the registry because
+// label names vary per alert.
```
leelavg, line 48: this needs a bit more discussion, not immediate but the way we add many labels it'll increase the cardinality, fine at lower numbers though.

internal/controller/storageclient_controller.go
ShravaniVangur: In situations where the clusterID that has a hyphen as a part of it, the check would split the handle into 10 parts and not 9. Since there is no error thrown, this is skipped while caching.

cmd/main.go
ShravaniVangur: Yes the reconcile was the missing piece. Updated the code to reflect the same.

internal/controller/alert/alert.go
ShravaniVangur: Added that with the reconcile change.

internal/controller/alert/resource_collector.go
ShravaniVangur: Moved the files

internal/controller/alert/alert.go
leelavg: I'm mentioning to have that code rather than inferring from configmap_controller.

internal/controller/alert/resource_collector.go
leelavg: last, is there any reason that all three files can't be in internal/controller/alert.go? the problem with above is duplicate indexnames and their field registration, could you pls raise a GH issue and fix this after FF?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -352,6 +360,16 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 
+	alertPollInterval := alert.DefaultPollInterval
+	if val := c.operatorConfigMap.Data[AlertPollIntervalKey]; val != "" {
+		if d, err := time.ParseDuration(val); err == nil {
+			alertPollInterval = d
+		} else {
+			c.log.Error(err, "failed to parse alertPollInterval, using default", "value", alertPollInterval)
+		}
+	}
+	c.AlertRunnable.PollInterval = alertPollInterval
```
leelavg: possible data race, this and alert controller runs concurrently. possible to do below?
1. make PollInterval an atomic uint32
2. create a set and get method for this var in runnable
3. to cm_controller make the last field `UpdateAlertPollInterval func()` and supply `alertRunnable.UpdatePollInterval`
4. when you read from the configmap, do `c.UpdateAlertPollInterval(duration)`?

ShravaniVangur:
>last, is there any reason that all three files can't be in internal/controller/alert.go? the problem with above is duplicate indexnames and their field registration, could you pls raise a GH issue and fix this after FF? 

These variables are shared between the storageclient_controller.go and the alerts. 
All the alert related files have been moved to internal/controllers/alerts. 
Changing the const in storageclient_controller.go would create more changes now. Will address this after FF.

internal/controller/operatorconfigmap_controller.go
ShravaniVangur: Updated the flow.

ShravaniVangur:
Testing output: 

Metrics: 

1. Service, ServiceMonitor, and Metrics Endpoint Validation:

```
{
    "metric": {
        "__name__": "up",
        "endpoint": "metrics",
        "instance": "xx.xxx.x.xx:xxxxx", (matching the set instance),
        "job": "ocs-client-operator-metrics",
        "namespace": "openshift-storage",
        "pod": "ocs-client-operator-controller-manager-646f76ddd8-dsk89",
        "service": "ocs-client-operator-metrics"
    },
    "value": [1775672952.712, "1"]
}
```
2. ocs_client_operator_cephfs_volume_snapshot_content_count metric:
```
{
    "metric": {
        "__name__": "ocs_client_operator_cephfs_volume_snapshot_content_count",
        "endpoint": "metrics",
        "instance": "xx.xxx.x.xx:xxxxx",
        "job": "ocs-client-operator-metrics",
        "namespace": "openshift-storage",
        "pod": "ocs-client-operator-controller-manager-8bf9789f9-n5np8",
        "service": "ocs-client-operator-metrics",
        "storage_client": "ocs-storagecluster"
    },
    "value": [1775675716.141, "1"]
}
```
3. ocs_client_operator_cephfs_volume_snapshot_content_count metric:
```
{
    "metric": {
        "__name__": "ocs_client_operator_cephfs_volume_snapshot_content_count",
        "endpoint": "metrics",
        "instance": "xx.xxx.x.xx:xxxxx",
        "job": "ocs-client-operator-metrics",
        "namespace": "openshift-storage",
        "pod": "ocs-client-operator-controller-manager-8bf9789f9-n5np8",
        "service": "ocs-client-operator-metrics",
        "storage_client": "ocs-storagecluster"
    },
    "value": [1775675716.141, "1"]
}
```
4. Alert 
```
ocs_client_alert{alert_name="CephFSOrphanedSnapshot", consumer_name="internal", severity="warning", storage_client="ocs-storagecluster"} 1
```

leelavg (approved):

PR #545, opened 2026-04-02
godeps: Update go version to 1.25.7 and toolchain to 1.25.8

karthikps97:
### Steps followed for update:
1. go get go@1.25.7 toolchain@1.25.8
2. cd api/ && go get go@1.25.7 toolchain@1.25.8
4. make godeps-update
5. Updated Dockerfile to match go version

**Note**
One of the dependecies (sigs.k8s.io/controller-runtime/tools/setup-envtest) requires go 1.26
But the toolchain version is fixed at 1.25.z. This causes go to throw the error `go: no such tool "covdata"` during `make container-build`

Workaround: Add an env in which sets `GOTOOLCHAIN` to `1.25.0+auto`, which would use the specified value before trying to switch versions.


karthikps97:
Fixed in #548 , closing this.

PR #546, opened 2026-04-02
Add support to update sub-resource on client clusters

karthikps97:
### Overview:
Add mechanism to update sub resource of kube objects on client cluster.
This is a continuation of the changes in ocs-operator [PR link](https://github.com/red-hat-storage/ocs-operator/pull/3747)

Explain the changes:
1. Update provider client to include the changes to `GetDesiredClientState` RPC.
3. Fetch the resource which needs to be updated 
4. Merge the desired state onto clone of the existing and check whether `update` is necessary.
5. Perform update if state has changed.

```sh
kubectl apply -f - <<EOF
apiVersion: objectbucket.io/v1alpha1
kind: ObjectBucketClaim
metadata:
  name: obc-test
  namespace: openshift-storage-client
spec:
  generateBucketName: obc-test
  storageClassName: openshift-storage.noobaa.io
EOF
```

OBC before status update:
```yaml
apiVersion: objectbucket.io/v1alpha1
kind: ObjectBucketClaim
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"objectbucket.io/v1alpha1","kind":"ObjectBucketClaim","metadata":{"annotations":{},"name":"obc-test","namespace":"openshift-storage-client"},"spec":{"generateBucketName":"obc-test","storageClassName":"openshift-storage.noobaa.io"}}
  creationTimestamp: "2026-04-02T10:35:03Z"
  finalizers:
  - ocs.openshift.io/obccleanup
  generation: 2
  labels:
    ocs.openshift.io/storageclient.name: openshift-storage-client
  name: obc-test
  namespace: openshift-storage-client
  resourceVersion: "13979293"
  uid: be058e6a-5ab5-435b-90bc-814fdd2adaf8
spec:
  bucketName: ""
  generateBucketName: obc-test
  storageClassName: openshift-storage.noobaa.io
```

OBC after status update:
```yaml
apiVersion: objectbucket.io/v1alpha1
kind: ObjectBucketClaim
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"objectbucket.io/v1alpha1","kind":"ObjectBucketClaim","metadata":{"annotations":{},"name":"obc-test","namespace":"openshift-storage-client"},"spec":{"generateBucketName":"obc-test","storageClassName":"openshift-storage.noobaa.io"}}
  creationTimestamp: "2026-04-02T10:35:03Z"
  finalizers:
  - ocs.openshift.io/obccleanup
  generation: 2
  labels:
    ocs.openshift.io/storageclient.name: openshift-storage-client
  name: obc-test
  namespace: openshift-storage-client
  resourceVersion: "13979768"
  uid: be058e6a-5ab5-435b-90bc-814fdd2adaf8
spec:
  bucketName: ""
  generateBucketName: obc-test
  storageClassName: openshift-storage.noobaa.io
status:
  phase: Bound
```

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -44,12 +44,14 @@ import (
 	quotav1 "github.com/openshift/api/quota/v1"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	odfgsapiv1b1 "github.com/red-hat-storage/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
+	provider "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
 	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -135,12 +137,14 @@ var (
 	}
 )
 
-type desiredKubeObject struct {
+type kubeObjectWithOpRecord struct {
 	types.NamespacedName
-	bytes []byte
+	bytes       []byte
+	operation   provider.KubeClientOp
+	subResource *provider.SubResource
 }
 
-type desiredKubeObjects []desiredKubeObject
+type kubeObjectWithOpRecords []kubeObjectWithOpRecord
 
 // StorageClientReconciler reconciles a StorageClient object
 type StorageClientReconciler struct {
@@ -281,6 +285,8 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=objectbucket.io,resources=objectbucketclaims,verbs=get;list;watch
 //+kubebuilder:rbac:groups=objectbucket.io,resources=objectbuckets,verbs=get;list;watch;update;create;delete
 //+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=objectbucket.io,resources=objectbucketclaims/status,verbs=update;patch
+//+kubebuilder:rbac:groups=objectbucket.io,resources=objectbuckets/status,verbs=update;patch
 
 func (r *StorageClientReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	handler := storageClientReconcile{StorageClientReconciler: r}
@@ -542,7 +548,7 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 
 	r.storageClient.Status.InMaintenanceMode = storageClientResponse.MaintenanceMode
 
-	kubeObjectsByGk := map[string]desiredKubeObjects{}
+	kubeObjectsByGk := map[string]kubeObjectWithOpRecords{}
 	for _, kubeObj := range storageClientResponse.KubeObjects {
 		if kubeObj == nil {
 			continue
@@ -552,12 +558,17 @@ func (r *storageClientReconcile) reconcilePhases() (ctrl.Result, error) {
 			return reconcile.Result{}, fmt.Errorf("failed to unmarshal metadata for the Object: %w", err)
 		}
 		gk := objectMeta.GroupVersionKind().GroupKind().String()
+
+		record := kubeObjectWithOpRecord{
+			NamespacedName: client.ObjectKeyFromObject(objectMeta),
+			bytes:          kubeObj.Bytes,
+			operation:      kubeObj.Op,
+			subResource:    kubeObj.SubResource,
+		}
+
 		kubeObjectsByGk[gk] = append(
 			kubeObjectsByGk[gk],
-			desiredKubeObject{
-				NamespacedName: client.ObjectKeyFromObject(objectMeta),
-				bytes:          kubeObj.Bytes,
-			},
+			record,
 		)
 	}
 	var combinedErr error
@@ -910,7 +921,7 @@ func (r *storageClientReconcile) own(dependent metav1.Object) error {
 
 func (r *storageClientReconcile) reconcileResourcesByGK(
 	kind client.Object,
-	desiredObjects map[string]desiredKubeObjects,
+	desiredObjects map[string]kubeObjectWithOpRecords,
 	combinedErr *error,
 ) {
 	gvk, err := apiutil.GVKForObject(kind, r.Scheme)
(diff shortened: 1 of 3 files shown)


karthikps97:
Changes also include RBAC changes (Update/Patch) for OB/status and OBC/status

karthikps97:
@nb-ohad , Could you please take a look.

internal/controller/storageclient_controller.go
```diff
@@ -905,10 +916,18 @@ func (r *storageClientReconcile) reconcileResourcesByGK(
...
+			} else {
+				reconciledObjects[desiredState.NamespacedName] = true
+			}
 		} else {
-			reconciledObjects[desiredState.NamespacedName] = true
+			if err := r.reconcileResource(kubeObject, desiredState); err != nil {
+				multierr.AppendInto(combinedErr, err)
+			} else {
+				reconciledObjects[desiredState.NamespacedName] = true
+			}
```
nb-ohad: As we are dealing with an enum, it make more sense to have this code as a switch and not an if/else.
In addition, you should not be assuming that if the op is not update sub resource then it must be createOrUpdate. It is the wrong assumption, you should be doing an exact check for explicit ops and ignore/skip (with an log message) ops that are unknown. 

This is very important when we will add additional ops in the future and need backport compatibility with older clients who do not support these ops.

internal/controller/storageclient_controller.go
```diff
@@ -929,7 +948,7 @@ func (r *storageClientReconcile) reconcileResourcesByGK(
 	}
 }
 
-func (r *storageClientReconcile) reconcileResource(obj client.Object, desiredState desiredKubeObject) error {
+func (r *storageClientReconcile) reconcileResource(obj client.Object, desiredState kubeObjectWithOpRecord) error {
```
nb-ohad: Why do we need this change? why can't you just send the namespace name and the bytes? At this point all of the other fields are meaningless

internal/controller/storageclient_controller.go
```diff
@@ -992,6 +1011,55 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, desiredSta
 	return nil
 }
 
+// reconcileSubResource performs an update of the sub-resource specified in desiredState
+func (r *storageClientReconcile) reconcileSubResource(desiredState kubeObjectWithOpRecord, kind client.Object) error {
```
nb-ohad: Don't send kubeObjectWithOpRecord. just send the stuff this function is interested in which is the namespaced name, the bytes and the sub resource name

internal/controller/storageclient_controller.go
```diff
@@ -1036,3 +1104,84 @@ func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
...
+}
+
+// getSubResourceString returns the name for the passed sub resource enum
+func getSubResourceString(subResourceEnum provider.SubResource) (subResource string) {
+	switch subResourceEnum {
+	case provider.SubResource_SUB_RESOURCE_STATUS:
+		subResource = "Status"
+	}
+	return
+}
```
nb-ohad: It seems to be an overkill to have a function for this. Even when reviewing I had to "jump" here to understand the code on the call site. In addition you are missing handling cases for invalid sub-resources

internal/controller/storageclient_controller.go
```diff
@@ -992,6 +1011,55 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, desiredSta
...
+
+	kubeObject := getKubeObject(kind)
+	subResBytes, err := json.Marshal(subRes)
+	if err != nil {
+		return fmt.Errorf("error while marshaling sub-resource, error: %w", err)
+	}
+
+	if err := json.Unmarshal(subResBytes, kubeObject); err != nil {
+		return fmt.Errorf("failed to unmarshal sub-resource into object, error:  %w", err)
+	}
```
nb-ohad: It make no sense to Unmarshal just to Marshal again then to Unmarshal again. 
You just need to Unmarshal the bytes once and use a sub struct of that struct (the sub resources) as your source for the update

internal/controller/storageclient_controller.go
```diff
@@ -1036,3 +1104,84 @@ func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
 	}
 	return false
 }
+
+// extractSubResource extracts the specified sub-resource from the passed kubeObject
+func extractSubResource(kubeObject client.Object, subResourceName string) (map[string]any, error) {
+	val := reflect.Indirect(reflect.ValueOf(kubeObject))
```
nb-ohad: Why do we need an indirect if kubeobject is an interface?

internal/controller/storageclient_controller.go
```diff
@@ -1036,3 +1104,84 @@ func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
...
+	// get the field by name (e.g., "Status")
+	field := val.FieldByName(subResourceName)
+
+	if !field.IsValid() {
+		return nil, fmt.Errorf("invalid value for the kind %v", val)
+	}
+
+	subResource := map[string]any{
+		strings.ToLower(subResourceName): field.Interface(),
+	}
```
nb-ohad: Why do you need to return a map and not just an any, or a pointer to an any?

internal/controller/storageclient_controller.go
```diff
@@ -1036,3 +1104,84 @@ func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
 	}
 	return false
 }
+
+// extractSubResource extracts the specified sub-resource from the passed kubeObject
+func extractSubResource(kubeObject client.Object, subResourceName string) (map[string]any, error) {
```
nb-ohad: This function has the assumption that the sub resource is a top level field. You cannot make this assumption. I would also not recommend to try implementing this logic in a generic way for all sub resources. Instead, I would suggest to have a specific logic for each sub resource type then have a switch based on the type to invoke to correct logic (which right now is just for the "status" sub resource)

CC @leelavg

internal/controller/storageclient_controller.go
```diff
@@ -1036,3 +1104,84 @@ func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
...
+	if equality.Semantic.DeepEqual(existingSubResource, subResource) {
+		r.log.Info(
+			"Skipping as no change to sub resource",
+			"name",
+			client.ObjectKeyFromObject(obj),
+		)
+		return nil
+	}
+
+	toUpdate.SetResourceVersion(obj.GetResourceVersion())
```
nb-ohad: You should not be updating resourceVersion manually, leave the resource version for the kubeapi to handle

internal/controller/storageclient_controller.go
```diff
@@ -1036,3 +1104,84 @@ func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
...
+	}
+	return nil
+}
+
+// getKubeObject returns a new instance of the same Kubernetes object type as the provided kind
+func getKubeObject(kind client.Object) client.Object {
+	goType := reflect.TypeOf(kind).Elem()
+	untypedInstance := reflect.New(goType).Interface()
+	return untypedInstance.(client.Object)
+}
```
nb-ohad: Why to we need a function for this? and more importantly, why do you need to use reflect to get an object kind where there are already built-in utils that do that?

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
```diff
@@ -931,10 +942,27 @@ func (r *storageClientReconcile) reconcileResourcesByGK(
...
+				multierr.AppendInto(combinedErr, err)
+			} else {
+				reconciledObjects[desiredState.NamespacedName] = true
+			}
+		case provider.KubeClientOp_CREATE_OR_UPDATE:
+			if err := r.reconcileResource(kubeObject, desiredState.bytes, desiredState.NamespacedName); err != nil {
+				multierr.AppendInto(combinedErr, err)
+			} else {
+				reconciledObjects[desiredState.NamespacedName] = true
+			}
```
nb-ohad: Can you please move this to be the first on case as this is the common case (for further readability)

internal/controller/storageclient_controller.go
```diff
@@ -1018,6 +1046,68 @@ func (r *storageClientReconcile) reconcileResource(obj client.Object, desiredSta
...
+		return fmt.Errorf("error unmarshaling desired kubeobject: %v. error is %w", namespacedName, err)
+	}
+
+	// compare both the objects and skip update if no change in state
+	if equality.Semantic.DeepEqual(kubeObject, desiredObject) {
+		r.log.Info(
+			"Skipping as no change to sub resource",
+			"name",
+			client.ObjectKeyFromObject(kubeObject),
+		)
```
nb-ohad: There is no point in balooning the log for a case where you just skip the update as an optimization. I would prefer if the check will be for !equal then have the update inside the if clause

nb-ohad (changes requested):

PR #548, opened 2026-04-02
vendor: pull latest provider server api & grpc client

leelavg:
provider api supporting updating status subresource, required bumping golang resulting into upgrading controller-gen and golangci-lint.


rewantsoni:
@leelavg noticed it now but the api pkg go version is still at 1.24.6

leelavg:
I don't see a problem with that or else we want to do go work sync / something similar?

PR #555, opened 2026-04-12
Gap - Update `buildCacheAvailableCRDs` Function

shirady:
### Describe the problem
It fixes a change that was added in #520 and was raised in PR #539 (see [comment](https://github.com/red-hat-storage/ocs-client-operator/pull/539#issuecomment-4154280336)), which was mentioned as a gap in the PR description of #539.

### Explain the changes
1. In main - Update `buildCacheAvailableCRDs` function:
- any config map and secret in the operator namespace
- any config map and secret with label key app and label value noobaa in all namespaces

### Testing Instructions:
1. Manual Test: Only tested that the manager got started with the code changes.
2. Automatic test: Used an AI-generated script for listing OBC creation - so we would have a secret and a configmap with a label key app and value noobaa that would be listed. Attaching output:
```
cache-scope-test: list API checks passed
operator namespace: "openshift-storage-client"
  ConfigMaps (all): 20
  Secrets    (all): 23
cluster-wide with label "app=noobaa":
  ConfigMaps: 1
  Secrets:    1
distinct ConfigMap keys (operator-ns all ∪ cluster app=noobaa): 21
distinct Secret keys    (operator-ns all ∪ cluster app=noobaa): 24
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -55,6 +55,7 @@ import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


shirady:
@rewantsoni @leelavg @nb-ohad , please take a look

cmd/main.go
```diff
@@ -380,20 +379,30 @@ func getAvailableCRDNames(ctx context.Context, cl client.Client) (map[string]boo
...
-	subscriptionwebhookSelector fields.Selector,
 	defaultNamespaces map[string]cache.Config,
+	operatorNamespace string,
 ) cache.Options {
+	subscriptionwebhookSelector := fields.SelectorFromSet(fields.Set{"metadata.name": templates.SubscriptionWebhookName})
+	noobaaLabelSelector := labels.SelectorFromSet(labels.Set{"app": "noobaa"})
+	configMapAndSecretCacheByNamespace := map[string]cache.Config{
+		operatorNamespace: {
+			LabelSelector: labels.Everything(),
+		},
```
rewantsoni: Why do we need to be specific here and not default to `{}`?

shirady: I can change that, but there is a difference in case someone would set additional selector - using `labels.Everything()` is not nil and would scope everything, while `{}` behaves as nil, and then the tighter scope will be filtered.

leelavg (approved):

PR #556, opened 2026-04-13
Fix - Avoid Pod Deletion When NooBaa OBC CRD Is Installed

shirady:
### Describe the problem
It fixes a change that was added in #539 (see [comment](https://github.com/red-hat-storage/ocs-client-operator/pull/539#issuecomment-4166071032)), which was mentioned as a gap in the PR description.

Note: I could not have a full dynamic OBC CRD and OBC watch since the cache was set at start, from what I checked, it can be done, but then I would have a different cache for the OBCs to be watched, and it gets more complicated when we need to GET or UPDATE the OBCs. To avoid pod deletion, I used process exit - reference:
https://github.com/red-hat-storage/ocs-client-operator/blob/dd10c95aad5569c6d235b167f3febf36ffd6fa5c/internal/controller/operatorconfigmap_controller.go#L349

### Explain the changes
1. In main - set the context with cancel and defer the cancel, define the function for container shutdown that would use the cancel call. We're using the cancel instead of the previous exit code because we want a graceful exit. 
2. Added a new controller that is triggered by created and deleted events only of CRDs with an opt-in list, because we want, wherever possible, to use dynamic watch (see example of OB in storage client controller) - it is the solution where we have a controller that has its main watch on a CR - OBC controller and MaintenanceMode controller.
3. Remove the process exit that we had in operatorconfigmap_controller related to MaintenanceMode CRD presence change.

#### Manual Test:
1. Client cluster: Check if you have OBC CRD on the cluster (if you have - delete them): `oc get crd | grep objectbucket.io`
2. Check the status of the pod - should be running (`oc get pods -n openshift-storage-client | grep ocs-client-operator-controller-manager`)
3. Install NooBaa OBC and OB CRDs:
- `kubectl apply -f noobaa-operator/deploy/obc/objectbucket.io_objectbucketclaims_crd.yaml.`
- `kubectl apply -f noobaa-operator/deploy/obc/objectbucket.io_objectbuckets_crd.yaml`
This can also be done with ODF-CLI: `bin/odf object enable remote-obc -n openshift-storage-client` (I build the CLI for arm arch)
4. You would see in the logs that the container was shut down and the pod was restarted (see [comment](https://github.com/red-hat-storage/ocs-client-operator/pull/556#issuecomment-4280160424)).
5. Apply OBC yaml and check in the logs that the reconcile was started and Notify was called (you can also check on the provider cluster that the OBC was created).

Note: I also downloaded MaintenanceMode CRD ([link](https://github.com/RamenDR/ramen/blob/main/config/crd/bases/ramendr.openshift.io_maintenancemodes.yaml)) and applied it to see the container shutdown using `kubectl apply -f ~/Downloads/ramendr.openshift.io_maintenancemodes.yaml`

### DIFF
--- a/internal/controller/crds_presence_controller.go
+++ b/internal/controller/crds_presence_controller.go
@@ -1,0 +1,72 @@
+package controller
+
+import (
+	"context"
+	"slices"
+
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+
+	"github.com/go-logr/logr"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// crdsWatchedForPresenceRestart lists CRDs whose install/removal must match process startup
+// (manager cache and conditional controller registration; see cmd/main.go).
+// IMPORTANT - only add the cases where the dynamic watch for the CRD did not match the case.
+var crdsWatchedForPresenceRestart = []string{
+	ObjectBucketClaimCrdName,
+	MaintenanceModeCRDName,
+}
+
+type CrdsPresenceReconciler struct {
+	client.Client
+	AvailableCrds     map[string]bool
+	ShutdownContainer func()
+	log               logr.Logger
+}
+
+func (r *CrdsPresenceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("CrdsPresence").
+		For(
+			&extv1.CustomResourceDefinition{},
+			builder.WithPredicates(
+				predicate.NewPredicateFuncs(func(obj client.Object) bool {
+					return slices.Contains(crdsWatchedForPresenceRestart, obj.GetName())
+				}),
+				utils.EventTypePredicate(true, false, true, false),
+			),
+		).
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+
+func (r *CrdsPresenceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.log = ctrl.LoggerFrom(ctx).WithName("CrdsPresence")
+
+	for _, name := range crdsWatchedForPresenceRestart {
+		crd := &metav1.PartialObjectMetadata{}
+		crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+		crd.Name = name
+		if err := r.Get(ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
+			r.log.Error(err, "Failed to get CRD", "CRD", crd.Name)
+			return ctrl.Result{}, err
+		}
+		presentNow := crd.UID != ""
+		if r.AvailableCrds[name] != presentNow {
+			r.log.Info("CRD presence changed, will restart container",
+				"presence at manager start", r.AvailableCrds[name],
+				"presence now", presentNow,
+			)
+			r.ShutdownContainer()
+			return ctrl.Result{}, nil
+		}
+	}
+	return ctrl.Result{}, nil
+}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -318,6 +318,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx, cancel := context.WithCancel(ctrl.SetupSignalHandler())
+	defer cancel()
+	shutdownContainer := func() {
+		cancel()
+	}
+
 	if err = (&controller.OperatorConfigMapReconciler{
 		Client:                  mgr.GetClient(),
 		Scheme:                  mgr.GetScheme(),
@@ -340,6 +346,15 @@ func main() {
 		}
 	}
 
+	if err = (&controller.CrdsPresenceReconciler{
+		Client:            mgr.GetClient(),
+		AvailableCrds:     availCrds,
+		ShutdownContainer: shutdownContainer,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "CrdsPresence")
+		os.Exit(1)
+	}
+
 	if availCrds[controller.ObjectBucketClaimCrdName] {
 		if err = (&controller.ObcReconciler{
 			Client: mgr.GetClient(),
@@ -355,7 +370,7 @@ func main() {
 	metrics.Registry.MustRegister(alertCollector, resourceCollector)
 
 	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -322,15 +322,6 @@ func (c *OperatorConfigMapReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	c.log = log.FromContext(ctx, "OperatorConfigMap", req)
 	c.log.Info("Reconciling OperatorConfigMap")
 
-	crd := &metav1.PartialObjectMetadata{}
-	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
-	crd.Name = MaintenanceModeCRDName
-	if err := c.Get(ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
-		c.log.Error(err, "Failed to get CRD", "CRD", crd.Name)
-		return reconcile.Result{}, err
-	}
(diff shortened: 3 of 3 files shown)


shirady:
@rewantsoni @leelavg @nb-ohad , please take a look

leelavg:
> I would have a different cache for the OBCs to be watched
- I don't think so, there is only one manager cache that's used by the cached client which allows dynamic watches, possible to explain more on why we need to go back to older method of pod restart?

shirady:
> > I would have a different cache for the OBCs to be watched
> 
> * I don't think so, there is only one manager cache that's used by the cached client which allows dynamic watches, possible to explain more on why we need to go back to older method of pod restart?

@leelavg the manager cache is set in main, at this point, we may be in 2 states:
1. OBC CRD was presented (include all the OBCs in all namespaces).
2. OBC CRD was not presented, and the manager does not contain configurations of OBCs in all namespaces.

https://github.com/red-hat-storage/ocs-client-operator/blob/2878a3b403e69a2a9cdc4fa583d8b05aa26f2fd5/cmd/main.go#L408-L414

If the OBC CRD was not installed, the cache was not configured to watch the OBCs in all namespaces.
After the CRD is installed later, restarting the process re-runs discovery and rebuilds cache options so OBCs are watched across all namespaces as intended.

leelavg:
ok, the limitation is not being able to dynamically watch namespaced resource in multiple namespaces, thanks for the explanation.

internal/controller/obc_controller.go
```diff
@@ -92,6 +119,22 @@ func (r *obcReconcile) reconcile(ctx context.Context, req ctrl.Request) (reconci
...
+// AssertEqual when the CRD appears after start; see cmd/main.go). If req is the enqueue key for the
+// OBC CRD itself (not an OBC instance), skip is true so we do not treat it as an OBC reconcile.
+func (r *ObcReconciler) ensureObcWatch(ctx context.Context, req ctrl.Request) (skip bool, err error) {
+	crd := &extv1.CustomResourceDefinition{}
+	crd.Name = ObjectBucketClaimCrdName
+	if err := r.Get(ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
+		return false, err
+	}
+	utils.AssertEqual(r.ObcCrdPresentAtStart, crd.UID != "", utils.ExitCodeThatShouldRestartTheProcess)
+	if req.Namespace == "" && req.Name == ObjectBucketClaimCrdName {
```
leelavg: sry, I can't agree to this change for solving a single time pod restart. The presence of this code here signifies a deviation from single reconciled object per controller.

I can even agree to a ~20 lines controller that just restart on obccrd without touching this controller.

shirady: I will create a controller for the OBC CRD and will not change the OBC controller.

nb-ohad: A restart is not welcomed as the entire point of this pr is to prevent the restart. A new controller is also problematic as it will need to share information with the OBC controller as well and that mean global data the might or might not cross gorutine boundaries.

Having this controller react to the creation/deletion of the CRD then do nothing but change some internal state (watches) is not wrong in mind. Another controller cannot do that for this controller and even if it can do, you now split the context and responsibility of the controller into multiple pieces which creates more confusion and is, at least in my mind, considered engineering. 

@leelavg Please lay down your alternative if you have one

leelavg: > A restart is not welcomed as the entire point of this pr is to prevent the restart.
- ok, I'm mistaken then, I thought the requirement is for the user not to manually do a restart and operator will take care of restarting it
- assuming operator takes care of restart, I commented a new controller (even though it's small) to restart the pod rather than OB controller always validating the Reconcile.Request to skip or reconcile the object

As mentioned in another comment dynamic watches is only applicable to cluster scoped & namespace scoped running in single namespace but ob is namespaced & requires to watch in all namespaces.

internal/controller/obccrd_controller.go
rewantsoni, line 1: We need to restart the pod when maintenance mode crd is also introduced. This is done in operator configmap controller, if we are moving the responsibility of restart of the client process to a new controller shouldn't we also move the maintenance mode restart to this and make this controller more generic?

shirady: I added the general CRD controller.
In this PR, I didn't use it directly, but added it as TODO comment.
I'm not sure what the best way is for the team (as I'm not testing the maintenance mode and don't want to introduce a new bug in this area).

shirady:
@leelavg @nb-ohad after using the shutdown in the logs, we would see:
`oc logs -n openshift-storage-client ocs-client-operator-controller-manager-8bf588b7d-rjc98 -f`
```
2026-04-20T11:04:51Z	INFO	Stopping and waiting for non leader election runnables
2026-04-20T11:04:51Z	INFO	Stopping and waiting for leader election runnables
2026-04-20T11:04:51Z	INFO	Stopping and waiting for warmup runnables
2026-04-20T11:04:51Z	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "OperatorConfigMapReconciler"}
2026-04-20T11:04:51Z	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "storageclient", "controllerGroup": "ocs.openshift.io", "controllerKind": "StorageClient"}
2026-04-20T11:04:51Z	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "CrdsPresence", "controllerGroup": "apiextensions.k8s.io", "controllerKind": "CustomResourceDefinition"}
2026-04-20T11:04:51Z	INFO	All workers finished	{"controller": "storageclient", "controllerGroup": "ocs.openshift.io", "controllerKind": "StorageClient"}
2026-04-20T11:04:51Z	INFO	All workers finished	{"controller": "OperatorConfigMapReconciler"}
2026-04-20T11:04:51Z	INFO	All workers finished	{"controller": "CrdsPresence", "controllerGroup": "apiextensions.k8s.io", "controllerKind": "CustomResourceDefinition"}
2026-04-20T11:04:51Z	INFO	Stopping and waiting for caches
2026-04-20T11:04:51Z	INFO	Stopping and waiting for webhooks
2026-04-20T11:04:51Z	INFO	controller-runtime.webhook	Shutting down webhook server with timeout of 1 minute
2026-04-20T11:04:51Z	INFO	Stopping and waiting for HTTP servers
2026-04-20T11:04:51Z	INFO	shutting down server	{"name": "health probe", "addr": "[::]:8081"}
2026-04-20T11:04:51Z	INFO	Wait completed, proceeding to shutdown the manager
```

and the restart of the container is also reflected in:
`oc get pods -n openshift-storage-client | grep ocs-client-operator-controller-manager`
```
ocs-client-operator-controller-manager-8bf588b7d-rjc98            1/1     Running     1 (16s ago)      69s
```

internal/controller/crds_presence_controller.go
```diff
@@ -0,0 +1,61 @@
...
+		).
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+
+func (r *CrdsPresenceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	for _, name := range CrdsWatchedForPresenceRestart {
+		crd := &extv1.CustomResourceDefinition{}
+		crd.Name = name
```
leelavg:
```suggestion
	crd := &metav1.PartialObjectMetadata{}
	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
	crd.Name = name
```
as we are only interested in presence/absence.

shirady: I will change it, so now it is exactly like it was in `operatorconfigmap_controller`.
So I removed the TODO, and now it would run on the MaintenanceMode CRD as well.

rewantsoni (changes requested):
Since we no longer rely on the hack script, I think we can remove the ExitCodeThatShouldRestartTheProcess, the hack script that we have for restarting the pod, and change the entrypoint in the Docker file to the ocs-client-op binary. 
Changes introduced with https://github.com/red-hat-storage/ocs-client-operator/commit/a2a3d0a93ea7e65640a824a503cb7867547668e4

internal/controller/crds_presence_controller.go
```diff
@@ -0,0 +1,71 @@
...
+	"slices"
+
+	"github.com/go-logr/logr"
+	"github.com/red-hat-storage/ocs-client-operator/pkg/utils"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
```
rewantsoni: could you format the imports?

internal/controller/crds_presence_controller.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// CrdsWatchedForPresenceRestart lists CRDs whose install/removal must match process startup
+// (manager cache and conditional controller registration; see cmd/main.go).
+// IMPORTANT - only add the cases where the dynamic watch for the CRD did not match the case.
+var CrdsWatchedForPresenceRestart = []string{
```
leelavg: no need to be exported?

internal/controller/crds_presence_controller.go
shirady: Do you mean the order of the imports? Yes, thanks.

internal/controller/crds_presence_controller.go
shirady: No, I will change it to lowercase, thanks.

shirady:
> Since we no longer rely on the hack script, I think we can remove the ExitCodeThatShouldRestartTheProcess, the hack script that we have for restarting the pod, and change the entrypoint in the Docker file to the ocs-client-op binary. Changes introduced with [a2a3d0a](https://github.com/red-hat-storage/ocs-client-operator/commit/a2a3d0a93ea7e65640a824a503cb7867547668e4)

@rewantsoni, I suggest that those changes be done in a separate PR and not part of this fix.

leelavg (approved):

PR #558, opened 2026-04-14
controllers: remove nginx pod restart behaviour on config update

SanjalKatiyar:
https://redhat.atlassian.net/browse/DFBUGS-6311

### Testing:
_On an existing 4.22 cluster (operator already deployed), updated CSV with custom image and updated volume mounts for the console pod -_

**Client Console pod (successful nginx config reload and expected traffic logs):**

> 10.129.2.2 - - [15/Apr/2026:08:17:24 +0000] "GET /plugin-manifest.json HTTP/1.1" 200 4228 "-" "kube-probe/1.34" "-"
> 10.128.0.96 - - [15/Apr/2026:08:17:29 +0000] "GET /plugin-manifest.json HTTP/1.1" 200 4228 "https://console-openshift-console.apps.DOMAIN.devcluster.openshift.com/k8s/ns/openshift-storage/pods/ocs-client-operator-controller-manager-587d977c7d-9tgnd/logs" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36" "223.177.68.17"
> nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
> nginx: configuration file /etc/nginx/nginx.conf test is successful
> 2026/04/15 08:17:34 [notice] 216#216: signal process started
> 10.128.0.96 - - [15/Apr/2026:08:17:43 +0000] "GET /vendors-node_modules_aws-sdk_client-iam_dist-es_IAMClient_js-node_modules_aws-sdk_client-iam_-7a941c-chunk.js HTTP/1.1" 200 704083 "https://console-openshift-console.apps.DOMAIN.devcluster.openshift.com/odf/object-storage" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36" "223.177.68.17"
> 10.128.0.96 - - [15/Apr/2026:08:17:54 +0000] "GET /ocs-s3-endpoints-list-8c8886f6-fb43-4d5c-9ccf-ebd6676e3ac4/noobaaS3/?max-buckets=100&x-id=ListBuckets HTTP/1.1" 200 338 "https://console-openshift-console.apps.DOMAIN.devcluster.openshift.com/odf/object-storage/buckets" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36" "223.177.68.17, 10.128.2.8"
> 10.128.0.96 - - [15/Apr/2026:08:17:59 +0000] "GET /plugin-manifest.json HTTP/1.1" 200 4228 "https://console-openshift-console.apps.DOMAIN.devcluster.openshift.com/k8s/ns/openshift-storage/pods/ocs-client-operator-controller-manager-587d977c7d-9tgnd/logs" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/147.0.0.0 Safari/537.36" "223.177.68.17"

**Client Operator pod (nginx ConfigMap updated info and other expected logs):**

> 2026-04-15T08:11:13Z	INFO	Reconciling OperatorConfigMap	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "updated", "name": "ocs-client-operator-console-nginx-conf"}
> 2026-04-15T08:11:13Z	INFO	nginx ConfigMap updated with new config, console pod will reload nginx automatically	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "updated", "name": "ocs-client-operator-console"}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "updated", "name": "odf-client-console"}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "updated", "name": "ceph-csi-op-scc"}
> 2026-04-15T08:11:13Z	INFO	Cluster is running in DualReplica topology (TwoNodeFenced)	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "DualReplica": false}
> 2026-04-15T08:11:13Z	INFO	searching for the most compatible CSI image version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "CSI": "4.20", "Platform": "4.21.0"}
> 2026-04-15T08:11:13Z	INFO	searching for the most compatible CSI image version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "CSI": "4.17", "Platform": "4.21.0"}
> 2026-04-15T08:11:13Z	INFO	searching for the most compatible CSI image version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "CSI": "4.19", "Platform": "4.21.0"}
> 2026-04-15T08:11:13Z	INFO	searching for the most compatible CSI image version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "CSI": "4.18", "Platform": "4.21.0"}
> 2026-04-15T08:11:13Z	INFO	searching for the most compatible CSI image version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "CSI": "4.22", "Platform": "4.21.0"}
> 2026-04-15T08:11:13Z	INFO	skipping imagesets with version greater than platform version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}}
> 2026-04-15T08:11:13Z	INFO	searching for the most compatible CSI image version	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "CSI": "4.16", "Platform": "4.21.0"}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "unchanged", "name": "ceph-csi-operator-config"}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "unchanged", "name": "openshift-storage.rbd.csi.ceph.com"}
> 2026-04-15T08:11:13Z	INFO	successfully created or updated	{"controller": "OperatorConfigMapReconciler", "namespace": "openshift-storage", "name": "ocs-client-operator-config", "reconcileID": "708af3fc-6988-4b07-8a79-573a106badc6", "OperatorConfigMap": {"name":"ocs-client-operator-config","namespace":"openshift-storage"}, "operation": "unchanged", "name": "openshift-storage.cephfs.csi.ceph.com"}

**Console (Object Browser):**

Screenshot 2026-04-15 at 1 48 00 PM

### DIFF
--- a/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
@@ -7,7 +7,7 @@ metadata:
     categories: Storage
     console.openshift.io/plugins: '["odf-client-console"]'
     containerImage: quay.io/ocs-dev/ocs-client-operator:latest
-    createdAt: "2026-04-30T07:33:13Z"
+    createdAt: "2026-05-13T10:42:29Z"
     description: OpenShift Data Foundation client operator enables consumption of
       storage services from a remote centralized OpenShift Data Foundation provider
       cluster.
@@ -502,13 +502,42 @@ spec:
                 app.kubernetes.io/name: ocs-client-operator-console
             spec:
               containers:
-              - image: quay.io/ocs-dev/ocs-client-console:latest
+              - command:
+                - /bin/sh
+                - -c
+                - |
+                  (
+                    echo "Waiting for Nginx master process to start."
+                    until nginx -s reload > /dev/null 2>&1; do
+                      sleep 1
+                    done
+
+                    LAST_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                    echo "Nginx is up. Monitoring config changes."
+
+                    while true; do
+                      CURRENT_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                      if [ "$CURRENT_HASH" != "$LAST_HASH" ]; then
+                        if nginx -t && nginx -s reload; then
+                          LAST_HASH="$CURRENT_HASH"
+                        fi
+                      fi
+                      sleep 45
+                    done
+                  ) &
+
+                  if [ -x /usr/libexec/s2i/run ]; then
+                    exec /usr/libexec/s2i/run
+                  else
+                    exec nginx -g "daemon off;"
+                  fi
+                image: quay.io/ocs-dev/ocs-client-console:latest
                 livenessProbe:
                   httpGet:
                     path: /plugin-manifest.json
                     port: 9001
                     scheme: HTTPS
-                  initialDelaySeconds: 1000
+                  initialDelaySeconds: 180
                   periodSeconds: 60
                 name: ocs-client-operator-console
                 ports:
@@ -533,10 +562,6 @@ spec:
                 - mountPath: /var/serving-cert
                   name: ocs-client-operator-console-serving-cert
                   readOnly: true
-                - mountPath: /etc/nginx/nginx.conf
-                  name: ocs-client-operator-console-nginx-conf
-                  readOnly: true
-                  subPath: nginx.conf
                 - mountPath: /opt/app-root/etc/nginx.d
                   name: ocs-client-operator-console-nginx-conf
                   readOnly: true
--- a/config/console/console_init.yaml
+++ b/config/console/console_init.yaml
@@ -15,6 +15,35 @@ spec:
       containers:
         - name: ocs-client-operator-console
           image: ocs-client-operator-console:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
+              (
+                echo "Waiting for Nginx master process to start."
+                until nginx -s reload > /dev/null 2>&1; do
+                  sleep 1
+                done
+
+                LAST_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                echo "Nginx is up. Monitoring config changes."
+
+                while true; do
+                  CURRENT_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                  if [ "$CURRENT_HASH" != "$LAST_HASH" ]; then
+                    if nginx -t && nginx -s reload; then
+                      LAST_HASH="$CURRENT_HASH"
+                    fi
+                  fi
+                  sleep 45
+                done
+              ) &
+
+              if [ -x /usr/libexec/s2i/run ]; then
+                exec /usr/libexec/s2i/run
+              else
+                exec nginx -g "daemon off;"
+              fi
           resources:
             limits:
               cpu: "250m"
@@ -27,7 +56,7 @@ spec:
               path: /plugin-manifest.json
               port: 9001
               scheme: HTTPS
-            initialDelaySeconds: 1000
+            initialDelaySeconds: 180
             periodSeconds: 60
           ports:
             - containerPort: 9001
@@ -43,10 +72,6 @@ spec:
           volumeMounts:
             - name: ocs-client-operator-console-serving-cert
               mountPath: /var/serving-cert
-              readOnly: true
-            - name: ocs-client-operator-console-nginx-conf
-              mountPath: /etc/nginx/nginx.conf
-              subPath: nginx.conf
               readOnly: true
             - name: ocs-client-operator-console-nginx-conf
               mountPath: /opt/app-root/etc/nginx.d
--- a/pkg/console/console.go
+++ b/pkg/console/console.go
@@ -117,18 +117,6 @@ func GetNginxProxyConf(uniqueIdentifier, exposeAs, endpointURL, endpointHost, ce
 	return sb.String(), nil
 }
 
-func GetNginxConfConfigMap(namespace string) *apiv1.ConfigMap {
-	return &apiv1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-nginx-conf", DeploymentName),
-			Namespace: namespace,
-		},
-		Data: map[string]string{
-			"nginx.conf": nginxRootConf,
-		},
-	}
-}
-
 func getConsolePluginProxy(port int32, serviceNamespace string) []consolev1.ConsolePluginProxy {
 	return []consolev1.ConsolePluginProxy{
 		{
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -864,13 +864,7 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 	}
 
 	if nginxConfigMapResult == controllerutil.OperationResultCreated || nginxConfigMapResult == controllerutil.OperationResultUpdated {
-		c.log.Info("nginx ConfigMap data changed, restarting console pod to pick up new config")
-		// Console pod must restart (or nginx must reload) to pick up updated mounted config files.
-		// Restart failures are currently logged only. If restart fails once and config does not change again,
-		// subsequent reconciles will not retry because this block is gated by data-diff check.
-		// ToDo: If issue is prominent, explore reliable retry semantics, like restart based on marker/hash or graceful nginx reload.
-		// Note: checksum rollout via Deployment pod-template annotation is not reliable as Deployment is currently OLM/CSV-owned.
-		c.restartConsolePodsByLabel()
+		c.log.Info("nginx ConfigMap updated with new config, console pod will reload nginx automatically")
 	}
 
 	consoleService := console.GetService(c.ConsolePort, c.OperatorNamespace)
@@ -903,28 +897,6 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 	}
 
 	return nil
-}
-
-func (c *OperatorConfigMapReconciler) restartConsolePodsByLabel() {
-	podList := &corev1.PodList{}
-	if err := c.list(
-		podList,
-		client.InNamespace(c.OperatorNamespace),
-		client.MatchingLabels{console.AppNameLabelKey: console.DeploymentName},
-	); err != nil {
-		c.log.Error(err, "failed to list console pods by label selector", "namespace", c.OperatorNamespace)
-		return
-	}
-	if len(podList.Items) == 0 {
-		c.log.Info("no console pods found for restart", "namespace", c.OperatorNamespace)
-		return
-	}
-	for i := range podList.Items {
-		pod := &podList.Items[i]
-		if err := c.delete(pod); err != nil {
-			c.log.Error(err, "failed to delete console pod", "pod", pod.Name, "namespace", c.OperatorNamespace)
-		}
-	}
 }
 
 func (c *OperatorConfigMapReconciler) buildDesiredNginxDataWithProxies() (map[string]string, error) {
@@ -1014,7 +986,7 @@ func (c *OperatorConfigMapReconciler) buildS3EndpointProxyConfigForClient(unique
 		exposeAsKeys = append(exposeAsKeys, exposeAs)
 	}
 	// Processing endpoints in a fixed order so the generated nginx config "string" wouldn't change unless the actual endpoint data changed.
-	// This is to avoid unnecessary pod restarts, since restarts are triggered when changes are detected.
+	// This avoids unnecessary ConfigMap updates and nginx reloads.
 	sort.Strings(exposeAsKeys)
 	for _, exposeAs := range exposeAsKeys {
 		cfg := endpoints[exposeAs]
--- a/internal/controller/operatorconfigmap_controller_test.go
+++ b/internal/controller/operatorconfigmap_controller_test.go
@@ -297,12 +297,12 @@ func TestParseEndpointConfigs(t *testing.T) {
 
 func TestBuildS3EndpointProxyConfigForClient(t *testing.T) {
 	tests := []struct {
-		name                  string
-		secretData            map[string][]byte
-		endpoints             map[string]s3EndpointConfig
-		expectedIncludes      []string
-		expectedExcludes      []string
-		expectErr             bool
+		name             string
+		secretData       map[string][]byte
+		endpoints        map[string]s3EndpointConfig
+		expectedIncludes []string
+		expectedExcludes []string
+		expectErr        bool
 	}{
 		{
 			name: "builds config for valid https endpoints only",


bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
```diff
@@ -512,7 +512,22 @@ spec:
...
+                  while true; do
+                    CURRENT_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                    if [ "$CURRENT_HASH" != "$LAST_HASH" ]; then
+                      if nginx -t && nginx -s reload; then
+                        LAST_HASH="$CURRENT_HASH"
+                      fi
+                    fi
+                    sleep 60
+                  done &
+                  exec /usr/libexec/s2i/run
```
SanjalKatiyar: I added it directly to the pod's YAML config since these are relatively just few lines of `command`.

Alternatively,
1. I can add a new key (`nginx_reload.sh`) to the existing nginx ConfigMap and mount it into the pod at `/opt/app-root/etc/nginx.d` (same as what happens with the root/proxy nginx config today).
-- But, IMHO this may be unnecessary, current reload configuration is minimal, introducing a separate script file does not seem a need right now.
2. Or, include the script in the `odf-console` repo and add it into the pod during the Docker build.
-- But, this would distribute nginx configurations across multiple repos, which is something I have been trying to avoid.

SanjalKatiyar: `/usr/libexec/s2i/run` is the standard way of running `ubi9/nginx-122` (https://catalog.redhat.com/en/software/containers/ubi9/nginx-122/63f7653b9b0ca19f84f7e9a1).

Today we add this command in the prod Dockerfile, but now since we added `command` to the pod's YAML config, running it from here instead.

Previously and currently, nginx runs as `PID 1`, so there is no change to the existing or expected behaviour. The only addition is a separate background process to execute the reload script (already tested).

config/console/console_init.yaml
```diff
@@ -15,6 +15,21 @@ spec:
...
+              LAST_HASH=""
+              while true; do
+                CURRENT_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                if [ "$CURRENT_HASH" != "$LAST_HASH" ]; then
+                  if nginx -t && nginx -s reload; then
+                    LAST_HASH="$CURRENT_HASH"
+                  fi
+                fi
+                sleep 60
+              done &
```
SanjalKatiyar: https://github.com/red-hat-storage/ocs-client-operator/pull/558/changes#r3085215293

config/console/console_init.yaml
```diff
@@ -15,6 +15,21 @@ spec:
...
+              exec /usr/libexec/s2i/run
```
SanjalKatiyar: https://github.com/red-hat-storage/ocs-client-operator/pull/558/changes#r3085243454

bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
```diff
@@ -543,10 +558,6 @@ spec:
                 - mountPath: /var/serving-cert
                   name: ocs-client-operator-console-serving-cert
                   readOnly: true
-                - mountPath: /etc/nginx/nginx.conf
-                  name: ocs-client-operator-console-nginx-conf
-                  readOnly: true
-                  subPath: nginx.conf
```
SanjalKatiyar, line 539: This is not needed now.

Since we removed explicit pod deletion from the controller, this mount could cause bug in edge cases.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

> A container using a ConfigMap as a subPath volume will not receive ConfigMap updates.

Reading all configs (root/proxy) from `/opt/app-root/etc/nginx.d` directory mount instead now.

config/console/console_init.yaml
```diff
@@ -44,10 +59,6 @@ spec:
             - name: ocs-client-operator-console-serving-cert
               mountPath: /var/serving-cert
               readOnly: true
-            - name: ocs-client-operator-console-nginx-conf
-              mountPath: /etc/nginx/nginx.conf
-              subPath: nginx.conf
-              readOnly: true
```
SanjalKatiyar, line 50: https://github.com/red-hat-storage/ocs-client-operator/pull/558/changes#r3085277158

pkg/console/console.go
```diff
@@ -117,18 +117,6 @@ func GetNginxProxyConf(uniqueIdentifier, exposeAs, endpointURL, endpointHost, ce
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-nginx-conf", DeploymentName),
-			Namespace: namespace,
-		},
-		Data: map[string]string{
-			"nginx.conf": nginxRootConf,
-		},
-	}
-}
-
```
SanjalKatiyar, line 131: this was an unused util... so just removed it...

bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
```diff
@@ -512,7 +512,22 @@ spec:
...
+                - |
+                  LAST_HASH=""
+                  while true; do
+                    CURRENT_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                    if [ "$CURRENT_HASH" != "$LAST_HASH" ]; then
+                      if nginx -t && nginx -s reload; then
+                        LAST_HASH="$CURRENT_HASH"
+                      fi
+                    fi
+                    sleep 60
```
SanjalKatiyar: Couple of points here:

1. I am using hash instead of timestamp as suggested originally here: https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3051503282. That is because timestamp doesn't seems to be reliable:
- "reload" never creates a new master process, so timestamp will always be the same (unless pod restarts).
- while multiple worker processes are recreated during a "reload", the operation is graceful, i.e., existing processes continue running until they finish serving active traffic.
- `*.conf` are very small files, `md5sum` should not be blocking or too time consuming, I agree there is a trade-off, but it's very reliable.
Ref: https://nginx.org/en/docs/beginners_guide.html
2. I am using `sleep`, which might be a concern ? There are 3rd party tools such as `inotifywatch` that are better suited for this purpose, however, they are not included in the `ubi9/nginx` base image. Since our build process does not allow external network access or downloads, I will probably need to commit its binary to the `odf-console` repo and packaging it into the image during the build, but that is something I would prefer to avoid completely unless it is absolutely necessary or no other way possible at all.
Ref: https://stackoverflow.com/questions/41031170/nginx-reload-configuration-best-practice

config/console/console_init.yaml
```diff
@@ -15,6 +15,21 @@ spec:
...
+            - |
+              LAST_HASH=""
+              while true; do
+                CURRENT_HASH=$(md5sum /opt/app-root/etc/nginx.d/*.conf 2>/dev/null | sort | md5sum)
+                if [ "$CURRENT_HASH" != "$LAST_HASH" ]; then
+                  if nginx -t && nginx -s reload; then
+                    LAST_HASH="$CURRENT_HASH"
+                  fi
+                fi
+                sleep 60
```
SanjalKatiyar: https://github.com/red-hat-storage/ocs-client-operator/pull/558/changes#r3085504091

bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
SanjalKatiyar: I tested my changes and things works as expected, but let me know in case there are any major concerns regarding current solution and we want something else instead.

SanjalKatiyar:
Alternative solutions:

1. I did not investigate **deployment rollout via the controller** as a viable solution because the original scheduling concern (mentioned [here](https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3018659224)) could still occur in rollout scenario. Additionally, although I have not tested it extensively, I still strongly believe the behaviour I described [here](https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3039717683) can happen.

2. There was also a suggestion to use a **liveness probe for this purpose** ([here](https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3051503282)), which I was not entirely comfortable with:

- AFAIU, probes are strictly for health checks and should not execute side-effect operations/commands such as a "reload".
- A container restart is not required in this case anymore. The "reload" command is safe + graceful, it spawns new worker processes with the updated config. If the configuration is invalid, nginx automatically falls back to the previous working configuration instead of applying a faulty one (Ref: https://nginx.org/en/docs/beginners_guide.html).
That said, if we still strongly want this to be a part of liveness probe, let me know I can make changes and test again.


Also, please check: https://github.com/red-hat-storage/ocs-client-operator/pull/558#discussion_r3085504091 and the PR's description (for the testing details).

cc @nb-ohad @leelavg @Madhu-1

Madhu-1:
cant we add the configmap hash as deployment annotation and get it restarted?

SanjalKatiyar:
> cant we add the configmap hash as deployment annotation and get it restarted?

we discussed something very similar here: https://github.com/red-hat-storage/ocs-client-operator/pull/558#issuecomment-4251019953 (alternative "1").
1. https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3018659224 (main concern about pod rescheduling still exists).
2. https://github.com/red-hat-storage/ocs-client-operator/pull/524#discussion_r3039717683 (will probably have to move the console deployment from OLM/CSV managed to the OperatorConfigMap controller, though this won't resolve rescheduling concern that was raised initially).

leelavg:
overall lg, the only problem I see depending on a specific entry-point assuming the base image. you aren't using liveness due to requiring old hash value? I don't have any other ways to solve it, will wait for Ohad for one more day?

SanjalKatiyar:
> the only problem I see depending on a specific entry-point assuming the base image

@leelavg do you mean `exec /usr/libexec/s2i/run` part ?? If so, I can try adding some guards here, but that will mostly be helpful for dev cases, if image used for local testing is not `ubi9/nginx`.

> you aren't using liveness due to requiring old hash value?

My initial reasoning for not investigating much into liveness probe was this: https://github.com/red-hat-storage/ocs-client-operator/pull/558#issuecomment-4251019953 (alternative "2").
But now that you mention it, seems like persisting an older hash value is another issue as well, something I hadn't considered yet.

> will wait for Ohad for one more day

ack.

config/console/console_init.yaml
leelavg: any reason that this still needs around ~15 min, when we do sleep 1 I expect the pod to restart after a couple of tries based on how much time it usually takes for nginx to come up or else we'll be looping for 15*60 times before the pods gets restarted?

config/console/console_init.yaml
```diff
@@ -15,6 +15,36 @@ spec:
       containers:
         - name: ocs-client-operator-console
           image: ocs-client-operator-console:latest
+          command:
+            - /bin/sh
+            - -c
+            - |
+              LAST_HASH=""
```
leelavg: process is started after mounts exists and we can also take the hash here itself saving us a spurious reload?

leelavg (approved):

config/console/console_init.yaml
SanjalKatiyar: I don't think we intentionally needed an initial delay of 1000s. Likely it was added early on and remained unchanged since then. Ideally, we should have configured it to a lower value.
In normal/common use cases, it should be much faster (as you expected), although I don't have exact numbers to quantify it.

config/console/console_init.yaml
SanjalKatiyar: make sense, updated...

bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
```diff
@@ -502,7 +502,38 @@ spec:
                 app.kubernetes.io/name: ocs-client-operator-console
             spec:
               containers:
-              - image: quay.io/ocs-dev/ocs-client-console:latest
+              - command:
+                - /bin/sh
+                - -c
+                - |
+                  LAST_HASH=""
```
leelavg: can be removed now?

config/console/console_init.yaml
leelavg: ok, pls change to some meaningful value, like 60s?

SanjalKatiyar: I would prefer not to reduce it from 1000s down to 60s directly. For now, I am setting it to 3min (180s), let me know if that's still a concern ??
I know logically thinking it's still somewhat on the higher side, but as this delay only happens once after a restart, even if we end up looping 3*60 times, it should be okay IMHO.

bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
SanjalKatiyar: my bad, missed it...

leelavg (approved):
unhold after doing a final test?

PR #561, opened 2026-04-20
controllers: approve installplan when it is ready to be approved

leelavg:
olm now does a two pass verification between creating installplan and allowing it to be approved, if we approve as soon as we seen the plan them olm creates a new plan and we need to wait till olm allows the installplan to be approved.

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -125,11 +125,11 @@ type s3EndpointConfig struct {
 // OperatorConfigMapReconciler reconciles a ClusterVersion object
 type OperatorConfigMapReconciler struct {
 	client.Client
-	OperatorNamespace        string
-	ConsolePort              int32
-	Scheme                   *runtime.Scheme
-	AvailableCrds            map[string]bool
-	UpdateAlertPollInterval  func(time.Duration)
+	OperatorNamespace       string
+	ConsolePort             int32
+	Scheme                  *runtime.Scheme
+	AvailableCrds           map[string]bool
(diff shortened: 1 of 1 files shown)


internal/controller/operatorconfigmap_controller.go
```diff
@@ -559,8 +559,10 @@ func (c *OperatorConfigMapReconciler) reconcileInstallPlans() error {
 	}
 	for idx := range installPlans.Items {
 		installPlan := &installPlans.Items[idx]
-		if !installPlan.Spec.Approved {
-			if err := c.Patch(c.ctx, installPlan, approvePatch); err != nil {
+		if installPlan.Spec.Approval == opv1a1.ApprovalManual && !installPlan.Spec.Approved {
+			if installPlan.Status.Phase != opv1a1.InstallPlanPhaseRequiresApproval {
+				return fmt.Errorf("installplan is not ready to be approved")
```
rewantsoni: would it benefit to also log the name of the install plan that is not ready to be approved?

leelavg:
[without.txt](https://github.com/user-attachments/files/26882963/without.txt)
- w/o fix we can see 4 installplans for 2 versions out of which there are 2 incomplete which client-op approving the last one, even that resolution failed and deployed incomplete resources

[with.txt](https://github.com/user-attachments/files/26882958/with.txt)
- w/ fix there are only 2 installplans for 2 versions and client-op waited properly before approving them

internal/controller/operatorconfigmap_controller.go
leelavg: done.

PR #580, opened 2026-04-27
Fix - Add Watch for OBC in Storageclient Controller In Case Status is Manually Deleted

shirady:
### Describe the problem

### Explain the changes
1. In main - add the map of `AvailableCrds` to `StorageClientReconciler`.
2. Add a watch on OBC in case (AND): (1) update event, (2) has storage client label value name, (3) status change
3. Added utility function `LabelExistsPredicate`
4. Added AI-generated tests for the utility function `LabelExistsPredicate`.

#### Unit tests:
Please run: `go test ./pkg/utils/ -v -run TestLabelExistsPredicate`

#### Manual Test:
Currently, we see that the storage client controller is triggered every minute, so we cannot reproduce the situation where the OBC status was deleted (because now it is always getting updated back, so effectively we don't see the new watch changes). So now I just verify that the pod does not panic when there is an OBC CRD installed and when there is not.

### DIFF
--- a/internal/controller/storageclient_controller.go
+++ b/internal/controller/storageclient_controller.go
@@ -67,6 +67,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -137,6 +138,7 @@ type StorageClientReconciler struct {
 	Scheme            *runtime.Scheme
 	OperatorNamespace string
 	OperatorPodName   string
+	AvailableCrds     map[string]bool
 
 	cache            cache.Cache
 	controller       controller.Controller
@@ -212,7 +214,33 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			return requests
 		},
 	)
-	controller, err := ctrl.NewControllerManagedBy(mgr).
+	obcStatusChangedPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldOBC, oldOk := e.ObjectOld.(*nbv1.ObjectBucketClaim)
+			newOBC, newOk := e.ObjectNew.(*nbv1.ObjectBucketClaim)
+			if !oldOk || !newOk || oldOBC == nil || newOBC == nil {
+				return false
+			}
+			return !reflect.DeepEqual(oldOBC.Status, newOBC.Status)
+		},
+	}
+	enqueueStorageClientRequestFromOBC := handler.EnqueueRequestsFromMapFunc(
+		func(_ context.Context, obj client.Object) []ctrl.Request {
+			if obj == nil {
+				return nil
+			}
+			labels := obj.GetLabels()
+			if labels == nil {
+				return nil
+			}
+			storageClientName := labels[storageClientNameLabel]
+			if storageClientName == "" {
+				return nil
+			}
+			return []ctrl.Request{{NamespacedName: types.NamespacedName{Name: storageClientName}}}
+		},
+	)
+	bldr := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.StorageClient{}).
 		Owns(&batchv1.CronJob{}).
 		Owns(&quotav1.ClusterResourceQuota{}, builder.WithPredicates(generationChangePredicate)).
@@ -235,8 +263,21 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				utils.EventTypePredicate(true, false, false, false),
 			),
 			builder.OnlyMetadata,
-		).
-		Build(r)
+		)
+	if r.AvailableCrds[ObjectBucketClaimCrdName] {
+		bldr = bldr.Watches(
+			&nbv1.ObjectBucketClaim{},
+			enqueueStorageClientRequestFromOBC,
+			builder.WithPredicates(
+				predicate.And(
+					utils.EventTypePredicate(false, true, false, false),
+					utils.LabelExistsPredicate(storageClientNameLabel),
+					obcStatusChangedPredicate,
+				),
+			),
+		)
+	}
+	controller, err := bldr.Build(r)
 
 	r.controller = controller
 	r.cache = mgr.GetCache()
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -285,6 +285,7 @@ func main() {
 		Scheme:            mgr.GetScheme(),
 		OperatorNamespace: utils.GetOperatorNamespace(),
 		OperatorPodName:   podName,
+		AvailableCrds:     availCrds,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "StorageClient")
 		os.Exit(1)
--- a/pkg/utils/predicates.go
+++ b/pkg/utils/predicates.go
@@ -34,3 +34,14 @@ func EventTypePredicate(create, update, del, generic bool) predicate.Predicate {
 		},
 	}
 }
+
+// LabelExistsPredicate return a predicate the filter events based on the existence of a label
+func LabelExistsPredicate(labelKey string) predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		if obj == nil || obj.GetLabels() == nil {
+			return false
+		}
+		_, found := obj.GetLabels()[labelKey]
+		return found
+	})
+}
--- a/pkg/utils/predicates_test.go
+++ b/pkg/utils/predicates_test.go
@@ -1,0 +1,202 @@
+package utils
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+func TestLabelExistsPredicate(t *testing.T) {
+	const labelKey = "ocs.openshift.io/storageclient.name"
+
+	tests := []struct {
+		name     string
+		obj      *corev1.ConfigMap
+		expected bool // true if any event path should allow; we assert Create matches filter(obj)
+	}{
+		{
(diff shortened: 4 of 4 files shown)


shirady:
@rewantsoni , @leelavg and @nb-ohad , please take a look

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,32 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	)
-	controller, err := ctrl.NewControllerManagedBy(mgr).
+	obcStatusPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldOBC := e.ObjectOld.(*nbv1.ObjectBucketClaim)
+			newOBC := e.ObjectNew.(*nbv1.ObjectBucketClaim)
+			if newOBC.GetLabels() == nil || newOBC.GetLabels()[storageClientNameLabel] == "" {
```
leelavg:
```suggestion
			if len(newOBC.GetLabels()) == 0 || newOBC.GetLabels()[storageClientNameLabel] == "" {
```

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,32 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				return false
+			}
+			return !reflect.DeepEqual(oldOBC.Status, newOBC.Status)
```
leelavg, line 224: is there no timestamp related field in status?

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,32 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		},
+	}
+	enqueueStorageClientRequestFromOBC := handler.EnqueueRequestsFromMapFunc(
+		func(_ context.Context, obj client.Object) []ctrl.Request {
+			if name := obj.GetLabels()[storageClientNameLabel]; name != "" {
+				return []ctrl.Request{{
+					NamespacedName: client.ObjectKeyFromObject(&v1alpha1.StorageClient{
+						ObjectMeta: metav1.ObjectMeta{Name: name},
+					}),
+				}}
```
leelavg:
```suggestion
					return []ctrl.Request{{NamespacedName: types.NamespacedName{Name: name}}}
```
if you have obj then can get key from it and don't need to get the key by specifically constructing the obj which only has name in it.

leelavg:
pls update the pr summary, ie, can remove "in case it is manually deleted", isn't this pr to register a watch if there is obc and the restart on obc deletion is a different pr?

rewantsoni (changes requested):
Instead of sending the AvailableCRD, can we not do dynamicWatches for OBC? something similar to what we did for OB?

shirady:
> pls update the pr summary, ie, can remove "in case it is manually deleted", isn't this pr to register a watch if there is obc and the restart on obc deletion is a different pr?

@leelavg this PR is for a minor bug where someone deleted the patched status of the OBC.
We can reproduce it as the storage client reconcile every minute.
You are correct that the restart of the pod is done in another PR.
I wanted to add a watch for the OBC status changes only.

shirady:
> Instead of sending the AvailableCRD, can we not do dynamicWatches for OBC? something similar to what we did for OB?

@rewantsoni, I'm not sure I understand.
The OB had dynamic watch (starts when the CRD is detected).
In the OBC when the CRD is detected there would be a pod resatrt, and I wanted to add a watch for the OBC status.

internal/controller/storageclient_controller.go
shirady: I will change it, thanks.

internal/controller/storageclient_controller.go
shirady: How can I check that?
In the code the status struct `ObjectBucketClaimStatus` contains only the phase string  `ObjectBucketClaimStatusPhase`- 
https://github.com/kube-object-storage/lib-bucket-provisioner/blob/5164e374648929412a95b1072274747a7efb13a1/pkg/apis/objectbucket.io/v1alpha1/objectbucketclaim_types.go#L78-L81

internal/controller/storageclient_controller.go
shirady: I will change it, thanks.

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,28 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		},
 	)
-	controller, err := ctrl.NewControllerManagedBy(mgr).
+	obcStatusPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldOBC := e.ObjectOld.(*nbv1.ObjectBucketClaim)
+			newOBC := e.ObjectNew.(*nbv1.ObjectBucketClaim)
```
nb-ohad: 1. There is no point to check for nils if you are going to cast. The cast itself can do the check for you.
2. You should not be casting without checking for casting errors.


```suggestion
			oldOBC, oldOk  = e.ObjectOld.(*nbv1.ObjectBucketClaim) 
			newOBC, newOk := e.ObjectNew.(*nbv1.ObjectBucketClaim)
			if !oldOK || !newOK {
				return false
			}
```

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,28 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			if len(newOBC.GetLabels()) == 0 || newOBC.GetLabels()[storageClientNameLabel] == "" {
+				return false
+			}
```
nb-ohad: I am not sure what you are trying to do here, but regardless if we need to have a predicate for labels we should have a predicate for labels. This predicate is about the status

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
shirady: I copied the pattern from here - 
https://github.com/red-hat-storage/ocs-client-operator/blob/dc864f1a5448dfd5ddb1e0116a07fb8fb2664416/internal/controller/operatorconfigmap_controller.go#L234-L243

I will change it, thanks.

@nb-ohad, what about adding a nil check for readability?
```
			if !oldOk || !newOk || oldOBC == nil || newOBC == nil {
```

internal/controller/storageclient_controller.go
shirady: I'm trying to filter in the OBCs where their statue was changed, AND their value of the label of storage client is matching the storage client.

internal/controller/storageclient_controller.go
nb-ohad: You are still missing the cast checks, and by adding them you will make the nil check irrelevant. Please see my suggestion from above

internal/controller/storageclient_controller.go
nb-ohad: So you are coupling 2 concerns into a single predicate. It will be better if you just add a label match predicate instead

internal/controller/storageclient_controller.go
nb-ohad: Apparently there is an edge case here where controller-runtime will pass in a nil pointer of the correct pointer type. in that case the cast will not fail and we do need the nil checks.

shirady: I will add this, thanks.

internal/controller/storageclient_controller.go
shirady: I will move to 2 predicate with and between them.
I got AI suggestion to have a stricter predicate of the label (not just existence), and I will change it.

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,37 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			oldOBC, oldOk := e.ObjectOld.(*nbv1.ObjectBucketClaim)
+			newOBC, newOk := e.ObjectNew.(*nbv1.ObjectBucketClaim)
+			if !oldOk || !newOk || oldOBC == nil || newOBC == nil {
+				return false
+			}
+			oldName := oldOBC.GetLabels()[storageClientNameLabel]
+			newName := newOBC.GetLabels()[storageClientNameLabel]
+			return oldName != "" && newName != "" && oldName == newName
+		},
+	}
```
nb-ohad: 1. This should not be an update predicate, but need to be applied to all events. You are searching to filter in any obc that have a storage client label.
2. If you are only accessing metadata you have no reason to cast into concrete types
3. This is a generic predicate for filtering by label, not specific for OBC. This will be needed in other places so the correct thing to do is to add this into our predicate utils (where we have the name predicate)





```go
func LabelExistsPredicate(labelKey string) predicate.Predicate {
    	return predicate.NewPredicateFuncs(func (obj client.Object) bool {
    		if obj == nil || obj.GetLabels() == nil {
				return false
    		}
			_, found := obj.GetLables()[key];
    		return found
    	})
}
```

internal/controller/storageclient_controller.go
```diff
@@ -212,7 +214,37 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return !reflect.DeepEqual(oldOBC.Status, newOBC.Status)
+		},
+	}
+	enqueueStorageClientRequestFromOBC := handler.EnqueueRequestsFromMapFunc(
+		func(_ context.Context, obj client.Object) []ctrl.Request {
+			if name := obj.GetLabels()[storageClientNameLabel]; name != "" {
+				return []ctrl.Request{{NamespacedName: types.NamespacedName{Name: name}}}
+			}
+			return nil
+		},
```
nb-ohad, line 241: Both obj and labels might be nil

```suggestion
		func(_ context.Context, obj client.Object) []ctrl.Request {
			if obj != nil && obj.GetLabels != nil {
				return nil
			}
		    
		    storageCleintName := obj.GetLabels()[storageClientNameLabel]
		    if storageCleintName == "" {
		    	return nil
		    }

			return []ctrl.Request{{ 
				NamespacedName: types.NamespacedName{Name: storageCleintName},
			}}
		},
```

nb-ohad (changes requested):

internal/controller/storageclient_controller.go
shirady: I will add this to the file `pkg/utils/predicates.go`, thanks.

internal/controller/storageclient_controller.go
shirady: I'm not sure about the suggestion, but I will add the check of obj and the labels as returned nil in this function.

PR #586, opened 2026-04-30
Remove Exit Code That Should Restart Process

shirady:
### Describe the problem
Tech debt that was mentioned in the comment https://github.com/red-hat-storage/ocs-client-operator/pull/556#pullrequestreview-4187935903:

> Since we no longer rely on the hack script, I think we can remove the ExitCodeThatShouldRestartTheProcess, the hack script that we have for restarting the pod, and change the entrypoint in the Docker file to the ocs-client-op binary.
> Changes introduced with https://github.com/red-hat-storage/ocs-client-operator/commit/a2a3d0a93ea7e65640a824a503cb7867547668e4

Note: When I reverted the changes, I didn't replace `entrypoint` with `manager` as it was in the commit but with `ocs-client-operator` due to the changes introduced in https://github.com/red-hat-storage/ocs-client-operator/commit/ea9b9e0cafeeac2f4698bfd6241812389d9220b6

### Explain the changes
1. Remove `ExitCodeThatShouldRestartTheProcess` as it is no longer being used.
2. Remove the hack script `hack/entrypoint.sh`.
3. Edit the Dockerfile to use the `ocs-client-operator` binary.
4. Edit the manger yaml with `ocs-client-operator` instead of `entrypoint`.
5. Run `make bundle`.
6. Remove the watch Maintenance Mode CRD in configmap controller and the kube-builder annotation related to the CRD permission in the controller (I ran `make manifests`, but it didn't result in any code change).

#### Manual Test:
I built the image and pushed it to quay, and edited the CSV it created an error `CreateContainerError` in the pod `ocs-client-operator-controller-manager`, partial output from `oc describe pod -n openshift-storage-client ocs-client-operator-controller-manager-<chars>`:
```
Warning  Failed          6s (x8 over 83s)  kubelet            Error: container create failed: executable file `/entrypoint` not found: No such file or directory
```
I got an AI-generated command - 
```
oc patch clusterserviceversion ocs-client-operator.v4.22.0-42.stable  -n openshift-storage-client --type='json' -p='[
  {"op": "replace", "path": "/spec/install/spec/deployments/0/spec/template/spec/containers/0/command/0", "value": "/ocs-client-operator"}
]'
```
Then I deleted the pod, and it was in status `Running` as expected.


shirady:
@rewantsoni @leelavg , please take a look

rewantsoni (commented):
I think we can remove the maintenance mode crd watch from the operator configmap controller as well.

shirady:
> I think we can remove the maintenance mode crd watch from the operator configmap controller as well.

I will remove, thanks.

leelavg (approved):

PR #597, opened 2026-06-01
bundle: remove dependencies

iamniting:
olm v1 does not support dependencies. To install the pkg with olm v1 we are required to remove the dependencies.

Signed-off-by: Nitin Goyal <nigoyal@redhat.com>


leelavg (approved):
for posterity, the merge is to a diff branch.

PR #599, opened 2026-06-03
controllers: remove auto-upgrade of client-op subs

rewantsoni:
stop reconciling client-op subs and update the deps only when client op is also upgraded.


Madhu-1 (approved):

leelavg (approved):
is an alert planned for surfacing the upgrade need?

PR #601, opened 2026-06-04
simplify servers managed by controller-runtime and make them tls ready

leelavg:
make metrics exporter server over https and update the ports along with proper naming for existing mounts, remove all the redundant or unused code while starting metrics.


bundle/manifests/ocs-client-operator.clusterserviceversion.yaml
```diff
@@ -632,20 +636,12 @@ spec:
...
-          - operators.coreos.com
+          - ""
           resources:
-          - clusterserviceversions
+          - configmaps
           verbs:
           - get
           - list
-          - watch
-        - apiGroups:
```
Madhu-1, line 642: what is the reason for this change?

leelavg: they are redundant rbac, will be confirming during testing as well.

PR #602, opened 2026-06-04
controllers: restart container on TLSProfile change for metrics & webhook server

leelavg:
the flow:
on start take the baseline tls profile config and configure the servers
when the profile gets updated restart the container to pick the new config


bundle/manifests/ocs-client-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
```diff
@@ -4,8 +4,13 @@ metadata:
...
 spec:
   endpoints:
-  - path: /metrics
-    port: metrics
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
+    port: https
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
```
Madhu-1, line 12: is this need to be set to true?

cmd/main.go
```diff
@@ -268,6 +172,55 @@ func main() {
...
 	}
 
+	var initialTLSGeneration int64
+	startupProfile := &ocstlsv1.TLSProfile{}
+	startupProfile.Name = controller.TLSProfileName
+	startupProfile.Namespace = utils.GetOperatorNamespace()
+	if err := apiClient.Get(apiCtx, client.ObjectKeyFromObject(startupProfile), startupProfile); err != nil {
+		if !kerrors.IsNotFound(err) {
+			setupLog.Error(err, "failed to get TLSProfile at startup")
+			os.Exit(1)
```
Madhu-1, line 182: can we have a helper to log and exit to avoid duplicate lines of code?

cmd/main.go
```diff
@@ -111,100 +112,20 @@ func init() {
...
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	// if the enable-http2 flag is false (the default), http/2 should be disabled
-	// due to its vulnerabilities. More specifically, disabling http/2 will
-	// prevent from being vulnerable to the HTTP/2 Stream Cancellation and
-	// Rapid Reset CVEs. For more information see:
-	// - https://github.com/advisories/GHSA-qppj-fm5r-hxr3
-	// - https://github.com/advisories/GHSA-4374-p667-p6c8
-	disableHTTP2 := func(c *tls.Config) {
```
Madhu-1, line 153: what about disableHTTP2? are we going to support http2?

cmd/main.go
```diff
@@ -111,100 +112,20 @@ func init() {
...
-	var secureMetrics bool
-	var consolePort int
-	var webhookPort int
-	var tlsOpts []func(*tls.Config)
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
-		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
-		"Enable leader election for controller manager. "+
-			"Enabling this will ensure there is only one active controller manager.")
```
Madhu-1, line 128: no leaderelection as well? to ensure that only one isntance of operator runs?

bundle/manifests/ocs-client-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
leelavg: reasoning already mentioned in config/prometheus/monitor.yaml

cmd/main.go
leelavg: we have been explicitly doing this directly on main func so far as not to exit from any other utils/funcs, can refactor in another small PR.

cmd/main.go
leelavg: yep, no one touched this after scaffolding, the links mentioned are long fixed in the libs.

cmd/main.go
leelavg: yep, we never ran client-op with replica=2. dead code, wanted to remove unused code whenever we are touching surrounding code.

Madhu-1: There are chances that two client operator pods might run for short duration of time. as we are not running leader election by default, we are good, can remove this dead code

leelavg: good point, however we aren't using leaderelection at all and so is the removal. During restarts, readiness just act as a hand over, pod restarts is the only problematic thing, unless we hit both upgrade & restart we are fine, moreover the behavior isn't changed by removing the code.

PR #609, opened 2026-06-15
hack: add container to CONTAINER_TOOL_CMD

rewantsoni:
Update the hack file to add [container](https://github.com/apple/container) build command to CONTAINER_TOOL_CMD so that we can. build using that on apple silicon.

### DIFF
--- a/hack/make-project-vars.mk
+++ b/hack/make-project-vars.mk
@@ -22,3 +22,7 @@ endif
 ifeq ($(IMAGE_BUILD_CMD),)
 IMAGE_BUILD_CMD := $(shell command -v podman || echo "")
 endif
+
+ifeq ($(IMAGE_BUILD_CMD),)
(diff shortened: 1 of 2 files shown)


hack/make-project-vars.mk
leelavg, line 1: I believe all the three programs have compatible cli args for push, build and run, I don't see a reason to explicitly define three new make vars for this.

rewantsoni: The container tool doesn't have container pull/push directly, the command is container image pull/push. I will try to check if docker/podamn has podman container push/pull and update it accordingly

rewantsoni: The problem was with only push/pull command the docker also has an docker image push/pull command which we can use. We don't require three different make vars

leelavg (approved):

PR #640, opened 2026-06-29
controllers: Update pods to have annotations with openshift.io/required-scc

ezio-auditore:
Epic: [RHSTOR-8757](https://redhat.atlassian.net/browse/RHSTOR-8757) [Enforce least-privileged Security Context Constraint (SCC) across ODF pods ](https://redhat.atlassian.net/browse/RHSTOR-8757)

Pins every Pod managed by ocs-client-operator to an explicit Security Context Constraint using the `openshift.io/required-scc` annotation, replacing reliance on OpenShift SCC admission to dynamically select the most permissive matching SCC.

- Add shared SCC annotation helpers in `[pkg/templates/csi.go](pkg/templates/csi.go)`: CSI driver pods use `ceph-csi-op-scc`; standard workloads use `restricted-v2`
- Annotate CSI controller and node plugin defaults in `ceph-csi-operator-config` (`OperatorConfig`) so delegated RBD/CephFS/NFS plugin Pods are bound to the existing custom SCC
- Annotate static OLM bundle Deployments: operator manager (`[config/manager/manager.yaml](config/manager/manager.yaml)`) and console plugin (`[config/console/console_init.yaml](config/console/console_init.yaml)`) with `restricted-v2`
- Annotate the per-`StorageClient` status-reporter CronJob pod template in `[internal/controller/storageclient_controller.go](internal/controller/storageclient_controller.go)` with `restricted-v2`

This improves auditability and least-privilege enforcement for regulated deployments by making SCC assignment deterministic and reproducible.

## SCC mapping

| Workload | SCC |
|----------|-----|
| Operator controller-manager | `restricted-v2` |
| ODF client console | `restricted-v2` |
| StorageClient status-reporter CronJob | `restricted-v2` |
| CSI controller + node plugins (via ceph-csi-operator) | `ceph-csi-op-scc` |

## Notes

- The custom CSI SCC (`ceph-csi-op-scc`) and its service-account bindings are already reconciled by this operator; this change only adds the required-SCC pin on the pod templates.
- CSI CNI network annotations merged at reconcile time continue to be additive on top of the SCC annotation.
- Updating pod-template annotations will cause affected Pods to be recreated on rollout.

## Test plan

- [ ] Deploy on an OpenShift cluster with at least one `StorageClient`
- [ ] Verify operator manager pod template and running pod:
  

```bash
  oc -n openshift-storage get deploy ocs-client-operator-controller-manager \
    -o jsonpath='{.spec.template.metadata.annotations.openshift\.io/required-scc}{"\n"}'
  oc -n openshift-storage get pods -l app=ocs-client-operator \
    -o custom-columns=NAME:.metadata.name,SCC:.metadata.annotations.openshift\.io/scc

### DIFF
--- a/pkg/templates/csi.go
+++ b/pkg/templates/csi.go
@@ -49,7 +49,19 @@ func InjectSnapshotMetadataTLSVolume(cp *csiopv1.ControllerPluginSpec) {
 }
 
 // security context constraints
-const SCCName = "ceph-csi-op-scc"
+const (
+	SCCName           = "ceph-csi-op-scc"
+	restrictedSCCName = "restricted-v2"
+)
+
+var (
+	csiSCCPodAnnotations = map[string]string{
+		secv1.RequiredSCCAnnotation: SCCName,
+	}
+	RestrictedSCCPodAnnotations = map[string]string{
+		secv1.RequiredSCCAnnotation: restrictedSCCName,
+	}
+)
 
 // TODO: could pull directly from ceph-csi-operator when available
 var securityContextConstraints = secv1.SecurityContextConstraints{
@@ -212,6 +224,7 @@ var CSIOperatorConfigSpec = csiopv1.OperatorConfigSpec{
 				},
 			},
 			PodCommonSpec: csiopv1.PodCommonSpec{
+				Annotations:        csiSCCPodAnnotations,
 				PrioritylClassName: ptr.To("system-cluster-critical"),
 				ImagePullPolicy:    corev1.PullIfNotPresent,
 				Tolerations: []corev1.Toleration{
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: pkg/templates/csi.go [hunk 1]
```diff
@@ -16,7 +16,19 @@ const CephFsDriverName = "openshift-storage.cephfs.csi.ceph.com"
 const NfsDriverName = "openshift-storage.nfs.csi.ceph.com"
 
 // security context constraints
-const SCCName = "ceph-csi-op-scc"
+const (
+	SCCName           = "ceph-csi-op-scc"
+	RestrictedSCCName = "restricted-v2"
```
leelavg: do a quick check and unexport these?

ezio-auditore: can unexport `RestrictedSCCName` but SCCName is being used in two places at
`internal/controller/operatorconfigmap_controller.go`
- deleteDelegatedCSI
- reconcileDelegatedCSI

pkg/templates/csi.go
```diff
@@ -49,7 +49,19 @@ func InjectSnapshotMetadataTLSVolume(cp *csiopv1.ControllerPluginSpec) {
...
 
 // security context constraints
-const SCCName = "ceph-csi-op-scc"
+const (
+	SCCName           = "ceph-csi-op-scc"
+	restrictedSCCName = "restricted-v2"
+)
+
+var (
+	CSISCCPodAnnotations = map[string]string{
```
Madhu-1: This doesnt need to be exported?

ezio-auditore: No, this too doesn't I will fix it

leelavg (approved):

PR #646, opened 2026-06-30
csi: adjust CSI pods CPU requests on IBM Z

malayparida2000:
Halve default CSI controller and node plugin CPU requests when the operator runs on s390x, matching the IBM Z resource adjustment in ocs-operator. The reason for this is individual CPU cores on IBM Z are more than 2X powerful than the ones on other archs.
Ref-https://redhat.atlassian.net/browse/RHSTOR-9444

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -22,6 +22,7 @@ import (
 	"maps"
 	"net/url"
 	"reflect"
+	goruntime "runtime"
 	"slices"
 	"sort"
 	"strconv"
@@ -114,6 +115,9 @@ const (
 
 	pvDriverIndexName  = "index:persistentVolumeDriver"
 	vscDriverIndexName = "index:volumeSnapshotContentDriver"
+
+	ibmZCpuArch         = "s390x"
+	ibmZCpuAdjustFactor = 0.5
 )
 
 // ConfigMapData value from the provider that contains the s3 endpoint info (key is the unique identifier, using which the endpoint is exposed).
@@ -633,6 +637,10 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 		}
 		templates.CSIOperatorConfigSpec.DeepCopyInto(&csiOperatorConfig.Spec)
 		driverSpecDefaults := csiOperatorConfig.Spec.DriverSpecDefaults
+		if goruntime.GOARCH == ibmZCpuArch {
+			adjustPluginCpuResourcesForIbmZ(&driverSpecDefaults.ControllerPlugin.Resources, ibmZCpuAdjustFactor)
+			adjustPluginCpuResourcesForIbmZ(&driverSpecDefaults.NodePlugin.Resources, ibmZCpuAdjustFactor)
+		}
 		if isTnfCluster {
 			templates.CSIOperatorTNFControllerPluginResourceSpec.DeepCopyInto(&driverSpecDefaults.ControllerPlugin.Resources)
 			templates.CSIOperatorTNFNodePluginResourceSpec.DeepCopyInto(&driverSpecDefaults.NodePlugin.Resources)
@@ -1416,4 +1424,34 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
 		return fmt.Errorf("failed to reconcile snapshot metadata spec ConfigMap: %w", err)
 	}
 	return nil
+}
+
+func adjustPluginCpuResourcesForIbmZ(resources any, adjustFactor float64) {
+	if resources == nil {
+		return
+	}
+	for _, req := range resourceRequirementsFromStruct(resources) {
+		adjustCpuResourcesForIbmZ(req, adjustFactor)
+	}
+}
+
+func resourceRequirementsFromStruct(spec any) []*corev1.ResourceRequirements {
+	v := reflect.ValueOf(spec).Elem()
+	reqs := make([]*corev1.ResourceRequirements, 0, v.NumField())
+	for i := 0; i < v.NumField(); i++ {
+		req, ok := v.Field(i).Interface().(*corev1.ResourceRequirements)
+		if ok && req != nil {
+			reqs = append(reqs, req)
+		}
+	}
+	return reqs
+}
+
+func adjustCpuResourcesForIbmZ(req *corev1.ResourceRequirements, adjustFactor float64) {
+	if req == nil {
+		return
+	}
+	if cpuQty, exists := req.Requests[corev1.ResourceCPU]; exists {
+		req.Requests[corev1.ResourceCPU] = utils.AdjustCPU(cpuQty, adjustFactor)
+	}
 }
--- a/pkg/utils/k8sutils.go
+++ b/pkg/utils/k8sutils.go
@@ -28,6 +28,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/red-hat-storage/ocs-operator/services/provider/api/v4/interfaces"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -217,4 +218,8 @@ func SetClusterInformation(
 	status.SetClusterName(clusterDNS.Spec.BaseDomain)
 
 	return nil
+}
+
+func AdjustCPU(cpuQty resource.Quantity, adjustFactor float64) resource.Quantity {
+	return *resource.NewMilliQuantity(int64(float64(cpuQty.MilliValue())*adjustFactor), resource.DecimalSI)
 }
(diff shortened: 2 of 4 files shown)


internal/controller/operatorconfigmap_controller.go
```diff
@@ -633,6 +637,10 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 		}
 		templates.CSIOperatorConfigSpec.DeepCopyInto(&csiOperatorConfig.Spec)
 		driverSpecDefaults := csiOperatorConfig.Spec.DriverSpecDefaults
+		if goruntime.GOARCH == ibmZCpuArch {
```
leelavg, line 640: why not goruntime.Is390x == 1 rather than using names?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -633,6 +637,10 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
...
+			adjustControllerPluginCpuResourcesForIbmZ(&driverSpecDefaults.ControllerPlugin.Resources, ibmZCpuAdjustFactor)
```
leelavg:
```suggestion
			adjustControllerPluginCpuRequestsForIbmZ(&driverSpecDefaults.ControllerPlugin.Resources, ibmZCpuAdjustFactor)
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -633,6 +637,10 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
...
+			adjustNodePluginCpuResourcesForIbmZ(&driverSpecDefaults.NodePlugin.Resources, ibmZCpuAdjustFactor)
```
leelavg:
```suggestion
			adjustNodePluginCpuRequestsForIbmZ(&driverSpecDefaults.NodePlugin.Resources, ibmZCpuAdjustFactor)
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1417,3 +1425,32 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
...
+	adjustCpuResourcesForIbmZ(resources.Plugin, adjustFactor)
+}
+
+func adjustNodePluginCpuResourcesForIbmZ(resources *csiopv1.NodePluginResourcesSpec, adjustFactor float64) {
+	adjustCpuResourcesForIbmZ(resources.Registrar, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Liveness, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Addons, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.LogRotator, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Plugin, adjustFactor)
+}
```
leelavg: if the adjust factor is same why not go over all the resources and set the values rather than picking each resource/sidecar because it creates a requirement to update things in two places when there is a new container. tnf also should've been like that.

pkg/utils/k8sutils.go
```diff
@@ -218,3 +220,8 @@ func SetClusterInformation(
 
 	return nil
 }
+
+func AdjustCPU(cpuQty resource.Quantity, adjustFactor float64) resource.Quantity {
+	str := strconv.FormatInt(int64(float64(cpuQty.MilliValue())*adjustFactor), 10) + "m"
```
leelavg: too many conversions i guess, took a quick look at the lib, could you try `resource.NewMilliQuantity(cpuQty.MilliValue() / 2, resource.DecimalSI)`?

leelavg (commented):
@Madhu-1 fyi.

internal/controller/operatorconfigmap_controller.go
malayparida2000: I could not find any variable or func of the type goruntime.Is390x / goruntime.IsS390x
Only GOARCH & GOOS are availble https://pkg.go.dev/runtime#pkg-constants

internal/controller/operatorconfigmap_controller.go
malayparida2000: I intentionally did not mention Request in the name, so that in future if required the same func can be expanded to adjust limits as well

internal/controller/operatorconfigmap_controller.go
malayparida2000: I intentionally did not mention Request in the name, so that in future if required the same func can be expanded to adjust limits as well

internal/controller/operatorconfigmap_controller.go
malayparida2000: You are right, going over the list would be a better approach so that we cover any new containers if they come up in future

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1417,3 +1425,32 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
...
 	return nil
 }
+
+func adjustControllerPluginCpuResourcesForIbmZ(resources *csiopv1.ControllerPluginResourcesSpec, adjustFactor float64) {
+	adjustCpuResourcesForIbmZ(resources.Attacher, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Snapshotter, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Resizer, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Provisioner, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.OMapGenerator, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Liveness, adjustFactor)
```
Madhu-1: we dont have this anymore should be removed

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1417,3 +1425,32 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
...
+	adjustCpuResourcesForIbmZ(resources.Addons, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.LogRotator, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Plugin, adjustFactor)
+}
+
+func adjustNodePluginCpuResourcesForIbmZ(resources *csiopv1.NodePluginResourcesSpec, adjustFactor float64) {
+	adjustCpuResourcesForIbmZ(resources.Registrar, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Liveness, adjustFactor)
```
Madhu-1: same here as well

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1417,3 +1425,32 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
 	}
 	return nil
 }
+
+func adjustControllerPluginCpuResourcesForIbmZ(resources *csiopv1.ControllerPluginResourcesSpec, adjustFactor float64) {
+	adjustCpuResourcesForIbmZ(resources.Attacher, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Snapshotter, adjustFactor)
+	adjustCpuResourcesForIbmZ(resources.Resizer, adjustFactor)
```
Madhu-1: i think there is two more containers missing or two more related to odf snapshotter and snapshotter metadata

malayparida2000: https://github.com/ceph/ceph-csi-operator/blob/bc21847a37a77e7e8b50ddce395700b3f8347e4e/api/v1/driver_types.go#L192-L219

I covered refering to this

internal/controller/operatorconfigmap_controller.go
malayparida2000: I got the list from here https://github.com/ceph/ceph-csi-operator/blob/bc21847a37a77e7e8b50ddce395700b3f8347e4e/api/v1/driver_types.go#L135-L150

internal/controller/operatorconfigmap_controller.go
malayparida2000: Got from here
https://github.com/ceph/ceph-csi-operator/blob/bc21847a37a77e7e8b50ddce395700b3f8347e4e/api/v1/driver_types.go#L199-L207

Madhu-1: its not used in downstream we can remove it

internal/controller/operatorconfigmap_controller.go
Madhu-1: odf-snapshotter uses the internal snapshotter and the metadata sidecar doesnt have the resource and limits set.

pkg/utils/k8sutils.go
malayparida2000: Updated

leelavg: why do you still require the conversions? you changed the value to milli and adjust factor leaves a reminder between 0 to that num (2 = 1, 9 = 8) and float to int anyways would curtail it making it redundant, am I missing something?

malayparida2000: @leelavg Did not get you, Are you asking about the float64 conv or the int64 conv

internal/controller/operatorconfigmap_controller.go
malayparida2000: Now switched to iterate over the struct

leelavg (approved):

PR #647, opened 2026-07-02
feat: add VolumeAttributesClass reconciliation support

Madhu-1:
Add conditional VolumeAttributesClass (storage.k8s.io/v1) support to the StorageClient controller. The API availability is checked at startup via REST mapper and dynamically during reconciliation, ensuring the operator works on clusters where the API is not yet available (K8s <1.34).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -53,6 +53,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
@@ -149,7 +150,7 @@ func main() {
 		setupLog.Error(err, "Unable to get API client")
 		os.Exit(1)
 	}
-	availCrds, err := getAvailableCRDNames(apiCtx, apiClient)
+	availCrdsOrResources, err := getAvailableCRDNames(apiCtx, apiClient)
 	if err != nil {
 		setupLog.Error(err, "Unable get a list of available CRD names")
 		os.Exit(1)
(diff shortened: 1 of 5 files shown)


cmd/main.go
```diff
@@ -221,6 +222,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	if isAPIResourceAvailable(mgr.GetRESTMapper(), schema.GroupVersion{Group: "storage.k8s.io", Version: "v1"}, "volumeattributesclasses") {
```
leelavg, line 225: why a new restmapper? we already have `availCrds` var which holds the names of all crds at startup.

internal/controller/storageclient_controller.go
```diff
@@ -279,6 +284,9 @@ func (r *StorageClientReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.crdsBeingWatched.Store(VolumeGroupSnapshotClassCrdName, false)
 	r.crdsBeingWatched.Store(OdfVolumeGroupSnapshotClassCrdName, false)
 	r.crdsBeingWatched.Store(ObjectBucketCrdName, false)
+	if !r.AvailableCrds[VolumeAttributesClassResourceName] {
```
leelavg: no need for conditional, i see the func already ignores nomatch.

internal/controller/storageclient_controller.go
```diff
@@ -546,6 +559,41 @@ func (r *storageClientReconcile) setupObjectBucketWatch() error {
...
+		return nil
+	}
+
+	vacList := &storagev1.VolumeAttributesClassList{}
+	if err := r.list(vacList, client.Limit(1)); err != nil {
+		if meta.IsNoMatchError(err) {
+			return nil
+		}
+		return err
+	}
```
leelavg, line 571: this is technically correct but the func is supposed to test presence of crd not CRs, since we already know the CRD name, could you pls use the same format as other watch funcs, ie, check crd uid?

internal/controller/storageclient_controller.go
```diff
@@ -89,6 +89,7 @@ const (
 	OdfVolumeGroupSnapshotClassCrdName = "volumegroupsnapshotclasses.groupsnapshot.storage.openshift.io"
 	ObjectBucketClaimCrdName           = "objectbucketclaims.objectbucket.io"
 	ObjectBucketCrdName                = "objectbuckets.objectbucket.io"
+	VolumeAttributesClassResourceName  = "volumeattributesclasses.storage.k8s.io"
```
leelavg, line 92: pls match the format
```suggestion
	VolumeAttributesClassCrdName  = "volumeattributesclasses.storage.k8s.io"
```

cmd/main.go
Madhu-1: Does it works for the standard API calls?

internal/controller/storageclient_controller.go
Madhu-1: This is done because its a native API not the CRD

internal/controller/storageclient_controller.go
Madhu-1: Do you still want me to call this as CRD as its native object?

leelavg (approved):

PR #650, opened 2026-07-07
DFBUGS-8403: [release-4.16] Bump go (stdlib) from 1.21.9 to 1.21.13

Nikhil-Ladha:
### Changes
- **go (stdlib)**: `1.21.9` → `1.21.13` (in `go.mod`)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/red-hat-storage/ocs-client-operator
 
-go 1.21.9
+go 1.21.13
 
 replace (
 	github.com/portworx/sched-ops => github.com/portworx/sched-ops v0.20.4-openstorage-rc3 // required by Rook v1.12


Nikhil-Ladha:
@leelavg any reason for closing the PR?
We were waiting for the QE approval on the bug, hence the PR was pending merge.

leelavg:
I was cleaning up some leftovers since we have GH action updating golang but that's doing upto 4.18, reopened. why do we need qe approval for patch version bump?

Nikhil-Ladha:
> I was cleaning up some leftovers since we have GH action updating golang but that's doing upto 4.18, reopened. why do we need qe approval for patch version bump?

Good question, ideally we shouldn't.
The bug is targetted for next z-version, so we should get it merged in that timeline.

leelavg:
Don't want to have more trackers & timelines, merging it as that issue wasn't serving any purpose if the d/s builder is already at .13, isn't it?

leelavg (approved):

PR #664, opened 2026-07-10
feat: add alert when client operator upgrade is required

aruniiird:
When the provider cluster upgrades, it sends the new desired channel to the client via gRPC which is stored in the StorageClient annotation.

This change adds a Prometheus metric and alert to notify administrators when the client operator subscription channel differs from the provider's desired channel, indicating a manual upgrade is required.

Changes:
- Add 'ocs_client_operator_upgrade_required' Prometheus metric that fires when desired channel != current subscription channel
- Add 'StorageClientUpgradeRequired' alert rule (fires after 5m)
- Support multiple StorageClients with per-client metrics

Ref: [RHSTOR-8681](https://redhat.atlassian.net/browse/RHSTOR-8681)

### DIFF
--- a/internal/controller/alert/resource_collector.go
+++ b/internal/controller/alert/resource_collector.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 Red Hat, Inc.
+Copyright 2026 Red Hat, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import (
 
 	csiopv1 "github.com/ceph/ceph-csi-operator/api/v1"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
+	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-client-operator/pkg/templates"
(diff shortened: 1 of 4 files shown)


internal/controller/alert/resource_collector.go
```diff
@@ -144,3 +151,61 @@ func (c *ResourceCollector) collectVSCCount(ctx context.Context, ch chan<- prome
...
+			// No desired channel annotation, skip this client
+			continue
+		}
+
+		// Check if upgrade is required
+		if desiredChannel != currentChannel {
+			m, err := prometheus.NewConstMetric(
+				desc,
+				prometheus.GaugeValue,
+				1, // upgrade required
```
leelavg, line 183: don't we need to explicitly set to 0 if upgrade is done or this'll be auto set? alert rule says to fire if it's 1 for 5m is that relative to scrape timing?

aruniiird: > don't we need to explicitly set to 0 if upgrade is done or this'll be auto set?

It will not be auto set to `0`, but the metric itself will not be emitted. When a metric is no longer emitted by a collector, Prometheus marks it as "stale" and the time series disappears. Thus the alert expression `ocs_client_operator_upgrade_required > 0` will not return any result and the alert automatically resolves. This is one of the prometheus best practices.
Ref:
https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/ [see the `keep_firing_for` section]

> alert rule says to fire if it's 1 for 5m is that relative to scrape timing?

The `for: 5m` is not directly tied to scrape timing - it's tied to the evaluation interval.
Ref: https://pracucci.com/prometheus-understanding-the-delays-on-alerting.html

leelavg: ok, if it's best practice fine, I was thinking the time series could be used to derive how much time the user didn't act on the alert rather than it disappearing.

leelavg (approved):

PR #671, opened 2026-07-14
csi: pass TLSProfile to csi-addons & csi-snapshot-metadata sidecars

iPraveenParihar:
### Summary

- Bump ceph-csi-operator/api to pick up ContainerExtraArgs on ControllerPluginSpec and NodePluginSpec
- Read the ocs-tls-profile TLSProfile CR and resolve config for `csiaddons.openshift.io` and `cbt.storage.k8s.io` domain
- Convert the resolved TLS settings to CSI-addons CLI flags (`--tls-min-version`, `--tls-cipher-suites`, `--tls-curve-preferences`) and set them on OperatorConfig.spec.driverSpecDefaults for both controller and node plugins
- Extract buildServerTLSOpts into pkg/utils/tls.go for reuse across cmd/main.go and the reconciler


internal/controller/operatorconfigmap_controller.go
```diff
@@ -662,6 +664,12 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 			}
 		}
 		driverSpecDefaults.GenerateOMapInfo = ptr.To(c.shouldGenerateRBDOmapInfo())
+		extraArgs, err := c.csiAddonsTLSArgs()
```
leelavg: move this call outside of createorupdate, we don't want to make new http call inside another, then abstract it with c.UpdateTlsArgs(driverSpecDefaults)

internal/controller/operatorconfigmap_controller.go
```diff
@@ -628,6 +630,11 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 	if err != nil {
 		return fmt.Errorf("failed to get desired imageset configmap name: %v", err)
 	}
+	csiAddonsTLSExtraArgs, err := c.csiAddonsTLSArgs()
```
leelavg: still need a split, csiaddons isn't strictly a concern of the reconciler, get the tlsprofile before updating the drivers and have a func (not method) which takes the tlsprofile along with the domain and return you the translated details for csiaddons.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -628,6 +630,20 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 	if err != nil {
 		return fmt.Errorf("failed to get desired imageset configmap name: %v", err)
 	}
+	tlsProfile := &ocstlsv1.TLSProfile{}
+	tlsProfile.Name = TLSProfileName
+	tlsProfile.Namespace = c.OperatorNamespace
+	if err := c.get(tlsProfile); err != nil {
+		if !kerrors.IsNotFound(err) {
```
leelavg:
```suggestion
	if err := c.get(tlsProfile); client.IgnoreNotFound(err) != nil {
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -628,6 +630,20 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
...
+			return fmt.Errorf("failed to get TLSProfile: %w", err)
+		}
+		tlsProfile = nil
```
leelavg: either make the next set of funcs not check for nil or no need to explicitly set it to nil here. the nil here is only being checked three levels down or if you still want to explicitly set, can do


val csiExtraArgs map[string][]string
....
replace tlsProfile = nil with
csiExtraArgs, err := buildContainerExtraArgs(tlsProfile)
....
if len(csiExtraArgs) > 0 {
        driverSpecDefaults.ControllerPlugin.ContainerExtraArgs = csiExtraArgs
		driverSpecDefaults.NodePlugin.ContainerExtraArgs = csiExtraArgs
}
....


rest lg, thanks for testing for snapshot as well!

internal/controller/operatorconfigmap_controller.go
iPraveenParihar: Instead passed startupProfile to Reconciler.
Thanks!

leelavg (approved):

PR #680, opened 2026-07-21
new grpc client creation without dial timeout

leelavg:
updated only the client submodule in ocs-op as deps but mcg is also clubbed in from 4.22 and so the whole deps are getting pulled, mcg wasn't able to divide their api into submodule.


go.mod
```diff
@@ -19,26 +19,26 @@ exclude (
 )
 
 require (
-	github.com/ceph/ceph-csi-operator/api v0.0.0-20260720045108-15b278476fc9
+	github.com/ceph/ceph-csi-operator/api v0.0.0-20260507023822-4b48ea04fd20
```
Madhu-1: @leelavg looks like this is downgraded, is this expected?

leelavg: thanks, fixed.

Madhu-1 (approved):

PR #705, opened 2026-08-04
console: configure client console nginx TLS from TLSProfile CR

vbnrh:
## Summary
- Wire the `ocs-client-operator-console` nginx server to read TLS settings from the `TLSProfile` CRD and inject `ssl_protocols`, `ssl_ciphers`, and `ssl_conf_command Groups` directives into the nginx ConfigMap
- Add a watch on `TLSProfile` in `OperatorConfigMapReconciler.SetupWithManager()` so nginx config is regenerated when the TLS profile changes
- Add `features.operators.openshift.io/tls-profiles: "true"` CSV annotation

## Changes
- `pkg/console/nginx_root.conf`: add `%s` placeholder for TLS directive injection
- `pkg/console/console.go`: replace `GetNginxRootConf()` with `GenerateNginxConf(ossl)` + `buildTLSDirectives()` 
- `internal/controller/operatorconfigmap_controller.go`: fetch `TLSProfile` in `ensureConsolePlugin()`, resolve OpenSSL config for domain `ocs.openshift.io` / server `client-console`, pass to nginx config generation; add TLSProfile watch with name/namespace/generation predicates
- `pkg/console/console_test.go`: unit tests for `GenerateNginxConf` (nil, TLS 1.3 full, protocol-only)
- CSV template + bundle: add `tls-profiles` feature annotation

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./pkg/console/...` passes (new + existing tests)
- [x] `go test -run 'TestBuildDesiredNginxDataWithProxies' ./internal/controller/...` passes
- [x] `make bundle` regenerates cleanly

Ref: [RHSTOR-8685](https://redhat.atlassian.net/browse/RHSTOR-8685)

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -867,6 +867,12 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 		return err
 	}
 
+	goTLS, err := utils.BuildServerTLSOpts(c.TlsProfile, "ocs.openshift.io", "client-console")
+	if err != nil {
+		return fmt.Errorf("invalid TLSProfile config for console nginx: %w", err)
+	}
+	ossl := ocstlsv1.OpenSSLConfigFrom(goTLS)
+
 	nginxConfigMap := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      console.NginxConfigMapName,
@@ -874,7 +880,7 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 		},
 	}
 	nginxConfigMapResult, err := c.createOrUpdateWithResult(nginxConfigMap, func() error {
-		desiredData, buildErr := c.buildDesiredNginxDataWithProxies()
+		desiredData, buildErr := c.buildDesiredNginxDataWithProxies(ossl)
 		if buildErr != nil {
 			return buildErr
 		}
@@ -922,10 +928,14 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 	return nil
 }
 
-func (c *OperatorConfigMapReconciler) buildDesiredNginxDataWithProxies() (map[string]string, error) {
+func (c *OperatorConfigMapReconciler) buildDesiredNginxDataWithProxies(ossl *ocstlsv1.OpenSSLConfig) (map[string]string, error) {
+	nginxConf, err := console.GenerateNginxRootConf(ossl)
+	if err != nil {
+		return nil, err
+	}
 	out := map[string]string{
 		// Root config is mandatory for nginx to start. Proxy configs (per client) are optional.
-		"nginx.conf": console.GetNginxRootConf(),
+		"nginx.conf": nginxConf,
 	}
 
 	if c.operatorConfigMap.Data != nil {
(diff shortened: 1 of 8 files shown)


leelavg (commented):
not reviewing, pls look at existing code, configmap controller already gets startuptlsprofile and just use it, see it's being reused for csi addons as well. should be straightforward.

pkg/console/console.go
```diff
@@ -85,25 +86,67 @@ func GetConsolePlugin(consolePort int32, serviceNamespace string) *consolev1.Con
...
 }
 
-func GetNginxRootConf() string {
-	return nginxRootConf
+func GenerateNginxConf(ossl *ocstlsv1.OpenSSLConfig) string {
+	return fmt.Sprintf(nginxRootConf, buildTLSDirectives(ossl))
 }
 
-func GetNginxProxyConf(uniqueIdentifier, exposeAs, endpointURL, endpointHost, certsPath string) (string, error) {
+func buildTLSDirectives(ossl *ocstlsv1.OpenSSLConfig) string {
```
SanjalKatiyar: nits:
```suggestion
func buildRootTLSDirectives(ossl *ocstlsv1.OpenSSLConfig) string {
```

pkg/console/console.go
```diff
@@ -85,25 +86,67 @@ func GetConsolePlugin(consolePort int32, serviceNamespace string) *consolev1.Con
 	}
 }
 
-func GetNginxRootConf() string {
-	return nginxRootConf
+func GenerateNginxConf(ossl *ocstlsv1.OpenSSLConfig) string {
```
SanjalKatiyar: nits:
```suggestion
func GetNginxRootConf(ossl *ocstlsv1.OpenSSLConfig) string {
```

SanjalKatiyar: Or,
```suggestion
func GenerateNginxRootConf(ossl *ocstlsv1.OpenSSLConfig) string {
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -941,11 +947,11 @@ func (c *OperatorConfigMapReconciler) buildDesiredNginxDataWithProxies() (map[st
 		}
 	}
 
-	err := c.computeDesiredProxyConfigByKey(out)
+	err := c.computeDesiredProxyConfigByKey(out, console.BuildProxyTLSDirectives(ossl))
```
SanjalKatiyar: nits (optional):  ideally `console.BuildProxyTLSDirectives(ossl)` should happen inside `computeDesiredProxyConfigByKey` func. That's the "desired" responsibility of the function, `ossl` should be the argument.

pkg/console/nginx_root.conf
```diff
@@ -102,7 +102,7 @@ http {
         listen       [::]:9001 ssl;
         ssl_certificate /var/serving-cert/tls.crt;
         ssl_certificate_key /var/serving-cert/tls.key;
-
+%s
```
SanjalKatiyar: can we plz keep formatting aligned with ssl_certificate_key and location strings ??

pkg/console/nginx_proxy.tmpl
```diff
@@ -21,4 +21,7 @@ location /{{.UniqueIdentifier}}/{{.ExposeAs}}/ {
     proxy_ssl_verify on;
     proxy_ssl_server_name on;
     proxy_ssl_trusted_certificate {{.CertsPath}};
+{{- if .ProxyTLSDirectives}}
+{{.ProxyTLSDirectives}}
+{{- end}}
```
SanjalKatiyar: can we plz keep formatting aligned with rest of the file ??

pkg/console/console.go
remypanicker: Real detection gap -- thanks. Command substitution and backtick interpolation inside a single-quoted SQL literal are a genuine injection shape distinct from `$VAR`. Fixed in 76231795ad: added a command-sub/backtick detection regex plus a unified `_span_has_real_interpolation` helper that checks every `$`-expansion, `$(` command-sub, and backtick per span with escape awareness (`\$(` and `` \` `` are bash-literal). Regression test added covering `$(printf %s "$UNTRUSTED")`, backtick `` `id` ``, and the escaped `\$(` negative. Also added the omitted `REINDEX`/`VACUUM` statement keywords in the same commit.

pkg/console/nginx_proxy.tmpl
vbnrh: image
Replaced with individual template fields

pkg/console/nginx_root.conf
SanjalKatiyar, line 159: nits: extension of this file should now be `.tmpl` as well, previously it was just a static config...

leelavg (approved):
thanks for the review @SanjalKatiyar

PR #707, opened 2026-08-06
Add NetworkPolicy for ocs-client-operator-console operand

OdedViner:
Phase 1: Add operand NetworkPolicy via Go reconciliation.

Creates a NetworkPolicy for the ocs-client-operator-console pod allowing ingress from openshift-console on port 9001 (HTTPS ConsolePlugin backend). Egress is unrestricted.

Excluded components:
- storageclient-status-reporter: CronJob with no custom pod labels and no listening ports; NetworkPolicy not applicable.
- ocs-client-operator-controller-manager: operator pod NetworkPolicy will be added in Phase 2.

Ref: [HPSTRAT-104](https://redhat.atlassian.net/browse/HPSTRAT-104)

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -332,6 +332,7 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=operatorconfigs,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=drivers,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
+//+kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=create;update;delete
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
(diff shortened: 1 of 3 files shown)


internal/controller/operatorconfigmap_controller.go
```diff
@@ -332,6 +333,7 @@ func (c *OperatorConfigMapReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=operatorconfigs,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=csi.ceph.io,resources=drivers,verbs=get;list;update;create;watch;delete
 //+kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
+//+kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=create;update;delete
```
rewantsoni, line 335: don't we need to watch, get and list the networkpolicy as well?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -919,9 +921,59 @@ func (c *OperatorConfigMapReconciler) ensureConsolePlugin() error {
 		return err
 	}
 
+	if err := c.reconcileConsoleNetworkPolicy(); err != nil {
+		c.log.Error(err, "failed to reconcile console network policy")
+		return err
+	}
```
rewantsoni: can we not do something similar to what we have for GetConsolePlugin?

rewantsoni (changes requested):

iamniting (commented):
We should not create policies for the console via code as the pod is being deployed by the OLM. We should only do it via bundle in phase 2. I believe we should close this PR for now.

PR #717, opened 2026-08-12
fix: disable csi-addons for nvmeof driver

OdedViner:
The NVMeoF CSI driver does not implement CSI-Addons. Setting `DeployCsiAddons=false` on the `NVMeoF` Driver CR prevents the `ceph-csi-operator` from deploying the `csi-addons` sidecar, which was crash-looping due to a missing csi-addons.sock socket.


rewantsoni (changes requested):

leelavg (approved):
approving as the code is correct and don't want to miss nightly.

PR #721, opened 2026-08-17
controllers: disable fencing for kubeVirt clusters

rewantsoni:
disable auto fencing for nodes tainted with out of service taint, for kubevirt clusters that led to IO being stopped for base clusters as well.

Fixes: [DFBUGS-9233](https://redhat.atlassian.net/browse/DFBUGS-9233)

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -675,6 +675,7 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 			driverSpecDefaults.ControllerPlugin.ContainerExtraArgs = csiExtraArgs
 			driverSpecDefaults.NodePlugin.ContainerExtraArgs = csiExtraArgs
 		}
+		driverSpecDefaults.EnableFencing = ptr.To(!c.shouldDisableAutoFencing())
 		return nil
 	}); err != nil {
 		return fmt.Errorf("failed to reconcile csi operator config: %v", err)
@@ -1491,6 +1492,19 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
 		return fmt.Errorf("failed to reconcile snapshot metadata spec ConfigMap: %w", err)
 	}
 	return nil
+}
+
+func (c *OperatorConfigMapReconciler) shouldDisableAutoFencing() bool {
+	infra := &configv1.Infrastructure{}
+	infra.Name = "cluster"
+	if err := c.get(infra); err != nil {
+		c.log.Info("failed to fetch infrastructure resource", "error", err)
+		return true
+	}
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.Type == configv1.KubevirtPlatformType {
+		return true
+	}
+	return false
 }
 
 func adjustPluginCpuResourcesForIbmZ(resources any, adjustFactor float64) {


internal/controller/operatorconfigmap_controller.go
```diff
@@ -1493,6 +1496,18 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) shouldDisableAutoFencing() bool {
+	infra := &configv1.Infrastructure{}
+	infra.Name = "cluster"
+	if err := c.get(infra); err != nil {
+		return true
+	}
+	if infra.Status.PlatformStatus == nil || infra.Status.PlatformStatus.Type == configv1.KubevirtPlatformType {
```
leelavg: I don't have much background, why are you deciding to turn off the feature when details are unknown (== nil)? if the toggle is a problem the same issue exists in current code as well.

rewantsoni: The PlatformStatus shouldn't toggle. This is to ensure that we don't turn it on until we are sure about the env we are on.

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1493,6 +1496,18 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) shouldDisableAutoFencing() bool {
+	infra := &configv1.Infrastructure{}
+	infra.Name = "cluster"
+	if err := c.get(infra); err != nil {
+		return true
```
Madhu-1: this seems wrong, we should atleast log the failure

rewantsoni: Yes, added a log

Madhu-1 (approved):

internal/controller/operatorconfigmap_controller.go
Madhu-1: @rewantsoni can you make this statement as below

```go
if infra.Status.PlatformStatus != nil &&  infra.Status.PlatformStatus.Type == configv1.KubevirtPlatformType
```

internal/controller/operatorconfigmap_controller.go
```diff
@@ -1493,6 +1496,19 @@ func (c *OperatorConfigMapReconciler) reconcileRbdSMSSpecConfigMap() error {
 	return nil
 }
 
+func (c *OperatorConfigMapReconciler) shouldDisableAutoFencing() bool {
+	infra := &configv1.Infrastructure{}
+	infra.Name = "cluster"
+	if err := c.get(infra); err != nil {
+		c.log.Info("failed to fetch infrastructure resource", "error", err)
+		return true
```
Madhu-1: return false as well?

internal/controller/operatorconfigmap_controller.go
```diff
@@ -675,6 +675,9 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 			driverSpecDefaults.ControllerPlugin.ContainerExtraArgs = csiExtraArgs
 			driverSpecDefaults.NodePlugin.ContainerExtraArgs = csiExtraArgs
 		}
+		if c.shouldDisableAutoFencing() {
+			driverSpecDefaults.EnableFencing = ptr.To(false)
```
Madhu-1: directly assign instead of if check as we always return

internal/controller/operatorconfigmap_controller.go
rewantsoni: If for some reason the api is not reachable and we were not able to fetch the CR in a kubeVirt cluster it could lead to fencing being enabled and it would stop IO from base cluster. Hence choose to return true in case of we can't fetch the CR

internal/controller/operatorconfigmap_controller.go
rewantsoni: sure, I can change it to 
```
driverSpecDefaults.EnableFencing = ptr.To(!c.shouldDisableAutoFencing())
```
The reason I kept it as above is so that it's not confusing, we are doing a "!" of the value we get from shouldDisableAutoFencing()

internal/controller/operatorconfigmap_controller.go
Madhu-1: in that case we should have specific error checks, if Not found return false, any other error return true, we should not assume all errors are same by default

rewantsoni: The chances of hitting the case for not found would be in non-openshift env. I don't think we need specific checks for it

PR #724, opened 2026-08-19
fix incorrect watches ballooning memory usage

leelavg:
all crd's along with specs were being silently cached due to reading full crd for finding the established field which isn't required to establish a watch, this saves a lot of memory and is regression.

in huge clusters with many OSDs, all the deployments were being cached adding more weight along with CSVs which have operators installed in AllNamespaces and tighten them.

Trimming out managedfields also would save non-trivial amount of memory on huge clusters.

Below are some of the numbers (averaged) based on profiling (including overhead):
Deployment: ~200KB
CRD (full): ~130KB
PV: ~12KB

Extrapolating:
[DFBUGS-4532](https://redhat.atlassian.net/browse/DFBUGS-4532): 200M from PVs, 100M from Deps, this was before CRD regr and that would add 45M. With the fix the PV memory would be dominating.
[DFBUGS-9939](https://redhat.atlassian.net/browse/DFBUGS-9939): 80M from CRDs (regr), don't have full data to calculate other overheads.

Overall the CRD regression is the dominating factor and remaining fixes helps huge clusters (ie, 100s of deployments and 1000s of PVs) and so we are optimizing for the general case to leave the niche be directly configurable on subscription itself.

Please note these numbers are based on 10-12 profiles but still may not be accurate, however these are on higher side.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -59,6 +59,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
@@ -379,7 +380,13 @@ func buildCacheAvailableCRDs(
 			LabelSelector: noobaaLabelSelector,
 		},
 	}
+	csvExcludeCopied, err := labels.NewRequirement("olm.copiedFrom", selection.DoesNotExist, nil)
+	if err != nil {
+		setupLog.Error(err, "failed to create CSV label requirement")
+		os.Exit(1)
+	}
 	cacheAvailableCrd := cache.Options{
+		DefaultTransform: cache.TransformStripManagedFields(),
 		ByObject: map[client.Object]cache.ByObject{
 			&admrv1.ValidatingWebhookConfiguration{}: {
 				// only cache our validation webhook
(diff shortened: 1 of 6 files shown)


leelavg:
@shirady fyi only to learn the regression.

internal/controller/storageclient_controller.go
```diff
@@ -1282,15 +1280,3 @@ func (r *storageClientReconcile) getOperatorVersion() (string, error) {
...
-// A CRD object can exist before the REST mapping for its resource is available.
-func crdEstablished(crd *extv1.CustomResourceDefinition) bool {
-	for i := range crd.Status.Conditions {
-		c := crd.Status.Conditions[i]
-		if c.Type == extv1.Established && c.Status == extv1.ConditionTrue {
-			return true
-		}
-	}
-	return false
-}
```
shirady: @leelavg, can you explain why you chose to remove this?
The function is supposed to handle the case where the object existed before it was available.

internal/controller/crds_presence_controller.go
```diff
@@ -41,6 +41,7 @@ func (r *CrdsPresenceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				}),
 				utils.EventTypePredicate(true, false, true, false),
 			),
+			builder.OnlyMetadata,
```
shirady: @leelavg good catch, thanks.

internal/controller/storageclient_controller.go
leelavg: look at OnlyMetadata docs once, if you are only watching metadata and in somewhere you do a full get, then the cache would get populated with all the resources of that Kind. So, the caller of this func, resolved the client.Object (interface) to crd (concrete) and does a get which caches all the crds in the cluster.

the case against that is, if crd isn't established we may error out established watches on CR, which is totally fine since we mark in the map only after establishing the watches.

cmd/main.go
```diff
@@ -391,6 +398,19 @@ func buildCacheAvailableCRDs(
...
+				},
+				// this label is set on client console and operator deployments as part of csv
+				Label: labels.SelectorFromSet(labels.Set{"app.kubernetes.io/part-of": "ocs-client-operator"}),
+			},
+			&opv1a1.ClusterServiceVersion{}: {
+				Namespaces: map[string]cache.Config{
+					operatorNamespace: {},
+				},
+				Label: labels.NewSelector().Add(*csvExcludeCopied),
+			},
```
rewantsoni: should we also cache only the fields of interest in PV, VolumeSnapshotContent and VolumeGroupSnapshotContent?

leelavg: yes, we can, but I tested for 80 PVs and it isn't worth the effort. even if it goes to 1k, it'll be aroubd 10M, what helps than that is separating the deletion flow somehow but can't spend more time on that.

rewantsoni:
LGTM, @Madhu-1 can you also take a look?

Madhu-1 (approved):

leelavg:
/cherry-pick release-5.0

seems it was missed on unreleased branch.

PR #547, opened 2026-04-02
controllers: disable reconcile of upstream volumegroupsnapshotclass

ShravaniVangur:
Disabling reconcile of volumegroupsnapshotclasses until the upstream api is GA.

PR #726, opened 2026-08-20
DFBUGS-10111: [release-4.22] csi: disable setting vgs policy

ShravaniVangur:
Upstream volumegroupsnapshotclass creation is disabled in odf-4.22. Ocp 5.0 has VGSC v1 enabled. Ocs-client-operator detects this CRD and sets `SnapshotPolicy` to `volumeGroupSnapshot`, which enables the csi-snapshotter sidecar to watch for public VGSC resources as they are available.
On an ocp 5.0 cluster, this results in errors in the `csi-snapshotter` container as it searches for v1beta2 api. 

The fix is commenting the check so that the default SnapshotPolicy (volumeSnapshot) is set.

Ref: https://redhat.atlassian.net/browse/DFBUGS-10111

### DIFF
--- a/internal/controller/operatorconfigmap_controller.go
+++ b/internal/controller/operatorconfigmap_controller.go
@@ -643,9 +643,9 @@ func (c *OperatorConfigMapReconciler) reconcileDelegatedCSI(storageClients *v1al
 		}
 		driverSpecDefaults.ImageSet = &corev1.LocalObjectReference{Name: cmName}
 		driverSpecDefaults.ClusterName = ptr.To(string(clusterVersion.Spec.ClusterID))
-		if c.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
-			driverSpecDefaults.SnapshotPolicy = csiopv1.VolumeGroupSnapshotPolicy
-		}
+		//if c.AvailableCrds[VolumeGroupSnapshotClassCrdName] {
+		//	driverSpecDefaults.SnapshotPolicy = csiopv1.VolumeGroupSnapshotPolicy
+		//}
 		if cniNetworkAnnotationValue != "" {
 			if driverSpecDefaults.ControllerPlugin.Annotations == nil {
 				driverSpecDefaults.ControllerPlugin.Annotations = map[string]string{}


leelavg:
based on description, shouldn't the fix be in snapshotter to watch correct api?

iPraveenParihar:
We are seeing the error because the build team is still using 4.22 sidecar (watching beta API) on 4.23 / 5.0 (will watch v1 API). This is not something to be fixed on ocs-client-operator.

iPraveenParihar:
> based on description, shouldn't the fix be in snapshotter to watch correct api?

This should hopefully fix https://github.com/kubernetes-csi/external-snapshotter/pull/1475

ShravaniVangur:
> based on description, shouldn't the fix be in snapshotter to watch correct api?

Previously, ODF 4.22 looked for vgsc crd with api version `v1beta2`, which was the version available behind the featuregate. We had stopped supporting creation of upstream vgsc in 4.22, and had disabled watches in ocs-operator and ocs-client-operator. The function in this pr was the only remaining watch for the vgsc crd, but since we were not creating/utilising the group snapshot functionality(upstream), even with the feature gate enabled, the external-snapshotter was only populating the cache with the type and having no value here: https://github.com/kubernetes-csi/external-snapshotter/blob/0cb7a5ee2d1dbca25ad30f23045487085682b0c5/pkg/sidecar-controller/snapshot_controller_base.go#L184. 

On an OCP 5.0 cluster vgsc is GA'ed with api version v1. The external-snapshotter does not proceed until the cache sync is finished, and since there is an api mismatch the error is observed. 

Commenting the check in ocs-client-operator would prevent the watch for the crd, consistent with how the other watches were handled. 
https://github.com/red-hat-storage/ocs-operator/pull/3741
https://github.com/red-hat-storage/ocs-client-operator/pull/547

leelavg:
then the actual functionality of using vgsc on odf 5.0 will be broken, isn't it?

leelavg (approved):

leelavg:
looking closely, the merge is to a release branch which is correct.

black-dragon74 (approved):
LGTM

PR #734, opened 2026-08-26
Update dependabot to skip indirect dependencies 

AnikaYadav1:
Update Dependabot to bump only direct Go modules and use a 7-day default cooldown.


leelavg (commented):
Sry, cross checking with go.mod, I see ntg needs to be skipped, so this PR cpuld just have the group fix and default days.

leelavg (approved):

PR #735, opened 2026-08-27
update ocs-tls-profiles subscription during upgrade

leelavg:
when installplan is set to auto approval client-op updates the channel on all subscriptions that it owns based on server sent channel and from 4.22 tls-profiles is included and got missed.


leelavg:
fixes: [DFBUGS-10226](https://redhat.atlassian.net/browse/DFBUGS-10226)

